PR #32, opened 2018-08-24
Update configuration to work with new actuator

bison:
This is just a quick fix so we can start testing with a newer actuator image. We need to spend more time thinking about the types here and exposing more of the available options.

### DIFF
--- a/machines/machine-set.yaml
+++ b/machines/machine-set.yaml
@@ -1,40 +1,55 @@
+---
 apiVersion: cluster.k8s.io/v1alpha1
 kind: MachineSet
 metadata:
   name: worker
   namespace: test
   labels:
-    machineapioperator.openshift.io/cluster: test
+    sigs.k8s.io/cluster-api-cluster: test
+    sigs.k8s.io/cluster-api-machine-role: worker
+    sigs.k8s.io/cluster-api-machine-type: worker
 spec:
   replicas: 3
   selector:
     matchLabels:
-      machineapioperator.openshift.io/machineset: worker
-      machineapioperator.openshift.io/cluster: test
+      sigs.k8s.io/cluster-api-machineset: worker
+      sigs.k8s.io/cluster-api-cluster: test
   template:
     metadata:
       labels:
-        machineapioperator.openshift.io/machineset: worker
-        machineapioperator.openshift.io/cluster: test
+        sigs.k8s.io/cluster-api-machineset: worker
+        sigs.k8s.io/cluster-api-cluster: test
+        sigs.k8s.io/cluster-api-machine-role: worker
+        sigs.k8s.io/cluster-api-machine-type: worker
     spec:
       providerConfig:
         value:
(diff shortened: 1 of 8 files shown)


pkg/render/machine-api-operator-config.yaml
```diff
@@ -1,6 +1,12 @@
 apiVersion: v1
```
enxebre: We should move this into the examples folder

manifests/clusterapi-controller.yaml
```diff
@@ -45,14 +45,14 @@ spec:
             cpu: 100m
             memory: 30Mi
       - name: aws-machine-controller
-        image: quay.io/alberto_lamela/aws-machine-controller:mvp # TODO: move this to openshift org
+        image: quay.io/bison/aws-machine-controller:cfc14e5 # TODO: move this to openshift org
```
enxebre: We should be able to use image built by CI now? cc @ingvagabund

machines/machine-set.yaml
```diff
@@ -1,40 +1,53 @@
...
-      versions:
-        kubelet: 0.0.0
-        controlPlane: 0.0.0
-      roles:
-      - Master
-
+          ami:
+            id: "{{ .MachineConfig.AMI }}"
+          credentialsSecret:
+            name: aws-credentials-secret
```
enxebre: This is not available on the installer, we should omit it so the aws client uses the machine ami role

manifests/clusterapi-controller.yaml
ingvagabund: Can't find the image yet.

enxebre:
See follow up for consolidating config here https://jira.coreos.com/browse/CLOUD-138
In general I think we should expose as config only which is strictly necessary for integrating with the installer.
Did you have the chance to adapt to this and test it via https://github.com/openshift/installer/pull/119?

machines/machine-set.yaml
bison: Makes sense. This is nice for testing locally with the AWS actuator though. It should probably just be optional.

bison:
I didn't get a chance to test via the installer yet. I'll give that a shot.

manifests/clusterapi-controller.yaml
bison: I see `machine-api-operator` images in the output of `oc describe is -n openshift origin-v3.11 --server https://api.ci.openshift.org` but they reference the internal registry on that cluster. The images are supposed to be mirrored to Quay and Docker Hub, but I'm not sure where to find them.

machines/machine-set.yaml
bison: Pushed a quick fix to make this optional.

pkg/render/machine-api-operator-config.yaml
bison: Done.

enxebre:
closed via https://github.com/openshift/machine-api-operator/pull/35

PR #34, opened 2018-08-30
add support for openshift actuator

enxebre:
This allows to integrate the openshift actuator with mao and the installer:
- https://github.com/enxebre/installer/blob/mao/modules/bootkube/resources/manifests/mao-config.yaml#L3
- https://github.com/openshift/cluster-api-provider-aws/pull/35

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -131,14 +131,14 @@ func deployMachineSet() {
 
 	for {
 		glog.Info("Trying to deploy Cluster object")
-		if _, err := v1alphaClient.Clusters("test").Create(cluster); err != nil {
+		if _, err := v1alphaClient.Clusters("default").Create(cluster); err != nil {
 			glog.Infof("Cannot create cluster, retrying in 5 secs: %v", err)
 		} else {
 			glog.Info("Created Cluster object")
 		}
 
 		glog.Info("Trying to deploy MachineSet object")
-		_, err = v1alphaClient.MachineSets("test").Create(machineSet)
+		_, err = v1alphaClient.MachineSets("default").Create(machineSet)
 		if err != nil {
 			glog.Infof("Cannot create MachineSet, retrying in 5 secs: %v", err)
 		} else {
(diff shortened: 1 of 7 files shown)


cmd/main.go
```diff
@@ -131,14 +131,14 @@ func deployMachineSet() {
...
-		if _, err := v1alphaClient.Clusters("test").Create(cluster); err != nil {
+		if _, err := v1alphaClient.Clusters("default").Create(cluster); err != nil {
 			glog.Infof("Cannot create cluster, retrying in 5 secs: %v", err)
 		} else {
 			glog.Info("Created Cluster object")
 		}
 
 		glog.Info("Trying to deploy MachineSet object")
-		_, err = v1alphaClient.MachineSets("test").Create(machineSet)
+		_, err = v1alphaClient.MachineSets("default").Create(machineSet)
```
ingvagabund: `default` still better than `test`. Though, is this the final namespace? I would rather make the namespace configurable. Either via configmap or via mao ``--namespace`` or similar option.

enxebre: thanks. Yes, there's separate tickets for addressing dedicated namespaces and service accounts.

machines/machine-set.yaml
```diff
@@ -1,40 +1,56 @@
...
+            id: {{.ClusterName}}-master-profile
+          keyName: tectonic
+          tags:
+            - name: tectonicClusterID
+              value: {{.ClusterID}}
+          securityGroups:
+            - filters:
+              - name: "tag:Name"
+                values:
+                - {{.ClusterName}}_worker_sg
```
bison: This is probably fine for now, but just a thought: If we're going to select security groups this way, we may want to also filter by cluster ID if that's available.

machines/machine-set.yaml
```diff
@@ -1,40 +1,56 @@
...
+            region: {{.Region}}
+            availabilityZone: {{.AvailabilityZone}}
+          subnet:
+            filters:
+            - name: "tag:Name"
+              values:
+              - {{.ClusterName}}-worker-{{.AvailabilityZone}}
+          publicIp: true
+          iamInstanceProfile:
+            id: {{.ClusterName}}-master-profile
```
bison: Should this be `-worker-profile`?

PR #35, opened 2018-08-30
Libvirt support

enxebre:
Follow up for https://github.com/openshift/machine-api-operator/pull/34
See https://github.com/enxebre/installer/blob/mao/modules/bootkube/manifests.tf#L92

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"flag"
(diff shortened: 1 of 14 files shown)


enxebre:
cc @frobware @ingvagabund @bison

bison:
Looks like a reasonable start to me. /lgtm

PR #40, opened 2018-09-05
add machineSet ami filters

enxebre:
Add ami filters so the actuator can get the right image

### DIFF
--- a/machines/aws/machine-set.yaml
+++ b/machines/aws/machine-set.yaml
@@ -28,6 +28,19 @@ spec:
           kind: AWSMachineProviderConfig
           ami:
             id: {{.AWS.Image}}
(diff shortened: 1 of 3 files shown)


enxebre:
cc @bison

bison:
@enxebre: Link to the test output is fixed. The expected output in the tests needs to be updated.

machines/aws/machine-set.yaml
```diff
@@ -28,6 +28,19 @@ spec:
           kind: AWSMachineProviderConfig
           ami:
             id: {{.AWS.Image}}
```
bison: What happens if you pass both here? Which does it choose? This seems like it should be optional.

PR #41, opened 2018-09-06
Allow for deploying into a specific namespace

bison:
This adds the target namepsace for the cluster-api components to the
operator config rather than deploying everything to the default
namespace.  Deploying to a different namespace requires different TLS
assets because the namespace is part of the service's DNS name in the
cluster.  Because of that, this also adds the aggregated API service
CA as part of the config, and removes the secret containing the TLS
assets from the manifests directory.  The OpenShift installer is
already generating these, so we should now be able to use the files
generated by the installer on real clusters, while still using the
pre-generated assets in the repo for development.


enxebre:
Nice!
For this to work:
- You'll need to update the namespace here https://github.com/openshift/installer/blob/master/modules/bootkube/resources/manifests/cluster-apiserver-secret.yaml#L6
- You'll need to update and make sure the secret with the ign config in the installer is using the targetNamespace as well so the actuator can pick it https://github.com/openshift/cluster-api-provider-aws/blob/master/cloud/aws/actuators/machine/actuator.go#L289
- nit: Ideally we might use one namespace for the controllers, another one for the machines (not a blocker)

I reckon the first two points would only work if the namespace exists, so you might need to add a 04-* here https://github.com/openshift/installer/tree/master/modules/bootkube/resources/manifests

PR #42, opened 2018-09-07
add go/yaml lints

enxebre:


enxebre:
See https://github.com/openshift/release/pull/1389
Ping @paulfantom @spangenberg @trawler

PR #43, opened 2018-09-07
update e2e aws

enxebre:
Update aws e2e to use latests config and verify cluster, machineSet and machines are created.

In follow ups we need to pre-create an infra env that satisfy the machineSet spec so we can verify the machines are running agains the aws API
Related to https://jira.coreos.com/browse/CLOUD-146 and https://jira.coreos.com/browse/CLOUD-163


enxebre:
ping @paulfantom @dankenigsberg @trawler @ingvagabund

PR #57, opened 2018-09-17
Use correct IAM profile for worker machines

bison:
We've been using the wrong IAM profile here.

### DIFF
--- a/machines/aws/machine-set.yaml
+++ b/machines/aws/machine-set.yaml
@@ -52,7 +52,7 @@ spec:
               - "{{.AWS.ClusterName}}-worker-*"
           publicIp: true
           iamInstanceProfile:
-            id: "{{.AWS.ClusterName}}-master-profile"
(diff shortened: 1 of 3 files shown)


machines/aws/machine-set.yaml
```diff
@@ -51,8 +51,10 @@ spec:
               values:
               - "{{.AWS.ClusterName}}-worker-*"
           publicIp: true
+          {{- if .AWS.IAMInstanceProfile }}
```
enxebre: why not just always `id: "{{ .AWS.IAMInstanceProfile }}"`?

bison: Just thought it might make testing easier for now since it's totally possible to create these nodes with no profile, which means there's one less thing you have to set up in advance.

bison: Just hard coding the name still for now.

PR #62, opened 2018-09-19
add install namespace

enxebre:


ingvagabund:
What does it do?

enxebre:
we'll mount this folder on the image at manifests/ which will be used to build the release payload for the CVO. Not sure is run level 1 is correct. I'm currently deploy this manually for testing purposes

PR #63, opened 2018-09-20
Add vikaschoudhary16 to OWNERS

vikaschoudhary16:


enxebre:
We need to clean up aws resources
`* aws_nat_gateway.this: Error creating NAT Gateway: NatGatewayLimitExceeded: Performing this operation would exceed the limit of 5 NAT gateways`

PR #64, opened 2018-09-20
E2E

enxebre:
Adds e2e tests:
- Drop appVersion in favour of cvo status
- Use yaml manifests for creating assumptions and operator deployment (then we can reuse some of the real onces within install/*)
- Re create tls certs for openshift-machine-api-operator namespace


enxebre:
green green now cc @vikaschoudhary16

tests/e2e/main.go
```diff
@@ -0,0 +1,433 @@
...
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Dir = assetsDir
+	return cmd.Run()
+}
+
+var rootCmd = &cobra.Command{
+	Use:   "e2e",
+	Short: "Test deployment of cluster-api stack",
+	RunE: func(cmd *cobra.Command, args []string) error {
```
vikaschoudhary16: we should break this into couple more functions like `prepareTestEnv()` and `runTest()`. May be in a follow-up PR.

tests/e2e/main.go
```diff
@@ -0,0 +1,433 @@
...
+			},
+		}
+		maoConfigPopulatedData, err := render.Manifests(configValues, maoConfigTemplateData)
+		if err != nil {
+			glog.Fatalf("Unable to render manifests %q: %v", maoConfigTemplateData, err)
+		} else {
+			mapConfigMap := &apiv1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "cluster-config-v1",
+					Namespace: "kube-system",
```
vikaschoudhary16: should we revert namespace [here](https://github.com/openshift/machine-api-operator/blob/cvo-integration/cmd/main.go#L12) to `kube-system` or have have mao specific namespace here as well?

tests/e2e/infra.tf
```diff
@@ -0,0 +1,92 @@
...
+
+  cidr = "10.0.0.0/16"
+
+  azs = [
+    "us-east-1a",
+    "us-east-1b",
+    "us-east-1c",
+  ]
+
+  private_subnets = [
```
vikaschoudhary16: are these really required for current test? Just for my understanding, if we dont create private/public subnets will aws actuator fail to create instances for machine objects?

tests/e2e/manifests/status-crd.yaml
```diff
@@ -0,0 +1,27 @@
...
+  group: operatorstatus.openshift.io
+  # list of versions supported by this CustomResourceDefinition
+  versions:
+    - name: v1
+      # Each version can be enabled/disabled by Served flag.
+      served: true
+      # One and only one version must be marked as the storage version.
+      storage: true
+  # either Namespaced or Cluster
+  scope: Namespaced
```
vikaschoudhary16: should not this be installed at "Cluster" scope? All operators, including mao,  will then create OperatorStatus CR using this CRD, no?

PR #72, opened 2018-09-24
Integrate node role labeling into the nodelink controller

ingvagabund:
Perform both node-machine matching and node role labelling inside the same controller.

Any machine (or machineset) is expected to set `sigs.k8s.io/cluster-api-node-role` annotation to `master`, `infra` or `compute`. Once done and a node joins the cluster it gets the role assigned.

### DIFF
--- a/cmd/nodelink-controller/main.go
+++ b/cmd/nodelink-controller/main.go
@@ -64,8 +64,8 @@ const (
 	controllerName = "nodelink"
 
 	// machineAnnotationKey is the annotation storing a link between a node and it's machine. Should match upstream cluster-api machine controller. (node.go)
-	machineAnnotationKey = "machine"
-	kubeClusterNamespace = "kube-cluster"
+	machineAnnotationKey  = "machine"
+	nodeRoleAnnotationKey = "sigs.k8s.io/cluster-api-node-role"
 )
 
 // NewController returns a new *Controller.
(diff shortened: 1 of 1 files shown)


derekwaynecarr:
@ingvagabund we should support arbitrary role labels (not just a fixed 3)

cmd/nodelink-controller/main.go
```diff
@@ -64,8 +64,8 @@ const (
 	controllerName = "nodelink"
 
 	// machineAnnotationKey is the annotation storing a link between a node and it's machine. Should match upstream cluster-api machine controller. (node.go)
-	machineAnnotationKey = "machine"
-	kubeClusterNamespace = "kube-cluster"
+	machineAnnotationKey  = "machine"
+	nodeRoleAnnotationKey = "sigs.k8s.io/cluster-api-node-role"
```
derekwaynecarr: is this annotation agreed to with upstream?

what was the motivation for this alternate annotation versus using the existing node role label directly?

derekwaynecarr: i.e. why not just use node-role.kubernetes.io/<role-name> already?

derekwaynecarr: if the annotation is not agreed upon, and upstream lacks a nodelink-controller last i reviewed, we should probably not reserve fields in that namespace.

derekwaynecarr:
@ingvagabund before you enable this, the node link controller cannot treat taints as authoritative in the cluster api., only additive.

cmd/nodelink-controller/main.go
ingvagabund: > is this annotation agreed to with upstream?

I have not spoken to upstream yet.

> what was the motivation for this alternate annotation versus using the existing node role label directly?

 This is randomly chosen annotation. First value that came to my mind so we have something functioning.

> why not just use node-role.kubernetes.io/ already?
> we should support arbitrary role labels (not just a fixed 3)

It's easier to check a value is present in annotations rather than for a prefix. The was the only reason to use other than the expected value. Though, given the `nodelink` controller copies machine's labels into node's labels, we can label a machine with `node-role.kubernetes.io/ROLE` and have it just propagate it into the node. Which basically removes this entire PR.

> @ingvagabund before you enable this, the node link controller cannot treat taints as authoritative in the cluster api., only additive.

I will fix it.

ingvagabund:
For further discussion: https://github.com/openshift/machine-api-operator/pull/74

PR #73, opened 2018-09-25
Add 'install/' and images-json support

vikaschoudhary16:


install/machineapioperator_02_deployment.yaml
```diff
@@ -0,0 +1,61 @@
...
+      k8s-app: machine-api-operator
+  template:
+    metadata:
+      labels:
+        k8s-app: machine-api-operator
+        tectonic-app-version-name: machine-api
+    spec:
+      containers:
+      - name: machine-api-operator
+        image: quay.io/coreos/machine-api-operator:b6a04c2
```
enxebre: this will need to point here now https://hub.docker.com/r/openshift/origin-machine-api-operator/tags/

manifests/clusterapi-controller.yaml
```diff
@@ -38,9 +38,9 @@ spec:
...
       - name: controller-manager
       {{- if .AWS }}
-        image: openshift/origin-aws-machine-controllers:v4.0.0
+        image: {{ .Images.ClusterAPIControllerAWS }}
       {{- else if .Libvirt}}
-        image: gcr.io/k8s-cluster-api/controller-manager:0.0.7
+        image: {{ .Images.ClusterAPIControllerLibvirt }}
       {{- end}}
         command:
         - "./controller-manager"
```
enxebre: what about lines 55 and 59?

install/image-references
```diff
@@ -0,0 +1,8 @@
+kind: ImageStream
+apiVersion: image.openshift.io/v1
+spec:
+  tags:
+  - name: machine-api-operator
+    from:
+      kind: DockerImage
+      name: quay.io/coreos/machine-api-operator:b6a04c2
```
enxebre: this will need to point here now https://hub.docker.com/r/openshift/origin-machine-api-operator/tags/

install/image-references
```diff
@@ -0,0 +1,8 @@
+kind: ImageStream
```
enxebre: we need to make sure install/ content is put on manifests/ within the docker image https://docs.google.com/document/d/1CGZVEyuloZ9oD4NUArW6dnEpi0WFc6BP2tqPSwFZTCY/edit?ts=5b916ae2#heading=h.1zgrwxmpgxbr

install/machineapioperator_02_deployment.yaml
vikaschoudhary16: kept it same as what is used today to keep new changes at minimum. Thought once we have new changes working with same old image, image update can be next step.

install/image-references
vikaschoudhary16: same as above

manifests/clusterapi-controller.yaml
vikaschoudhary16: thanks for pointing out. will update

install/image-references
vikaschoudhary16: sure, will update

enxebre:
for creating the machines we need to wait for api to be available and make sure cluster object is deployed, something like this should work:
```
diff --git a/pkg/operator/operator.go b/pkg/operator/operator.go
index a6b759b..84155fa 100644
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -147,6 +147,11 @@ func (optr *Operator) Run(workers int, stopCh <-chan struct{}) {
 		return
 	}
 	glog.Info("Synched up caches")
+
+	go wait.Poll(machineRolloutPollInterval, machineRolloutTimeout, func() (bool, error) {
+		return deployMachines(*optr)
+	})
+
 	for i := 0; i < workers; i++ {
 		go wait.Until(optr.worker, time.Second, stopCh)
 	}
@@ -154,6 +159,31 @@ func (optr *Operator) Run(workers int, stopCh <-chan struct{}) {
 	<-stopCh
 }

+func deployMachines(optr Operator) (bool, error) {
+	operatorConfig, err := render.Config(optr.config)
+	if err != nil {
+		glog.Errorf("Error decoding operator config: %v", err)
+		panic(err)
+	}
+
+	glog.Info("Trying to deploy Cluster object")
+	if err := optr.syncCluster(*operatorConfig); err != nil {
+		glog.Infof("Cannot create cluster, retrying: %v", err)
+		return false, nil
+	} else {
+		glog.Info("Created Cluster object")
+		glog.Info("Trying to deploy MachineSet object")
+		if err := optr.syncMachineSets(*operatorConfig); err != nil {
+			glog.Infof("Cannot create MachineSet, retrying: %v", err)
+			return false, nil
+		} else {
+			glog.Info("Created MachineSet object Successfully")
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
 func (optr *Operator) eventHandler() cache.ResourceEventHandler {
 	workQueueKey := fmt.Sprintf("%s/%s", optr.namespace, optr.name)
 	return cache.ResourceEventHandlerFuncs{
diff --git a/pkg/operator/sync.go b/pkg/operator/sync.go
index c248aa3..7b8fc3a 100644
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -24,10 +24,10 @@ func (optr *Operator) syncAll(rconfig render.OperatorConfig) error {
 	// syncFuncs is the list of sync functions that are executed in order.
 	// any error marks sync as failure but continues to next syncFunc
 	syncFuncs := []syncFunc{
-		// TODO(alberto): implement this once https://github.com/kubernetes-sigs/cluster-api/pull/488/files gets in
-		//optr.syncMachineClasses,
-		optr.syncMachineSets,
-		optr.syncCluster,
+	// TODO(alberto): implement this once https://github.com/kubernetes-sigs/cluster-api/pull/488/files gets in
+	//optr.syncMachineClasses,
+	//optr.syncNodeLinkController,
+
 	}

 	//if err := optr.syncStatus(v1.OperatorStatusCondition{
@@ -228,6 +228,9 @@ func (optr *Operator) syncClusterAPIController(config render.OperatorConfig) err
 }

 const (
+	machineRolloutPollInterval = 5 * time.Second
+	machineRolloutTimeout      = 10 * time.Minute
+
 	deploymentRolloutPollInterval = time.Second
 	deploymentRolloutTimeout      = 5 * time.Minute
```

frobware:
> +	if err := optr.syncMachineSets(*operatorConfig); err != nil { +	glog.Infof("Cannot create MachineSet, retrying: %v", err) +	return false, nil

In your diff:

```console
+		if err := optr.syncMachineSets(*operatorConfig); err != nil {
+			glog.Infof("Cannot create MachineSet, retrying: %v", err)
+			return false, nil
```
why doesn't this `return false, err`?

pkg/operator/operator.go
```diff
@@ -147,13 +147,47 @@ func (optr *Operator) Run(workers int, stopCh <-chan struct{}) {
...
 	}
 
 	<-stopCh
 }
 
+func deployMachines(optr Operator) (bool, error) {
+	operatorConfig, err := render.Config(optr.config)
+	if err != nil {
+		glog.Errorf("Error decoding operator config: %v", err)
+		panic(err)
```
frobware: Why `panic`? Why not return `err`? Ditto for the other occurrences.

vikaschoudhary16:
non-nil error will result in exit from `wait.Poll()`

frobware:
> non-nil error will result in exit from `wait.Poll()`

So does this mean that it's really `tryDeployMachines() bool` where we log errors that happen in the function and success is indicated when it returns true but the caller isn't supposed to cope with error cases?

pkg/operator/operator.go
frobware: These are two separate things. Sure, wait requires `bool, error`.

However reviewing this function I see that we always return `nil` for error - so as a reader/reviewe of the function I'm asking why does the signature allegedly return an error, when in reality it never (appears) to return one.

If this function is never going to return an error, then just change the caller to be:

```
	go wait.Poll(machineRolloutPollInterval, machineRolloutTimeout, func() (bool, error) {
		return deployMachines(*optr), nil
	})
```
because, in actuality, `deployMachines()` doesn't return error. We seem to have conflated the function signature to make it a one-line call from `wait.Poll()`.

vikaschoudhary16: panic because there is no chance on retry err will get fixed. And not panic'ing at places where there is chance that condition will be met i.e deployment may appear in some time.

enxebre:
this lgtm please @frobware apply the label if you are good with it

frobware:
> this lgtm please @frobware apply the label if you are good with it

The function signature for `deployMachines()` is wrong.

vikaschoudhary16:
@frobware any better :) ?

pkg/operator/operator.go
```diff
@@ -147,13 +147,45 @@ func (optr *Operator) Run(workers int, stopCh <-chan struct{}) {
...
 		go wait.Until(optr.worker, time.Second, stopCh)
 	}
 
 	<-stopCh
 }
 
+func deployMachines(optr Operator) (bool, error) {
+	operatorConfig, err := render.Config(optr.config)
+	if err != nil {
+		glog.Fatalf("Error decoding operator config: %v", err)
```
frobware: This still won't return. Is that the intention? Why can't we drop out of the calling function with an error?

trawler:
@vikaschoudhary16 just a nit: could you also update the run command in https://github.com/openshift/machine-api-operator/blob/master/README.md?

lib/resourcemerge/clusterapi.go
```diff
@@ -31,56 +29,9 @@ func ensureMachineSetSpec(modified *bool, existing *clusterv1.MachineSetSpec, re
...
-}
-
-func clusterProviderConfigValueEqual(existing *clusterv1.ClusterSpec, required clusterv1.ClusterSpec) bool {
-	existingBytes := existing.ProviderConfig.Value.Raw
-	requiredBytes := required.ProviderConfig.Value.Raw
-	if bytes.Equal(existingBytes, requiredBytes) {
-		return true
-	}
-	return false
+	return
```
frobware: If the return is always `false` do we need this function at all?

frobware: Nevermind. Didn't expand the diff selection. :blush:

PR #75, opened 2018-09-25
Fix makefile

trawler:
This commit fixes the build command in the Makefile after changes introduced here: d740ae5


Makefile
```diff
@@ -22,7 +22,7 @@ check: lint fmt vet test
 build: ## Build binary
 	@echo -e "\033[32mBuilding package...\033[0m"
 	mkdir -p bin
-	$(DOCKER_CMD) go build -v -o bin/machine-api-operator cmd/main.go
+	$(DOCKER_CMD) go build -v -o bin/machine-api-operator cmd/*.go
```
frobware: Why can't this remain as `cmd/main.go`?

enxebre:
Any particular reason why you are updating the vendors?

Makefile
trawler: the binary opts (kubeconfig) etc are now moved to start.go, it seems

frobware: But we're still building a binary rooted at main.go, no?

trawler: this fixes the build comand after d740ae53

frobware: > this fixes the build comand after [d740ae5](https://github.com/openshift/machine-api-operator/commit/d740ae53d78c6b6a455fa96d9dca8da2e744f86f)

Building seems really brittle now. 

I'm wondering why we don't just split these into:

- cmd/nodelink-controller
- cmd/machine-api-operator

That way we can  `go build cmd/nodelink-controller/...` and `go build cmd/machine-api-operator/...`

frobware: > But we're still building a binary rooted at main.go, no?

So since d740ae5 `main.go` doesn't actually do anything as the logic is captured by the `init()` block in `start.go`. And that doesn't need to be built as there's nothing in `main.go` that references anything in `start.go`. Yuck!

/cc @vikaschoudhary16

enxebre: It builds with https://github.com/openshift/machine-api-operator/blob/master/Dockerfile#L9
This change is fixing a bug though. apart from the vendor/go updates looks good to me

Makefile
```diff
@@ -22,7 +22,7 @@ check: lint fmt vet test
 build: ## Build binary
 	@echo -e "\033[32mBuilding package...\033[0m"
 	mkdir -p bin
-	$(DOCKER_CMD) go build -v -o bin/machine-api-operator cmd/main.go
+	$(DOCKER_CMD) go build -v -o bin/machine-api-operator cmd/machine-api-operator/*.go
```
ingvagabund: No need to specify any `*.go` file if you use the full path as Andrew suggested.

PR #77, opened 2018-09-25
Bake nodelink-controller into the mao docker image

ingvagabund:
@enxebre as discussed let's put the `nodelink-controller` binary in the mao image so we can deploy it with the actuators.


frobware:
lgtm, somebody else for +2.

PR #80, opened 2018-09-28
Run MAO in its own namespace

vikaschoudhary16:
This is being done so that MAO can be run in its own namespace instead kube-system

### DIFF
--- a/cmd/machine-api-operator/main.go
+++ b/cmd/machine-api-operator/main.go
@@ -22,7 +22,6 @@ var (
 )
 
 func init() {
-	rootCmd.PersistentFlags().StringVar(&config, "config", "/etc/mao-config/config", "path to the mao config")
 	rootCmd.PersistentFlags().AddGoFlagSet(flag.CommandLine)
 }
 
--- a/cmd/machine-api-operator/main.go
+++ b/cmd/machine-api-operator/main.go
@@ -9,7 +9,7 @@ import (
 
 const (
 	componentName      = "machine-api-operator"
-	componentNamespace = "kube-system"
+	componentNamespace = "openshift-machine-api-operator"
 )
 
 var (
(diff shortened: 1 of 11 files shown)


cmd/machine-api-operator/main.go
```diff
@@ -9,7 +9,7 @@ import (
 
 const (
 	componentName      = "machine-api-operator"
-	componentNamespace = "kube-system"
+	componentNamespace = "openshift-machine-api-operator"
```
enxebre: do we need to have componentNamespace and default namespace?

pkg/operator/operator.go
```diff
@@ -294,3 +298,34 @@ func (optr *Operator) sync(key string) error {
...
+	)
+	clusterConfig, err := optr.kubeClient.CoreV1().ConfigMaps(clusterConfigNamespace).Get(clusterConfigName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("could not find cluster-config-v1 namespace: %v", err)
+	}
+	return mcFromClusterConfig(clusterConfig)
+}
+
+func mcFromClusterConfig(cm *v1.ConfigMap) (*render.OperatorConfig, error) {
+	var (
```
enxebre: can't we use same syntax everywhere?

cmd/machine-api-operator/main.go
bison: Keep in mind the installer is currently using `openshift-cluster-api` for our stuff.

enxebre: @bison This should be transparent so the operator would be deployed to "openshift-machine-api-operator" and the cluster-api stack along with the machineSet object will fall under "openshift-cluster-api" as the installer config is overriding the targetNamespace, @vikaschoudhary16 correct?

vikaschoudhary16: right. If targetNamespace is not set in the configmap, only then it will be initialized to openshift-machine-api-operator.

PR #82, opened 2018-09-28
update apisever image

enxebre:
Update api server to include addresses in status spec
We can drop this one this lands https://github.com/kubernetes-sigs/cluster-api/pull/494


enxebre:
cc @ingvagabund

PR #87, opened 2018-10-03
update namespace for informers

enxebre:
Currently informers are filtering to namespace "openshift-machine-api-operator" https://github.com/openshift/machine-api-operator/blob/master/cmd/machine-api-operator/start.go#L52

But we deploy the components to a different namespace from the installer config (openshift-cluster-api) https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/operator.go#L320

This pass the namespace where the components are deployed to the informers. We need to respect the name "openshift-cluster-api" as the cluster-apiserver-certs are deployed there from the installer

Once CRDs lands we need to:
Drop cluster-apiserver-certs
Remove ability to inject targetNamespace
Reconsider name "openshift-cluster-api"


enxebre:
ping @vikaschoudhary16 @bison @ingvagabund

enxebre:
need to update the test certs, on it

bison:
/lgtm

Still.

Issue #84, opened 2018-10-02
Dropping Container Linux?


wking:
The installer dropped support for Container Linux in openshift/installer#221.  But there appear to still be some assumptions about Container Linux in this repository (e.g. [here][1]).  Is there a plan for dropping that?  Or am I just misunderstanding something about how this operator fits into the cluster?

[1]: https://github.com/openshift/machine-api-operator/blob/f2689d525af210706c152075a7570cd60855999d/pkg/render/config.go#L39-L40

enxebre:
The filter assumptions are actually not applying as the ami id is passed. Yes, we might wanna change to rhcos filters or just expect always an id

enxebre:
Closing this as machines definition lives on the installer now
For reference:
openshift/installer#468
https://github.com/openshift/installer/tree/master/pkg/asset/machines

PR #89, opened 2018-10-05
Add OpenStack as a supported platform

flaper87:
I'm adding support for OpenStack to the various operators, controllers, and providers that manage an openshift cluster.

I'm not sure if this PR should depend on others so I'd really appreciate some guidance in the process. :smile:


machines/openstack/worker.machineset.yaml
```diff
@@ -0,0 +1,72 @@
...
+          {{- if .OpenStack.WithCreds }}
+          credentialsSecret:
+            name: openstack-credentials-secret
+          {{- end}}
+          ami:
+            id: {{.OpenStack.Image}}
+            filters:
+              - name: "name"
+                values:
+                - CoreOS-{{.OpenStack.ReleaseChannel}}-{{.OpenStack.ContainerLinuxVersion}}-*
```
wking: These filters are consistent with what we're doing now for AWS.  But that approach is stale (openshift/machine-api-operator#84).  I'd rather drop this section until we want to come back and address RHCOS/oscontainer auto-updates.  There's more background on future plans in openshift/installer#281.

wking:
[images][1]:

```
error: unable to upload new layer (0): Patch https://registry.svc.ci.openshift.org/v2/ci-op-n7xnxtmq/release/blobs/uploads/366fa0dc-673f-4cae-8b52-ac4d1f096bdf?_state=Co6UHx_0IOuMH8d4klUWwSfQFAl352HY2XT-iHRctp17Ik5hbWUiOiJjaS1vcC1uN3hueHRtcS9yZWxlYXNlIiwiVVVJRCI6IjM2NmZhMGRjLTY3M2YtNGNhZS04YjUyLWFjNGQxZjA5NmJkZiIsIk9mZnNldCI6MCwiU3RhcnRlZEF0IjoiMjAxOC0xMC0wNVQwOTo0OTowMS45MDI0NTcxMDlaIn0%3D: operator "machine-api-operator" failed to map images: requested mapping for "docker.io/openshift/origin-openstack-machine-controllers:v4.0", but no input image could be located
```

So getting origin-openstack-machine-controllers pushed (with an openshift/cluster-api-provider-openstack matching [this][2]) blocks this PR.

[1]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-api-operator/89/pull-ci-openshift-machine-api-operator-master-images/81/build-log.txt
[2]: https://github.com/openshift/cluster-api-provider-aws

machines/openstack/worker.machineset.yaml
enxebre: agree, let's only rely on explicit image id for now

pkg/render/config.go
```diff
@@ -29,6 +30,19 @@ type LibvirtConfig struct {
...
+type OpenStackConfig struct {
+	ClusterName           string `json:"clusterName"`
+	ClusterID             string `json:"clusterID"`
+	Region                string `json:"region"`
+	AvailabilityZone      string `json:"availabilityZone"`
+	Image                 string `json:"image"`
+	ReleaseChannel        string `json:"releaseChannel"`
+	ContainerLinuxVersion string `json:"containerLinuxVersion"`
+	Replicas              string `json:"replicas"`
+	WithCreds             bool   `json:"withCreds"`
```
enxebre: WithCreds is used in aws so if set the actuator use those credentials to interact with the aws api, otherwise it uses the ami role in the given machine. How would this be handle for openstack?

machines/openstack/worker.machineset.yaml
enxebre: below this line this is all specific for the aws actuator config https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/cloud/aws/providerconfig/v1alpha1/types.go
This will need to go in hand with the openstack actuator is there a link for that? Is this to be used with upstream one https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/cloud/openstack/openstackproviderconfig/v1alpha1/types.go?

enxebre:
thanks a lot! @flaper87 dropped a few comments/questions

pkg/render/config.go
flaper87: This would be handled by reading a `clouds.yaml` file from the system. It's a similar architecture to AWS, just different file format.

flaper87:
@wking @enxebre thanks a bunch for the reviews.

> So getting origin-openstack-machine-controllers pushed (with an openshift/cluster-api-provider-openstack matching this) blocks this PR.

What's the process to get `cluster-api-provider-openstack` pushed to the openshift registry? I noticed the respective `aws` project upstream has been forked under the openshift organization. Should the same be done with OpenStack? Do we have any docs on how I can build and upload those images?

machines/openstack/worker.machineset.yaml
flaper87: yup, removed these lines. Using just the ID now.

machines/openstack/worker.machineset.yaml
```diff
@@ -0,0 +1,57 @@
...
+      sigs.k8s.io/cluster-api-cluster: {{.OpenStack.ClusterName}}
+  template:
+    metadata:
+      labels:
+        sigs.k8s.io/cluster-api-machineset: worker
+        sigs.k8s.io/cluster-api-cluster: {{.OpenStack.ClusterName}}
+        sigs.k8s.io/cluster-api-machine-role: worker
+        sigs.k8s.io/cluster-api-machine-type: worker
+    spec:
+      providerConfig:
```
enxebre: this is specific for the aws actuator you'll need to update this config to match the definition on the actuator https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/cloud/openstack/openstackproviderconfig/v1alpha1/types.go#L24

enxebre:
@flaper87 we'll need to go through https://jira.coreos.com/browse/CLOUD-145 to add the openstack images
For building the images from a given commit upstream with could add automation here and embed the openstack binaries on the mao image. @bison @ingvagabund @vikaschoudhary16 thoughts?

machines/openstack/worker.machineset.yaml
flaper87: A-ha, gotcha! Will do

install/image-references
```diff
@@ -10,7 +10,11 @@ spec:
...
+  - name: openstack-machine-controllers
+    from:
+      kind: DockerImage
+      name: docker.io/openshift/origin-openstack-machine-controllers:v4.0.0
   - name: libvirt-machine-controllers
     from:
       kind: DockerImage
-      name: docker.io/openshift/origin-libvirt-machine-controllers:v4.0.0
\ No newline at end of file
+      name: docker.io/openshift/origin-libvirt-machine-controllers:v4.0.0
```
flaper87: @enxebre @wking  Is it possible to use an image from the dev ImageStream here? The `openstack-machine-controllers` image is now being built by our CI system but it won't be published on docker hub for a while and the mirror to quay.io has not been set yet.

smarterclayton: Just reference registry.svc.ci.openshift.org/openshift/origin-v4.0:openstack-machine-controllers

That’s where the image will be pulled from anyway.

flaper87: a-ha, thanks a lot. I'll do that.

flaper87:
I think this is ready for another review. I'm working on an `e2e-openstack` job in parallel but I don't think we should block this patch on that. We'll need more iterations on this.

install/image-references
```diff
@@ -10,7 +10,11 @@ spec:
     from:
       kind: DockerImage
       name: docker.io/openshift/origin-aws-machine-controllers:v4.0.0
+  - name: openstack-machine-controllers
+    from:
+      kind: DockerImage
+      name: registry.svc.ci.openshift.org/openshift/origin-v4.0:openstack-machine-controllers
```
enxebre: why `registry.svc.ci.openshift.org`?

machines/openstack/worker.machineset.yaml
```diff
@@ -0,0 +1,45 @@
+---
```
enxebre: We are planning to move the machineSet specs generation into the installer, so may be move this into that repo, so then we just make sure we deploy the right machine controller here

pkg/operator/sync.go
```diff
@@ -91,6 +91,10 @@ func (optr *Operator) syncMachineSets(config render.OperatorConfig) error {
 		machineSets = []string{
 			"machines/aws/worker.machineset.yaml",
 		}
+	case providerOpenStack:
```
enxebre: we need to drop this

pkg/render/config.go
```diff
@@ -29,6 +30,28 @@ type LibvirtConfig struct {
 	Replicas    string `json:"replicas"`
 }
 
+// OpenStackConfig contains specific config for OpenStack
+type OpenStackConfig struct {
```
enxebre: we can drop all this based as we move the machine spec to the installer

pkg/render/config.go
```diff
@@ -29,6 +30,16 @@ type LibvirtConfig struct {
 	Replicas    string `json:"replicas"`
 }
 
+// We should switch to predictable names/filters as soon as
+// it is available in the OpenStack actuator.
```
enxebre: this is not needed

pkg/render/render_test.go
```diff
@@ -152,6 +152,38 @@ spec:
         controlPlane: ""`)
 }
 
+func TestClusterOpenStackManifest(t *testing.T) {
+	config := OperatorConfig{
+		TargetNamespace: "go-test",
+		Provider:        "OpenStack",
+		OpenStack: &OpenStackConfig{
```
enxebre: this is not needed

pkg/render/config.go
flaper87: ooops

pkg/render/config.go
```diff
@@ -42,12 +43,20 @@ type AWSConfig struct {
 	WithCreds             bool   `json:"withCreds"`
 }
 
+// OpenStackConfig contains specific config for OpenStack
+type OpenStackConfig struct {
+	ClusterName           string `json:"clusterName"`
```
enxebre: you won't need these values, `ClusterName` can be something arbitrary and clusterID can be generated on the spec it self, right?

flaper87: ok ok, I'll remove it :smile:

pkg/render/config.go
```diff
@@ -42,12 +43,17 @@ type AWSConfig struct {
 	WithCreds             bool   `json:"withCreds"`
 }
 
+type OpenStackConfig struct {
```
enxebre: this can be dropped. Then and as we move all machiineSets spec the discrimination for the controller should be done `Provider`

PR #92, opened 2018-10-08
remove hard-coding of AWS instance type

### LINKED ISSUE #85: Configurable instance types on AWS?
This could be me just not understanding how this operator is wired up, but it looks like you currently hard-code `m4.large` [here][1].  I'd have expected that information to come from [the installer][2] via the [cluster config][3].  Am I missing something?  Or is this just future work that hasn't happened yet?

[1]: https://github.com/openshift/machine-api-operator/blob/f2689d525af210706c152075a7570cd60855999d/machines/aws/worker.machineset.yaml#L48
[2]: https://github.com/openshift/installer/blob/52ddd51ae6bbab5a795cd08ca59c41dd33a643c9/pkg/types/machinepools.go#L32-L34
[3]: https://github.com/openshift/installer/blob/52ddd51ae6bbab5a795cd08ca59c41dd33a643c9/pkg/asset/manifests/operators.go#L90-L96

trawler:
resolves #85

### DIFF
--- a/machines/aws/worker.machineset.yaml
+++ b/machines/aws/worker.machineset.yaml
@@ -45,7 +45,7 @@ spec:
               - name: "owner-id"
                 values:
                 - "595879546273"
-          instanceType: m4.large
+          instanceType: {{.AWS.InstanceType}}
           placement:
             region: {{.AWS.Region}}
             availabilityZone: {{.AWS.AvailabilityZone}}
--- a/pkg/render/config.go
+++ b/pkg/render/config.go
@@ -40,6 +40,7 @@ type AWSConfig struct {
 	ContainerLinuxVersion string `json:"containerLinuxVersion"`
 	Replicas              string `json:"replicas"`
 	WithCreds             bool   `json:"withCreds"`
+	InstanceType          string `json:"instanceType"`
 }
 
 // Images allows build systems to inject images for MAO components.
--- a/pkg/render/render_test.go
+++ b/pkg/render/render_test.go
@@ -38,11 +38,11 @@ func TestClusterAWSManifest(t *testing.T) {
 		TargetNamespace: "go-test",
 		Provider:        "AWS",
 		AWS: &AWSConfig{
-			ClusterName:      "TestClusterManifest-ClusterName",
-			ClusterID:        "TestClusterManifest-ClusterID",
-			Region:           "TestClusterManifest-Region",
 			AvailabilityZone: "TestClusterManifest-AvailabilityZone",
+			ClusterID:        "TestClusterManifest-ClusterID",
+			ClusterName:      "TestClusterManifest-ClusterName",
 			Image:            "TestClusterManifest-Image",
+			Region:           "TestClusterManifest-Region",
 			Replicas:         "TestClusterManifest-Replicas",
 		},
 	}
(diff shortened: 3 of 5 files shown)


enxebre:
is the installer passing this value? what happens if this is empty?

enxebre:
@trawler this is probably a legit fail based on my comment above as the test config don't set that value

enxebre:
@trawler The installer is currently not setting this value, we'll want a counter part PR to set this from the installer config.
How everr looking at the actuator https://github.com/openshift/cluster-api-provider-aws/blob/master/cloud/aws/actuators/machine/actuator.go#L381
```type RunInstancesInput struct``` would use a default value 
```
// The instance type. For more information, see Instance Types (http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html)
	// in the Amazon Elastic Compute Cloud User Guide.
	//
	// Default: m1.small
	InstanceType *string `type:"string" enum:"InstanceType"
```
and the e2e-aws shows green, so not sure why our e2e was failing when the value was not set, can you please verify this?

**Comment by trawler** [comment]
> @trawler The installer is currently not setting this value, we'll want a counter part PR to set this from the installer config.
> How everr looking at the actuator https://github.com/openshift/cluster-api-provider-aws/blob/master/cloud/aws/actuators/machine/actuator.go#L381
> `type RunInstancesInput struct` would use a default value
> 
> ```
> // The instance type. For more information, see Instance Types (http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html)
> 	// in the Amazon Elastic Compute Cloud User Guide.
> 	//
> 	// Default: m1.small
> 	InstanceType *string `type:"string" enum:"InstanceType"
> ```
> and the e2e-aws shows green, so not sure why our e2e was failing in your previews commit when the value was not set in the config, can you please verify this to ensure there's no gaps from installer->actuator?

@enxebre I can confirm that when the instanceType is not being set, installation fails with this reply from the server:
`time="2018-10-11T12:55:03Z" level=error msg="error creating EC2 instance: InvalidParameterValue: Invalid value '' for InstanceType.\n\tstatus code: 400, request id: 7b516eca-f3f4
-40cb-b6c7-5d3b87057ccb" controller=awsMachine machine=openshift-cluster-api/worker-xkpbr`

I'm not sure why the comment you quoted claims to set a default value, but actually, the aws-sdk will fail the install for an empty `InstanceType` with an InvalidParams error: 
https://github.com/aws/aws-sdk-go/blob/master/service/ec2/api.go#L23606

`// Validate inspects the fields of the type to determine if they are valid.
func (s *AllocateHostsInput) Validate() error {
  invalidParams := request.ErrInvalidParams{Context: "AllocateHostsInput"}
  if s.AvailabilityZone == nil {
    invalidParams.Add(request.NewErrParamRequired("AvailabilityZone"))
  }
  if s.InstanceType == nil {
    invalidParams.Add(request.NewErrParamRequired("InstanceType"))`

**Comment by trawler** [comment]
e2e-aws test failing for: 
```
* aws_eip.nat_eip.3: Error creating EIP: AddressLimitExceeded: The maximum number of addresses has been reached.
	status code: 400, request id: 2f81367d-af16-4462-b41e-84b7fc3ba5a2
```

wking:
> is the installer passing this value? what happens if this is empty?

Where do we want the default value to live?  If it's here, that's great.  If it's in the installer, what happens if a user later bumps the config to remove the property (in an attempt to revert to the default value)?  Or will tgat sort of adjustment be blocked somehow?

wking:
Also, it may not be worth sinking time into this if [the installer starts creating worker machine-sets directly][1].

[1]: https://github.com/openshift/installer/pull/451#issuecomment-428625419

enxebre:
closing as specs are now moved to the installer

PR #94, opened 2018-10-10
fix rolebinding namespace

deads2k:
The `extension-apiserver-authentication-reader` must always be in the `kube-system` namespace because that is where the role is and that is where  the resources it wants access to are located.

I'm not sure if there's a bindata I need to update somewhere.

/assign @enxebre

### DIFF
--- a/owned-manifests/clusterapi-apiserver-role-binding.yaml
+++ b/owned-manifests/clusterapi-apiserver-role-binding.yaml
@@ -2,7 +2,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: clusterapi
-  namespace: {{ .TargetNamespace }}
+  namespace: kube-system
 roleRef:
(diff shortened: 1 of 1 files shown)


owned-manifests/clusterapi-apiserver-role-binding.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: clusterapi
-  namespace: {{ .TargetNamespace }}
+  namespace: kube-system
```
ingvagabund: @enxebre what's the `TargetNamespace` set to by default?

deads2k: > @enxebre what's the `TargetNamespace` set to by default?

Based on usage below in this file and what I saw, it is the namespace where you're creating a server.  Which is fine in general, but this resource belongs in a different namespace.

PR #98, opened 2018-10-18
update images

enxebre:
update images names to match image-references and update libvirt image


enxebre:
cc @vikaschoudhary16 @paulfantom

PR #100, opened 2018-10-18
Crds

enxebre:
Drop aggregated server in favour of CRDs
Points to CRDs custom actuator image
TODO:
update libvirt image
update aws image


enxebre:
CI system seems unavailable.. @paulfantom yes this goes against master. It has to point to the version of the actuators using CRDs, so this becomes part of the release payload pointing to those specific images.
After that we merge the actuators into master so image stream gets updated
Then we can create a final PR here for putting back the the actuator images which will contain the CRD code

install/0000_50_machine-api-operator_01_images.configmap.yaml
```diff
@@ -4,4 +4,4 @@ metadata:
...
+  images.json: '{"clusterAPIControllerAWS": "registry.svc.ci.openshift.org/openshift/aws-machine-controllers:crd", "clusterAPIControllerManagerAWS": "registry.svc.ci.openshift.org/openshift/aws-machine-controllers:crd", "clusterAPIControllerOpenStack": "docker.io/openshift/origin-openstack-machine-controllers:v4.0.0", "clusterAPIControllerManagerOpenStack": "docker.io/openshift/origin-openstack-machine-controllers:v4.0.0", "clusterAPIControllerManagerLibvirt": "registry.svc.ci.openshift.org/openshift/libvirt-machine-controllers:crd", "clusterAPIControllerLibvirt": "registry.svc.ci.openshift.org/openshift/libvirt-machine-controllers:crd"}'
```
ingvagabund: Here as well: https://github.com/openshift/machine-api-operator/blob/3a9961e8bf8e616f9db233607a434f57679065a4/tests/e2e/manifests/images.configmap.yaml#L10-L11

ingvagabund:
Not related to the PR but we need to deploy the `nodelink-controller` in upcoming PR as well.

install/0000_50_machine-api-operator_01_images.configmap.yaml
enxebre: thanks! fyi that does not have any effect, libvirt is not being tested here. We need to add framework like testing in this repo

PR #101, opened 2018-10-23
Remove machineset and cluster objects

enxebre:
Remove machineSet and cluster definitions as this is being created by the installer https://github.com/openshift/installer/tree/master/pkg/asset/machines
Prepare config to be dropped https://github.com/openshift/installer/pull/518
blocked by https://github.com/openshift/machine-api-operator/pull/100


ingvagabund:
Looks good in overall. So from now on the installer is responsible for deploying the cluster object, right?

PR #106, opened 2018-10-26
drop deprecated config. Use installer config

enxebre:
since machineSets objects are being generate on the installer now the config is no needed in the operator anymore.
- Drops old config
- Get provider from installConfig


enxebre:
cc @bison @vikaschoudhary16

owned-manifests/clusterapi-manager-controllers.yaml
```diff
@@ -55,18 +47,8 @@ spec:
...
-      {{- if eq .Provider "openstack" }}
-      - name: openstack-machine-controller
-        image: {{ .Images.ClusterAPIControllerOpenStack }}
-      {{end}}
-      {{- if eq .Provider "libvirt" }}
-      - name: libvirt-machine-controller
-        image: {{ .Images.ClusterAPIControllerLibvirt }} 
-      {{- end}}
+      - name: machine-controller
+        image: {{ .Controller }}
```
frobware: How does it now chose the correct one?

enxebre: discrimination happens in go https://github.com/openshift/machine-api-operator/pull/106/files#diff-b77facdf2f36a2cc873ec9ea3fae6b00R138

pkg/operator/config.go
```diff
@@ -0,0 +1,163 @@
...
+	ClusterAPIControllerOpenStack string `json:"clusterAPIControllerOpenStack"`
+	ClusterAPIControllerLibvirt   string `json:"clusterAPIControllerLibvirt"`
+}
+
+// InstallConfig contains the mao relevant config coming from the install config, i.e provider
+type InstallConfig struct {
+	IntallPlatform `json:"platform"`
+}
+
+// Platform is the configuration for the specific platform upon which to perform
```
frobware: Comment should be "InstallPlatform" to match the type.

pkg/operator/config.go
```diff
@@ -0,0 +1,163 @@
...
+func getInstallConfig(client kubernetes.Interface) (*InstallConfig, error) {
+	cm, err := client.CoreV1().ConfigMaps(ClusterConfigNamespace).Get(ClusterConfigName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed getting clusterconfig %s/%s: %v", ClusterConfigNamespace, ClusterConfigName, err)
+	}
+
+	return getInstallConfigFromClusterConfig(cm)
+}
+
+// UnmarshalInstallConfig builds an install config from the cluster config.
```
frobware: Comment doesn't match func name.

pkg/operator/config.go
```diff
@@ -0,0 +1,163 @@
...
+func getInstallConfigFromClusterConfig(clusterConfig *corev1.ConfigMap) (*InstallConfig, error) {
+	icYaml, ok := clusterConfig.Data[InstallConfigKey]
+	if !ok {
+		return nil, fmt.Errorf("missing %q in configmap", InstallConfigKey)
+	}
+	var ic InstallConfig
+	if err := yaml.Unmarshal([]byte(icYaml), &ic); err != nil {
+		return nil, fmt.Errorf("invalid InstallConfig: %v\nyaml:\n%s", err, icYaml)
```
frobware: Returning an error string with embedded newlines seems odd (to me at least).

pkg/operator/config.go
```diff
@@ -0,0 +1,163 @@
...
+
+	var i Images
+	if err := json.Unmarshal(data, &i); err != nil {
+		return nil, err
+	}
+	return &i, nil
+}
+
+func getProviderControllerFromImages(provider Provider, images Images) (string, error) {
+	switch provider{
```
frobware: This looks badly formatted. Would expect `gofmt` to leave a space before `{`.

pkg/operator/config.go
```diff
@@ -0,0 +1,163 @@
...
+		return images.ClusterAPIControllerAWS, nil
+	case LibvirtProvider:
+		return images.ClusterAPIControllerLibvirt, nil
+	case OpenStackProvider:
+		return images.ClusterAPIControllerOpenStack, nil
+	}
+	return "", fmt.Errorf("not known platform provider given %s", provider)
+}
+
+// PopulateTemplate receives a template file path and renders is content populated with the config
```
frobware: typo: ... renders its content ...

(I think.)

pkg/operator/config.go
```diff
@@ -0,0 +1,163 @@
...
+func PopulateTemplate(config *OperatorConfig, path string) ([]byte, error) {
+
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed reading file, %v", err)
+	}
+
+	buf := new(bytes.Buffer)
```
frobware: Is there a reason to use `new` over `buf := bytes.Buffer{}`?

pkg/operator/config_test.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	res, err := getInstallConfigFromClusterConfig(&cfg)
+	if err != nil {
+		t.Errorf("failed to get install config: %v", err)
+	}
+	if res.IntallPlatform.AWS != nil && res.IntallPlatform.Libvirt == nil && res.IntallPlatform.OpenStack == nil {
+		t.Logf("got install config successfully: %+v", res)
+	} else {
+		t.Errorf("failed to installConfigFromClusterConfig, expected: '&{Platform:{AWS:0xc4201fd960 Libvirt:<nil> OpenStack:<nil>}}', got: %+v", res)
```
frobware: typo: `failed to getInstallConfigFromClusterConfig`.

frobware: Also asking how `0xc4201fd960` is a known value.

pkg/operator/operator.go
```diff
@@ -232,14 +203,9 @@ func (optr *Operator) sync(key string) error {
...
-	operatorConfig, err := optr.getOperatorConfig()
+	operatorConfig, err := optr.maoConfigFromInstallConfig()
 	if err != nil {
-		glog.Fatalf("Error decoding operator config: %v", err)
-		return err
-	}
-	err = optr.updateImageDetails(operatorConfig)
-	if err != nil {
-		glog.Fatalf("Error getting image details: %v", err)
+		glog.Fatalf("failed getting operator config: %v", err)
```
frobware: This should be `Errorf`. `err` will never be returned as `Fatalf` exits.

pkg/operator/sync.go
```diff
@@ -45,7 +44,7 @@ func (optr *Operator) syncCustomResourceDefinitions(config render.OperatorConfig
 	}
 
 	for _, crd := range crds {
-		crdBytes, err := render.PopulateTemplate(&config, ownedManifestsDir+crd)
+		crdBytes, err := PopulateTemplate(&config, ownedManifestsDir+crd)
```
frobware: Use https://golang.org/pkg/path/#Join for constructing the Path. Assuming I'm reading this correctly. I saw `...Dir + crd`. And assumed Dir was as directory.

pkg/operator/sync.go
```diff
@@ -63,8 +62,8 @@ func (optr *Operator) syncCustomResourceDefinitions(config render.OperatorConfig
 	return nil
 }
 
-func (optr *Operator) syncClusterAPIController(config render.OperatorConfig) error {
-	crBytes, err := render.PopulateTemplate(&config, fmt.Sprintf("%s/clusterapi-manager-cluster-role.yaml", ownedManifestsDir))
+func (optr *Operator) syncClusterAPIController(config OperatorConfig) error {
+	crBytes, err := PopulateTemplate(&config, fmt.Sprintf("%s/clusterapi-manager-cluster-role.yaml", ownedManifestsDir))
```
frobware: Use https://golang.org/pkg/path/#Join for creating paths.

pkg/operator/sync.go
```diff
@@ -73,7 +72,7 @@ func (optr *Operator) syncClusterAPIController(config render.OperatorConfig) err
 	if err != nil {
 		return err
 	}
-	crbBytes, err := render.PopulateTemplate(&config, fmt.Sprintf("%s/clusterapi-manager-cluster-role-binding.yaml", ownedManifestsDir))
+	crbBytes, err := PopulateTemplate(&config, fmt.Sprintf("%s/clusterapi-manager-cluster-role-binding.yaml", ownedManifestsDir))
```
frobware: Use https://golang.org/pkg/path/#Join for creating paths.

pkg/operator/sync.go
```diff
@@ -82,7 +81,7 @@ func (optr *Operator) syncClusterAPIController(config render.OperatorConfig) err
 	if err != nil {
 		return err
 	}
-	controllerBytes, err := render.PopulateTemplate(&config, fmt.Sprintf("%s/clusterapi-manager-controllers.yaml", ownedManifestsDir))
+	controllerBytes, err := PopulateTemplate(&config, fmt.Sprintf("%s/clusterapi-manager-controllers.yaml", ownedManifestsDir))
```
frobware: Use https://golang.org/pkg/path/#Join for creating paths.

tests/e2e/main.go
```diff
@@ -254,29 +253,17 @@ var rootCmd = &cobra.Command{
 			}
 		}
 
-		// generate mao config
-		maoConfigTemplateData, err := ioutil.ReadFile(filepath.Join(assetsPath, "manifests/mao-config.yaml"))
+		installConfigData, err := ioutil.ReadFile(filepath.Join(assetsPath, "manifests/install-config.yaml"))
```
frobware: Use https://golang.org/pkg/path/#Join for all elements. Here's there's an embedded `/`.

pkg/operator/config.go
frobware: Why doesn't CI pick this up? @paulfantom

pkg/operator/sync.go
```diff
@@ -36,7 +35,7 @@ func (optr *Operator) syncAll(rconfig render.OperatorConfig) error {
 	})
 }
 
-func (optr *Operator) syncCustomResourceDefinitions(config render.OperatorConfig) error {
+func (optr *Operator) syncCustomResourceDefinitions(config OperatorConfig) error {
 	crds := []string{
 		"/machine.crd.yaml",
```
frobware: Is there a reason we prepend `/` and not just add those in via `path.Join()` when needed. It seems overloaded here.

pkg/operator/config.go
enxebre: indeed, is gofmt broken on CI? cc @paulfantom

pkg/operator/sync.go
```diff
@@ -37,14 +38,14 @@ func (optr *Operator) syncAll(config OperatorConfig) error {
 
 func (optr *Operator) syncCustomResourceDefinitions(config OperatorConfig) error {
 	crds := []string{
-		"/machine.crd.yaml",
-		"/machineset.crd.yaml",
-		"/machinedeployment.crd.yaml",
-		"/cluster.crd.yaml",
+		"machine.crd.yaml",
```
frobware: +1

tests/e2e/main.go
```diff
@@ -224,7 +224,7 @@ var rootCmd = &cobra.Command{
 		apiextensionsscheme.AddToScheme(scheme.Scheme)
 		decode := scheme.Codecs.UniversalDeserializer().Decode
 		// create rbac
-		if rbacData, err := ioutil.ReadFile(filepath.Join(assetsPath, "manifests/rbac.yaml")); err != nil {
+		if rbacData, err := ioutil.ReadFile(filepath.Join(assetsPath, "manifests", "rbac.yaml")); err != nil {
```
frobware: Create a utility function:

    func readManifestFile(filename string) ... {
        ... ioutil.ReadFile(filepath.Join(assetsPath, "manifests", filename)
        ...
    }

that way you don't have to worry about typos in "manifests" and easier to change in place should the location change over time.

tests/e2e/main.go
```diff
@@ -287,7 +287,7 @@ var rootCmd = &cobra.Command{
 		}
 
 		// create aws creds secret
-		if secretData, err := ioutil.ReadFile(filepath.Join(assetsPath, "manifests/aws-credentials.yaml")); err != nil {
+		if secretData, err := ioutil.ReadFile(filepath.Join(assetsPath, "manifests", "aws-credentials.yaml")); err != nil {
 			glog.Fatalf("Error reading %#v", err)
```
frobware: Why is this a `Fatalf` error?

tests/e2e/main.go
```diff
@@ -301,7 +301,7 @@ var rootCmd = &cobra.Command{
 		}
 
 		// create ign config secret
-		if secretData, err := ioutil.ReadFile(filepath.Join(assetsPath, "manifests/ign-config.yaml")); err != nil {
+		if secretData, err := ioutil.ReadFile(filepath.Join(assetsPath, "manifests", "ign-config.yaml")); err != nil {
 			glog.Fatalf("Error reading %#v", err)
```
frobware: Why is this a Fatalf error?

Can we not return `err` instead?

enxebre:
Agreed but that's out of the scope for this feature changing PR. I'll create a different one for that

frobware:
/approve

Leaving for somebody else to take a look through for lgtm. @ingvagabund

pkg/operator/config.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// InstallConfig contains the mao relevant config coming from the install config, i.e provider
+type InstallConfig struct {
+	IntallPlatform `json:"platform"`
+}
+
+// IntallPlatform is the configuration for the specific platform upon which to perform
+// the installation. Only one of the platform configuration should be set
+type IntallPlatform struct {
```
ingvagabund: s/IntallPlatform/InstallPlatform/

pkg/operator/config.go
```diff
@@ -0,0 +1,163 @@
...
+	for i := 0; i < v.NumField(); i++ {
+		if v.Field(i).Interface() != nil {
+			nonNilFields = nonNilFields + 1
+		}
+		if nonNilFields > 1 {
+			return "", fmt.Errorf("more than one platform provider given")
+		}
+	}
+
+	if installConfig.AWS != nil {
```
ingvagabund: This way AWS > Libvirt > OpenStack. What happens in case one specifies both AWS and Libvirt? Maybe add a check there can be only one provider set?

enxebre: that's checked here https://github.com/openshift/machine-api-operator/pull/106/files#diff-614eabe934833da7d7ec9938dfb3f1faR97 adn unit tested here https://github.com/openshift/machine-api-operator/pull/106/files#diff-b77facdf2f36a2cc873ec9ea3fae6b00R108

PR #112, opened 2018-11-02
Nodelink controller

enxebre:
Includes the nodelink controller in the controllers manifests
Move the nodelink controller to glog: `--log-level is gone` and `--logtostderror=true` is used instead.

### DIFF
--- a/owned-manifests/clusterapi-manager-controllers.yaml
+++ b/owned-manifests/clusterapi-manager-controllers.yaml
@@ -37,7 +37,7 @@ spec:
         operator: Exists
       containers:
       - name: controller-manager
-        image: {{ .Controller }}
+        image: {{ .Controllers.Provider }}
         command:
         - "./manager"
         resources:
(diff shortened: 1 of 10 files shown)


enxebre:
cc @ingvagabund @vikaschoudhary16

ingvagabund:
It fails with:
```
time="2018-11-02T12:40:41Z" level=info msg="Waiting for all clusterapi-manager-controllers deployment pods to be ready, have 0, expecting 1"
```

owned-manifests/clusterapi-manager-controllers.yaml
```diff
@@ -59,6 +59,13 @@ spec:
         args:
           - --logtostderr=true
           - --v=3
+      - name: nodelink-controller
+        image: {{ .Controller }}
```
ingvagabund: Oh, we need to use the mao docker image. The `nodelink-controller` is built as part of it.

PR #118, opened 2018-11-14
introduce remediation

enxebre:
introduce remediation logic:
- if machine belongs to healthCheck CRD then remediate
- Remediate:
   - if it is master skip
   - if is not owned by a machineSet skip
   - if it is recently unhealthy requeue
   - if it is unhealthy for too long delete

### DIFF
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -150,6 +150,10 @@ func getMachineHealthCheckListOptions() *client.ListOptions {
 
 func remediate(r *ReconcileMachineHealthCheck, machine *capiv1.Machine) (reconcile.Result, error) {
 	glog.Infof("Initialising remediation logic for machine %s", machine.Name)
+	if isMaster(*machine, r.client) {
+		glog.Info("The machine %s is a master node, skipping remediation", machine.Name)
+		return reconcile.Result{}, nil
+	}
 	if !hasMachineSetOwner(*machine) {
 		glog.Info("Machine %s has no machineSet controller owner, skipping remediation", machine.Name)
 		return reconcile.Result{}, nil
@@ -178,7 +182,7 @@ func remediate(r *ReconcileMachineHealthCheck, machine *capiv1.Machine) (reconci
 			currentTime := time.Now()
 			lastTimeHealthy := lastTransitionTime(node, healthCriteria)
 			durationUnhealthy := currentTime.Sub(lastTimeHealthy.Time)
-			glog.Warningf("machine %s have had node %s unhealthy for %s", machine.Name, node.Name, durationUnhealthy.String())
+			glog.Warningf("machine %s have had node %s unhealthy for %s. Requeuing...", machine.Name, node.Name, durationUnhealthy.String())
 			return reconcile.Result{Requeue: true, RequeueAfter: remediationWaitTime}, nil
 		}
 	}
@@ -260,4 +264,34 @@ func hasMatchingLabels(machineHealthCheck *healthcheckingv1alpha1.MachineHealthC
 		return false
 	}
 	return true
+}
+
+func isMaster(machine capiv1.Machine, client client.Client) bool {
+	machineMasterLabels := []string{
+		"sigs.k8s.io/cluster-api-machine-role",
+		"sigs.k8s.io/cluster-api-machine-type",
+	}
+	nodeMasterLabels := []string{
+		"node-role.kubernetes.io/master",
+	}
+
+	machineLabels := labels.Set(machine.Labels)
+	for _, masterLabel := range machineMasterLabels {
+		if machineLabels.Get(masterLabel) == "master" {
+			return true
+		}
+	}
+
+	node, err := getNodeFromMachine(machine, client)
+	if err != nil {
+		glog.Warningf("Couldn't get node for machine %s", machine.Name)
+		return false
+	}
+	nodeLabels := labels.Set(node.Labels)
+	for _, masterLabel := range nodeMasterLabels {
+		if nodeLabels.Has(masterLabel) {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -3,6 +3,7 @@ package machinehealthcheck
 import (
 	"context"
 
+	golangerrors "errors"
 	"github.com/golang/glog"
 	healthcheckingv1alpha1 "github.com/openshift/machine-api-operator/pkg/apis/healthchecking/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -140,18 +146,93 @@ func getMachineHealthCheckListOptions() *client.ListOptions {
...
+	if !hasMachineSetOwner(*machine) {
+		glog.Info("Machine %s has no machineSet controller owner, skipping remediation", machine.Name)
+		return reconcile.Result{}, err
+	}
+
 	if !isHealthy(node) {
+		if unhealthyForTooLong(node) {
+			// delete
+			glog.Infof("machine %s has been unhealthy for too long, deleting", machine.Name)
+			if err := r.client.Delete(context.TODO(), machine); err != nil {
```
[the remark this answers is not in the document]
enxebre: We are adding support for that at the actuator level so no api consumer can delete masters. I'll make sure this is also implemented and unit tested here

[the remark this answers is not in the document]
enxebre: added https://github.com/openshift/machine-api-operator/pull/118/commits/771e6f416339c7f556085a82f9dd1e8bd1e1636f

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -120,14 +127,15 @@ func (r *ReconcileMachineHealthCheck) Reconcile(request reconcile.Request) (reco
 	for _, hc := range allMachineHealthChecks.Items {
 		if hasMatchingLabels(&hc, machine) {
 			glog.V(4).Infof("Machine %s has a matching machineHealthCheck: %s", machineKey, hc.Name)
-			remediate(node)
+			return remediate(r, machine)
```
frobware: We're deliberating ignoring the rest of `range allMachineHealthChecks.Items` - yes? if so, why don't the other items come into play?

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -265,19 +284,193 @@ func TestReconcile(t *testing.T) {
 		}
 		result, err := r.Reconcile(request)
 		if result != tc.expected.result {
-			t.Errorf("Expected %v, got: %v", tc.expected.result, result)
+			t.Errorf("Test case : %v. Expected: %v, got: %v", tc.node.Name, tc.expected.result, result)
```
frobware: nit: `"Test case:` (i.e., no space) like the formatting in `TestGetNodeCondition`.

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -265,19 +284,193 @@ func TestReconcile(t *testing.T) {
...
 		}
 		if tc.expected.error != (err != nil) {
-			t.Errorf("Expected error, got %v", err)
+			var errorExpectation string
+			if !tc.expected.error == true {
```
frobware: Drop `== true` and just use as the implicit boolean expression.

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -265,19 +284,193 @@ func TestReconcile(t *testing.T) {
...
-func newFakeReconciler(initObjects ...runtime.Object) reconcile.Reconciler {
+func newFakeReconciler(initObjects ...runtime.Object) *ReconcileMachineHealthCheck {
 	fakeClient := fake.NewFakeClient(initObjects...)
 	return &ReconcileMachineHealthCheck{
 		client: fakeClient,
 		scheme: scheme.Scheme,
 	}
 }
+
+func TestHasMachineSetOwner(t *testing.T) {
```
frobware: nit: wording: `TestMachineSetHasOwner`?

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -265,19 +284,193 @@ func TestReconcile(t *testing.T) {
...
+
+	testsCases := []struct {
+		machine  *capiv1alpha1.Machine
+		expected expectedReconcile
+	}{
+		{
+			machine: machineWithNodeHealthy,
+			expected: expectedReconcile{
+				result: reconcile.Result{},
+				error:  false,
```
frobware: I would advocate using the "zero" value here, and for the other cases that default to `false`. Omitting this entry will default to `false`.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -261,3 +265,33 @@ func hasMatchingLabels(machineHealthCheck *healthcheckingv1alpha1.MachineHealthC
...
+	machineLabels := labels.Set(machine.Labels)
+	for _, masterLabel := range machineMasterLabels {
+		if machineLabels.Get(masterLabel) == "master" {
+			return true
+		}
+	}
+
+	node, err := getNodeFromMachine(machine, client)
+	if err != nil {
+		glog.Warningf("Could't get node for machine %s", machine.Name)
```
frobware: typo: "Couldn't"

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -474,3 +475,57 @@ func TestRemediate(t *testing.T) {
...
+		machine  *capiv1alpha1.Machine
+		expected bool
+	}{
+		{
+			machine:  masterMachine,
+			expected: true,
+		},
+		{
+			machine:  workerMachine,
+			expected: false,
```
frobware: Can drop this; `false` is the "zero" value.

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
enxebre: Actually is right, is testing if "it has a machineSet owner"? happy to reword if you have a better suggestion

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
enxebre: The initial approach is simple. A MachineHealthCheck CRD is not configurable, it just select machines based on labels, if there's a least one matching a machine it should be remediated.

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
frobware: ok

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
enxebre: thanks

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
enxebre: I'd rather keep it explicitly for the test cases, it makes it easier to visualise and understand the expectation when looking at it

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
enxebre: thanks thanks

PR #124, opened 2018-11-19
Nodelinktest

enxebre:
Add ExpectAllMachinesLinkedToANode test
Needs https://github.com/openshift/machine-api-operator/pull/123

### DIFF
--- a/test/e2e/operator_expectations.go
+++ b/test/e2e/operator_expectations.go
@@ -5,9 +5,11 @@ import (
 
 	"context"
 	"errors"
+	"fmt"
 	"github.com/golang/glog"
 	osv1 "github.com/openshift/cluster-version-operator/pkg/apis/operatorstatus.openshift.io/v1"
 	kappsapi "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	capiv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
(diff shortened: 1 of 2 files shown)


test/e2e/operator_expectations.go
```diff
@@ -75,3 +77,47 @@ func ExpectOperatorStatusConditionDone() error {
...
+func ExpectAllMachinesLinkedToANode() error {
+	machineAnnotationKey := "machine"
+	listOptions := client.ListOptions{
+		Namespace: namespace,
+	}
+	machineList := capiv1alpha1.MachineList{}
+	nodeList := corev1.NodeList{}
+
+	err := wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
+		if err := F.Client.List(context.TODO(), &listOptions, &machineList); err != nil {
```
spangenberg: Why `context.TODO()` do you plan to change it in the future to an injected context?

test/e2e/operator_expectations.go
```diff
@@ -75,3 +77,47 @@ func ExpectOperatorStatusConditionDone() error {
...
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		return err
+	}
+
+	err = wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
+		if err := F.Client.List(context.TODO(), &listOptions, &nodeList); err != nil {
```
spangenberg: Same as above.

test/e2e/operator_expectations.go
ingvagabund: It has been done this way for a long time :-|

spangenberg: Meh, we should rather use Background. IMHO `go vet` should mark that.

PR #126, opened 2018-11-20
Remove unused clients and informers

enxebre:
Remove unused clients and informers:
- No need for ctx.KubeInformerFactory.Core().V1().ConfigMaps() since only config is the one time provider got from installer config
- Remove old apiregistrationclientset and clusterapiclientset

Increase minResyncPeriod

### DIFF
--- a/cmd/common/helpers.go
+++ b/cmd/common/helpers.go
@@ -22,7 +22,7 @@ const (
 	// RetryPeriod is the default duration for the leader electrion retrial.
 	RetryPeriod = 30 * time.Second
 
-	minResyncPeriod = 10 * time.Second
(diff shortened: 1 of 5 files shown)


cmd/common/helpers.go
```diff
@@ -22,7 +22,7 @@ const (
 	// RetryPeriod is the default duration for the leader electrion retrial.
 	RetryPeriod = 30 * time.Second
 
-	minResyncPeriod = 10 * time.Second
+	minResyncPeriod = 10 * time.Minute
```
bison: This seems high. Are we sure we want 10 minutes?

PR #127, opened 2018-11-20
Crds install

enxebre:
Moves CRDs creation out of the operator to the CVO
Needs https://github.com/openshift/machine-api-operator/pull/126


bison:
Looks like we'll need to update the integration tests?

This mostly looks fine. Might be a little more readable with a `crdFromYAML()` helper in the tests or something. Not a big deal though.

enxebre:
> Looks like we'll need to update the integration tests?
> 
> This mostly looks fine. Might be a little more readable with a `crdFromYAML()` helper in the tests or something. Not a big deal though.

Yeh will either polish or just drop "integration" since its value is minimum now we have aws-e2e-operator in place

tests/e2e/main.go
```diff
@@ -240,10 +240,70 @@ var rootCmd = &cobra.Command{
 		}
 
 		// create status CRD
-		if statusCRD, err := ioutil.ReadFile(filepath.Join(assetsPath, manifestsFolder, "status-crd.yaml")); err != nil {
+		if CRDBytes, err := ioutil.ReadFile(filepath.Join(assetsPath, manifestsFolder, "status-crd.yaml")); err != nil {
 			glog.Fatalf("Error reading %#v", err)
 		} else {
-			CRDObj, _, err := decode([]byte(statusCRD), nil, nil)
+			CRDObj, _, err := decode([]byte(CRDBytes), nil, nil)
```
spangenberg: CRDBytes is already a byte array, no need for conversion here.

tests/e2e/main.go
```diff
@@ -240,10 +240,70 @@ var rootCmd = &cobra.Command{
 		}
 
 		// create status CRD
-		if statusCRD, err := ioutil.ReadFile(filepath.Join(assetsPath, manifestsFolder, "status-crd.yaml")); err != nil {
+		if CRDBytes, err := ioutil.ReadFile(filepath.Join(assetsPath, manifestsFolder, "status-crd.yaml")); err != nil {
```
spangenberg: Would prefer crdBytes

tests/e2e/main.go
enxebre: this is converting bytes into a runtime.Object them cast to a specfic api object

tests/e2e/main.go
enxebre: We use consistently upper case 
 for acronyms https://github.com/golang/go/wiki/CodeReviewComments#initialisms

tests/e2e/main.go
spangenberg: `ioutil.ReadFile` returns `[]byte` and decode accepts `[]byte` so no need for `[]byte(CRDBytes)`

tests/e2e/main.go
spangenberg: ok

tests/e2e/main.go
enxebre: oh right, seems we are doing the same everywhere in integration. I will either polish or maybe just drop "integration" in a follow up since its value is minimum now we have aws-e2e-operator in place.
If there's no other concern I'd suggest to get this in so I can put a few follow ups in place blocked by this one

spangenberg: sure sounds good

PR #128, opened 2018-11-21
Cluster operator status

enxebre:
Move from old operatorstatus.openshift.io to Cluster Operator object


pkg/operator/status.go
```diff
@@ -1,61 +1,49 @@
...
-	if cond.Type == v1.OperatorStatusConditionTypeDegraded {
-		return fmt.Errorf("invalid cond %s", cond.Type)
-	}
-
-	// TODO(yifan): Fill in the Extention field for the status
+//syncStatus applies the new condition to the mao ClusterOperator object.
+func (optr *Operator) syncStatus(cond osconfigv1.ClusterOperatorStatusCondition) error {
 	// to report the status of all the managed components.
-	status := &v1.OperatorStatus{
+	clusterOperator := &osconfigv1.ClusterOperator{
```
ingvagabund: What's the relation between mao operator status and cluster operator status? Is cluster operator an abstract way how to update status? Or WIP at the moment? If anything, I would expect `osconfigv1.Operator` struct instead for generic status updates.

Hmm, mao cluster operator. What do I need to read to understand why is the `cluster` part of mao operator name? Assuming it means a cluster operator is a name for a class of operators, right? On the cluster level?

pkg/operator/sync.go
```diff
@@ -39,8 +40,18 @@ func (optr *Operator) syncAll(config OperatorConfig) error {
 	}
 	glog.Info("Synched up cluster api controller")
 
-	return optr.syncStatus(v1.OperatorStatusCondition{
-		Type:    v1.OperatorStatusConditionTypeDone,
+	// TODO(alberto): create funcs to for atomic conds changes
```
ingvagabund: You can make the `syncStatus` to accept variable number of arguments, resp. a list with variable size.

pkg/operator/status.go
```diff
@@ -1,61 +1,49 @@
...
 
-// syncStatus applies the new condition to the mco's OperatorStatus object.
-func (optr *Operator) syncStatus(cond v1.OperatorStatusCondition) error {
-	if cond.Type == v1.OperatorStatusConditionTypeDegraded {
-		return fmt.Errorf("invalid cond %s", cond.Type)
-	}
-
-	// TODO(yifan): Fill in the Extention field for the status
+//syncStatus applies the new condition to the mao ClusterOperator object.
+func (optr *Operator) syncStatus(cond osconfigv1.ClusterOperatorStatusCondition) error {
```
ingvagabund: Not sure if this is going to work but worth trying :) 
```go
func (optr *Operator) syncStatus(statusConditions ...osconfigv1.ClusterOperatorStatusCondition) error {
	// to report the status of all the managed components.
	clusterOperator := &osconfigv1.ClusterOperator{
		ObjectMeta: metav1.ObjectMeta{
			Namespace: optr.namespace,
			Name:      optr.name,
		},
		Status: osconfigv1.ClusterOperatorStatus{
			Version: version.Raw,
		},
	}
	for _, cond := range statusConditions {
		cvoresourcemerge.SetOperatorStatusCondition(&clusterOperator.Status.Conditions, cond)
	}
	_, _, err := ApplyClusterOperator(optr.osClient.ConfigV1(), clusterOperator)
	return err
}
```

pkg/operator/sync.go
enxebre: Yeh, this is a first step to have the right CRD object in place intentionally simple, hence the TODO. Once we get clarity on the CVO expectations for updgrades we can model some functions

ingvagabund:
Ok, fair enough
/approve

PR #139, opened 2018-11-26
WIP - Explore library-go for openshift/machine-api-operator

derekwaynecarr:
Explores the adoption of openshift/library-go for the machine-api-operator since no formal framework is used at this time.

/cc @deads2k


derekwaynecarr:
A few teams have been evaluating operator-sdk versus library-go to build their second level operator.

The machine-api-operator uses no formal framework.  I used this as an exercise to evaluate library-go relative to the current code to provide an informed recommendation.

Using library-go provides the following:

1. common make scripts that mostly just work for operator style repositories
   - useful to avoid potential future maintenance
1. pattern for operator development that matches many cluster-kube-* and cluster-openshift-* operators
1. uses bindata by convention for managing assets
1. pattern for workload controller that
   - watches subset of cluster config sources and syncs output to match
   - key thing i like is its use of generation to know that it should reconcile a change
1. assumes the existence of a CR to manage operator output
   - the machine-api-operator today is a 'naked' operator
   - it just has a stamp of a deployment it renders out with fixed resources
   - i do think team should explore driving the rendered asset from a singleton CR that lets customers change logging and resource configuration for component from the CR

I had to remove two pieces of function to get in a compiling state as moving to `glide` made this much easier to look at than trying to keep `dep` happy.  I would like to get team's perspective to know if this is worth pursuing further.  It probably requires another couple days to get back to fully functional state.

derekwaynecarr:
reflecting on this a little more it’s also possible that being a naked operator with no config is possibly enough for 4.0 initial release.  config can always get added later.

PR #142, opened 2018-11-29
Document how to test the healthchecking mechanism

ingvagabund:
Just demonstrate what is needed to test the machine health-checking mechanism.


ingvagabund:
@enxebre anything else in mind that needs to be mentioned?

PR #147, opened 2018-12-07
replace tectonicClusterID tag with openshiftClusterID

staebler:
The tectonicClusterID tag is being replaced with openshiftClusterID in the installer. This change will use the new tag name.

/hold on https://github.com/openshift/installer/pull/817


ingvagabund:
per comments in https://github.com/openshift/installer/pull/817#issue-236729725:
> As part of removing references to tectonic, the tectonicClusterID tags will
be changed to openshiftClusterID. This is the first part of that.

> This creates the openshiftClusterID tag that will be laid down alongside the
tectonicClusterID tag.

/lgtm
/approve

PR #148, opened 2018-12-10
machineset: add additional printer columns to CRD

frobware:
Additionally print:

- Desired Replicas
- Current Replicas
- Ready Replicas

which is similar to the replicaset output.

If you're going to watch output then be wary of:

https://github.com/kubernetes/kubernetes/issues/66538

### DIFF
--- a/install/0000_50_machine-api-operator_03_machineset.crd.yaml
+++ b/install/0000_50_machine-api-operator_03_machineset.crd.yaml
@@ -94,6 +94,22 @@ spec:
           - replicas
           type: object
   version: v1alpha1
+  additionalPrinterColumns:
+  - JSONPath: .spec.replicas
+    description: Desired Replicas
+    name: Desired
+    type: integer
+  - JSONPath: .status.replicas
+    description: Current Replicas
+    name: Current
+    type: integer
+  - JSONPath: .status.readyReplicas
+    description: Ready Replicas
+    name: Ready
+    type: integer
+  - JSONPath: .metadata.creationTimestamp
+    name: Age
+    type: date
 status:
   acceptedNames:
     kind: ""
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -151,11 +151,11 @@ func getMachineHealthCheckListOptions() *client.ListOptions {
 func remediate(r *ReconcileMachineHealthCheck, machine *capiv1.Machine) (reconcile.Result, error) {
 	glog.Infof("Initialising remediation logic for machine %s", machine.Name)
 	if isMaster(*machine, r.client) {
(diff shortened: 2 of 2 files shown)


ingvagabund:
@frobware can you make sure we distribute the change to aws-actuator [1] as well?

[1] https://github.com/openshift/cluster-api-provider-aws/blob/master/config/crds/machineset.crd.yaml

/lgtm
/approve

ingvagabund:
Before applying the change:
```
NAMESPACE   NAME                                AGE
test        jchaloup-worker-machineset-de2720   3m
```
After:

```
# kubectl get machinesets --all-namespaces
NAMESPACE   NAME                                DESIRED   CURRENT   READY
test        jchaloup-worker-machineset-de2720   1         1         
```

frobware:
Corresponding PR for cluster-api-provider-aws: https://github.com/openshift/cluster-api-provider-aws/pull/124

install/0000_50_machine-api-operator_03_machineset.crd.yaml
```diff
@@ -94,9 +94,25 @@ spec:
...
   version: v1alpha1
+  additionalPrinterColumns:
+  - JSONPath: .spec.replicas
+    description: Desired Replicas
+    name: Desired
+    type: string
+  - JSONPath: .status.replicas
+    description: Current Replicas
+    name: Current
+    type: string
```
bison: Should these not be `type: integer`?

ingvagabund: Good point. The docs [1] says:
> If the value inside a CustomResource does not match the type specified for the column, the value is omitted. Use CustomResource validation to ensure that the value types are correct.

So wondering why is it printed even though string != int. Maybe a type cast that anything can be casted to string (e.g. `fmt.Sprintf("...%v", columnValue)`)? @frobware did you notice any difference?

[1] https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#type

frobware: I didn't notice any difference. My choice was influenced by the replica set TableColumnDefinitions. But let me experiment with switching to integer. 

Currently, with `string`, I see: 

```console
$ kubectl get machinesets  --all-namespaces 
NAMESPACE   NAME                                             DESIRED   CURRENT   READY     AGE
test        amcdermo-aws-actuator-worker-machineset-69e4fb   3         3         3         13m
```

frobware: I switched to integers and squashed commits in https://github.com/openshift/machine-api-operator/pull/148/commits/d49b90530c5acb6f4a6a81232b1f2052d0582523

frobware:
> Missing `Age` column, please re-introduce it back.

Done.

PR #154, opened 2018-12-20
Sync taints from machine spec to the node in additive fashion

vikaschoudhary16:
/cc @derekwaynecarr @enxebre

### DIFF
--- a/cmd/nodelink-controller/main.go
+++ b/cmd/nodelink-controller/main.go
@@ -393,8 +393,7 @@ func (c *Controller) processNode(node *corev1.Node) error {
 		modNode.Labels[k] = v
 	}
 
-	// Taints are to be an authoritative list on the machine spec per cluster-api comments:
-	modNode.Spec.Taints = matchingMachine.Spec.Taints
+	addTaintsToNode(modNode, matchingMachine)
 
 	if !reflect.DeepEqual(node, modNode) {
 		glog.V(3).Infof("node has changed, updating")
@@ -405,6 +404,26 @@ func (c *Controller) processNode(node *corev1.Node) error {
 		}
 	}
 	return nil
+}
+
+// addTaintsToNode adds taints from machine object to the node object
+// Taints are to be an authoritative list on the machine spec per cluster-api comments.
+// However, we believe many components can directly taint a node and there is no direct source of truth that should enforce a single writer of taints
+func addTaintsToNode(node *corev1.Node, machine *capiv1.Machine) {
+	for _, mTaint := range machine.Spec.Taints {
+		glog.V(3).Infof("machine taint: %v", mTaint)
+		alreadyPresent := false
+		for _, nTaint := range node.Spec.Taints {
+			if nTaint.Key == mTaint.Key && nTaint.Effect == mTaint.Effect {
+				glog.V(3).Infof("Skipping to add machine taint, %v, to the node. Node already has a taint with same key and effect", mTaint)
+				alreadyPresent = true
+				break
+			}
+		}
+		if !alreadyPresent {
+			node.Spec.Taints = append(node.Spec.Taints, mTaint)
+		}
+	}
 }
 
 var (
--- a/test/e2e/operator_expectations.go
+++ b/test/e2e/operator_expectations.go
@@ -12,6 +12,7 @@ import (
 	kappsapi "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	capiv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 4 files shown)


cmd/nodelink-controller/main.go
```diff
@@ -393,8 +393,22 @@ func (c *Controller) processNode(node *corev1.Node) error {
 		modNode.Labels[k] = v
 	}
 
-	// Taints are to be an authoritative list on the machine spec per cluster-api comments:
```
enxebre: Nice!
Can you add a PR upstream even if it's to update the field comment and why we think should be additive?
Can you please add a validation for this machine/node taint reconciliation here https://github.com/openshift/machine-api-operator/blob/master/test/e2e/operator_expectations.go?

derekwaynecarr:
Change looks good.   Link to upstream feature for why it cannot be authoritative.  See https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/#taint-nodes-by-condition

cmd/nodelink-controller/main.go
```diff
@@ -393,8 +393,22 @@ func (c *Controller) processNode(node *corev1.Node) error {
...
-	modNode.Spec.Taints = matchingMachine.Spec.Taints
+	// Taints are to be an authoritative list on the machine spec per cluster-api comments.
+	// However, we believe many components can directly taint a node and there is no direct source of truth that should enforce a single writer of taints
+	for _, mTaint := range matchingMachine.Spec.Taints {
```
enxebre: It'd be nice to create a function and unit test it

cmd/nodelink-controller/main.go
```diff
@@ -393,8 +393,22 @@ func (c *Controller) processNode(node *corev1.Node) error {
...
+		alreadyPresent := false
+		for _, nTaint := range modNode.Spec.Taints {
+			if nTaint.Key == mTaint.Key && nTaint.Effect == mTaint.Effect {
+				glog.V(3).Infof("Skipping to add machine taint, %v, to the node. Node already has a taint with same key and effect", mTaint)
+				alreadyPresent = true
+				break
+			}
+		}
+		if !alreadyPresent {
+			modNode.Spec.Taints = append(modNode.Spec.Taints, mTaint)
```
enxebre: we are appending here when `nTaint.Key == mTaint.Key && nTaint.Effect != mTaint.Effect`, what will be the behaviour?

vikaschoudhary16: This is list of taints and for a taint to be unique key and effect as a pair must be unique. PTAL at example here:
https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/#concepts

cmd/nodelink-controller/main.go
vikaschoudhary16: sure. how about handling e2e in  follow-up PR?

cmd/nodelink-controller/main.go
vikaschoudhary16: sure.

cmd/nodelink-controller/main.go
enxebre: please let's include it in this one, the point of having it in the same PR is that it prevent us from merging code which does not work as expected rather than having to fix later

cmd/nodelink-controller/main.go
```diff
@@ -407,6 +406,26 @@ func (c *Controller) processNode(node *corev1.Node) error {
...
 
+// addTaintsToNode adds taints from machine object to the node object
+// Taints are to be an authoritative list on the machine spec per cluster-api comments.
+// However, we believe many components can directly taint a node and there is no direct source of truth that should enforce a single writer of taints
+func addTaintsToNode(node *corev1.Node, machine *capiv1.Machine) {
+	for _, mTaint := range machine.Spec.Taints {
+		glog.V(3).Infof("machine taint: %v", mTaint)
+		alreadyPresent := false
+		for _, nTaint := range node.Spec.Taints {
+			if nTaint.Key == mTaint.Key && nTaint.Effect == mTaint.Effect {
```
enxebre: Why don't we want to take into consideration the `.value` here?

test/e2e/operator_expectations.go
```diff
@@ -175,6 +176,75 @@ func (tc *testConfig) ExpectReconcileControllersDeployment() error {
...
 }
 
+func (tc *testConfig) ExpectAdditiveReconcileMachineTaints() error {
+	glog.Info("Verify machine taints are getting applied to node")
+	listOptions := client.ListOptions{
+		Namespace: namespace,
+	}
+	machineList := capiv1alpha1.MachineList{}
+
+	if err := tc.client.List(context.TODO(), &listOptions, &machineList); err != nil {
```
enxebre: this should be wrapped into `wait.PollImmediate` 186-195

test/e2e/operator_expectations.go
```diff
@@ -175,6 +176,75 @@ func (tc *testConfig) ExpectReconcileControllersDeployment() error {
...
+	}
+	glog.Infof("Got the machine, %s", machine.Name)
+	nodeName := machine.Status.NodeRef.Name
+	nodeKey := types.NamespacedName{
+		Namespace: namespace,
+		Name:      nodeName,
+	}
+	node := &corev1.Node{}
+
+	if err := tc.client.Get(context.TODO(), nodeKey, node); err != nil {
```
enxebre: this should be wrapped into `wait.PollImmediate`

test/e2e/operator_expectations.go
```diff
@@ -175,6 +176,75 @@ func (tc *testConfig) ExpectReconcileControllersDeployment() error {
...
+		Value:  "true",
+		Effect: corev1.TaintEffectNoSchedule,
+	}
+	machine.Spec.Taints = []corev1.Taint{machineTaint}
+	if err := tc.client.Update(context.TODO(), &machine); err != nil {
+		return fmt.Errorf("error updating machine object with taint: %v", err)
+	}
+	glog.Info("Updated machine object with taint")
+	var expectedTaints = sets.NewString("not-from-machine", "from-machine")
+	err := wait.PollImmediate(1*time.Second, waitLong, func() (bool, error) {
```
enxebre: we could just return inline here

test/e2e/operator_expectations.go
```diff
@@ -175,6 +176,75 @@ func (tc *testConfig) ExpectReconcileControllersDeployment() error {
...
+		if err := tc.client.Get(context.TODO(), nodeKey, node); err != nil {
+			glog.Errorf("error querying api for node object: %v", err)
+			return false, nil
+		}
+		glog.Info("Got the node again for verification of taints")
+		var observedTaints = sets.NewString()
+		for _, taint := range node.Spec.Taints {
+			observedTaints.Insert(taint.Key)
+		}
+		if expectedTaints.Difference(observedTaints).HasAny("not-from-machine", "from-machine") == false {
```
frobware: Use `!<expression>` and not `== false`.

PR #158, opened 2019-01-03
Remove old labels

enxebre:
This labels are carry over from a tectonic context, they are not needed now


ingvagabund:
What is equivalent for `managed-by-channel-operator: "true"` currently? :P

enxebre:
there is no equivalent

PR #159, opened 2019-01-03
Add logtostderr flag for controller-manager

enxebre:
Let's enable this for the sake of better visibility, specially since we are introducing support for machine deployments


ingvagabund:
/lgtm
both flags are already used in the CI, seems fine to mirror it here as well

PR #164, opened 2019-01-07
Deployment expect

enxebre:
e2e test for deployment rolling updates.
- Scale to 3 replicas
- Set `maxUnavailable = 0` `maxSurge = 1`
- Trigger a rolling update by updating the machineType of a deployment
- Verify machines are rolled our one by one and each associated node goes available
- Verify expected final number of nodes

Needs https://github.com/openshift/installer/pull/990


test/e2e/utils.go
```diff
@@ -0,0 +1,31 @@
+package main
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	capiv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+)
+
+// IsNodeReady returns true if a node is ready; false otherwise.
+func IsNodeReady(node *corev1.Node) bool {
```
frobware: Is there a reason why we can't use the canonical implementation in: https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/node/util.go

frobware:
Your commit message would be really handy in the code too; it would explain what the test does and expects as success. I know it is here in the commit message but soon the PR will be long gone whereas that same description would be really handy at the beginning of the test function.

frobware:
> Your commit message would be really handy in the code too; it would explain what the test does and expects as success. I know it is here in the commit message but soon the PR will be long gone whereas that same description would be really handy at the beginning of the test function.

Put it this way, it became way more obvious reading the comment than grokking the code...

test/e2e/utils.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+	for _, c := range node.Status.Conditions {
+		if c.Type == corev1.NodeReady {
+			return c.Status == corev1.ConditionTrue
+		}
+	}
+	return false
+}
+
+// MachineSetsByCreationTimestamp sorts a list of MachineSet by creation timestamp, using their names as a tie breaker.
```
frobware: Comment doesn't match type.

test/e2e/operator_expectations.go
```diff
@@ -340,3 +344,188 @@ MachineLoop:
...
+		}
+		glog.Infof("Initial number of nodes: %d. Ready nodes: %d", clusterInitialTotalNodes, readyNodes)
+		return readyNodes == clusterInitialTotalNodes, nil
+	})
+	if err != nil {
+		return err
+	}
+
+	replicas := int32(3)
+	dep := machineDeploymentList.Items[0]
```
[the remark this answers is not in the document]
ingvagabund: `s/dep/machineDeploymen/`, `dep` smells like dependency.

test/e2e/operator_expectations.go
```diff
@@ -340,3 +344,188 @@ MachineLoop:
...
+		for _, node := range nodeList.Items {
+			if IsNodeReady(&node) {
+				readyNodes++
+			}
+		}
+		glog.Infof("Initial number of nodes: %d. Ready nodes: %d", expectedNumberOfNodes, readyNodes)
+		return readyNodes == expectedNumberOfNodes, nil
+
+	})
+	if err != nil {
```
ingvagabund: `if err := wait.PollImmediate...; err != nil {`

test/e2e/operator_expectations.go
```diff
@@ -340,3 +344,188 @@ MachineLoop:
...
+				glog.Errorf("error querying api for nodeList object: %v, retrying...", err)
+				return false, nil
+			}
+			glog.Infof("Wait for new node %s to become available", newNode.Name)
+			return IsNodeReady(newNode), nil
+		})
+		if err != nil {
+			return err
+		}
+	}
```
ingvagabund: You test new machines are created. However, I don't see any test verifying new instances are of `m4.large` flavor.

test/e2e/operator_expectations.go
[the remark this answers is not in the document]
enxebre: this is handled here now https://github.com/openshift/machine-api-operator/pull/164/commits/6722573abfd5ac85f0dfa78a8e04be310b3d3082#diff-7a5de36d018010fe027cb621ddf222dfR633

test/e2e/operator_expectations.go
enxebre: yes, we can add farther AWS API validation in follow ups once this settle

test/e2e/util.go
```diff
@@ -0,0 +1,17 @@
...
+type MachineByCreationTimestamp []capiv1alpha1.Machine
+
+func (o MachineByCreationTimestamp) Len() int      { return len(o) }
+func (o MachineByCreationTimestamp) Swap(i, j int) { o[i], o[j] = o[j], o[i] }
+func (o MachineByCreationTimestamp) Less(i, j int) bool {
+	if o[i].CreationTimestamp.Equal(&o[j].CreationTimestamp) {
+		return o[i].Name < o[j].Name
+	}
+	return o[i].CreationTimestamp.After(o[j].CreationTimestamp.Time)
+}
```
frobware: All of the swap functionality can be replaced and simplified by using `sort.Slice()`.

```go
sort.Slice(o, func(i, j int) bool {
	if o[i].CreationTimestamp.Equal(&o[j].CreationTimestamp) {
		return o[i].Name < o[j].Name
	}
	return o[i].CreationTimestamp.After(o[j].CreationTimestamp.Time)
})
```

enxebre: thanks! I didn't know that. If you are ok with it I'd like to keep the interface impl. rather than inline it since I find it more readable/reusable

frobware: I still think you would have less code. You can still create a wrapper func:

```go
func sortByCreationTimestamp() ... 
```

with the implementation I posted. For me this is win/win; using stdlib code (so no surprises) and reusable as it is encapsulated by a named function.

enxebre: changed, thanks

test/e2e/operator_expectations.go
ingvagabund: But, is not the that whole point of the test? You run rolling update changing flavor from one to another? The deployment can flip the machineset but that does not guarantee the new machines have different flavors.

enxebre: Added a validation for the new machine to come with the expected type https://github.com/openshift/machine-api-operator/pull/164/files#diff-7a5de36d018010fe027cb621ddf222dfR759
Provider specific validations and iteration is wider topic that I'd like to scope separately

enxebre:
closing until we reprioritise deployments support

PR #165, opened 2019-01-08
Add possibility to specify baremetal actuator under the install config

cynepco3hahue:
Adding bare-metal actuator, currently, bare-metal actuator only gives the possibility to fence bare metal nodes via IPMI interface(uses fence-agents), we still do not support provisioning of the bare-metal hosts.

TODO:
- [x] add unit tests

I can see that libvirt provider does not have e2e tests under the repository, I believe it due to the fact that current `installer` supports only `aws` deployment, I prefer to leave integration tests until we will baremetal support under the installer

KubeVirt project needs this PR to provide fencing logic on top of the OpenShift


cynepco3hahue:
Not relevant anymore.

PR #167, opened 2019-01-09
Revert "Support both providerConfig and providerSpec under machine CRD's"

ingvagabund:
Reverts openshift/machine-api-operator#155 and keeps only the providerSpec field.


enxebre:
before merging this, please make sure to update https://github.com/openshift/console/search?q=providerConfig&unscoped_q=providerConfig

PR #170, opened 2019-01-10
Drop deprecated providerConfig field

enxebre:
The machine provideConfig field has been renamed upstream kubernetes-sigs/cluster-api#548. AWS/Libvirt providers have been accommodated for supporting the new field openshift/cluster-api-provider-aws#127 and openshift/cluster-api-provider-libvirt#104 in a backward compatible manner https://github.com/openshift/machine-api-operator/pull/155, hence we want to only allow the new field here and eventually drop the old one on the providers.


ingvagabund:
looks good:
- `providerConfig` removed from all 4 machine crds 
- `providerConfig` renamed to `providerSpec` in `test/integration/manifests/machineset.yaml` 

/approve

PR #173, opened 2019-01-10
Crds drop old field

enxebre:
The machine provideConfig field has been renamed upstream kubernetes-sigs/cluster-api#548. AWS/Libvirt providers have been accommodated for supporting the new field openshift/cluster-api-provider-aws#127 and openshift/cluster-api-provider-libvirt#104 in a backward compatible manner #155, hence we want to only allow the new field here and eventually drop the old one on the providers.


ingvagabund:
Let's merge it before the CI starts to get flaky again.

PR #174, opened 2019-01-10
Set all expected ClusterOperator status conditions

bison:
The operator's syncStatus method has been setting a single condition
and removing others already present, but most components checking
status expect all three default conditions (Available, Progressing,
and Failing) to be present and set to either True or False.  This
changes the status reporting to report all three at once.


enxebre:
thanks! you might wanna update https://github.com/openshift/machine-api-operator/blob/master/test/e2e/operator_expectations.go#L67

PR #176, opened 2019-01-12
Add CLI printer columns for machine, machinesets

derekwaynecarr:
This adds pretty printers when rendering tables in CLI.

Machines
```sh
oc get machines -n openshift-cluster-api
NAME                                 INSTANCE              STATE     TYPE       REGION      ZONE         AGE
decarr-dev-master-0                  i-0c5c6028992720b04   running   m4.large   us-east-2   us-east-2a   4h
```

MachineSets
```sh
oc get machinesets -n openshift-cluster-api
NAME                           DESIRED   CURRENT   READY     AVAILABLE   AGE
decarr-dev-worker-us-east-2a   1         1         1         1           4h
```

### DIFF
--- a/install/0000_50_machine-api-operator_02_machine.crd.yaml
+++ b/install/0000_50_machine-api-operator_02_machine.crd.yaml
@@ -5,6 +5,30 @@ metadata:
     controller-tools.k8s.io: "1.0"
   name: machines.cluster.k8s.io
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.providerStatus.instanceId
+    name: Instance
+    description: Instance ID of machine created in AWS
+    type: string
+  - JSONPath: .status.providerStatus.instanceState
+    name: State
+    description: State of the AWS instance
+    type: string
+  - JSONPath: .spec.providerSpec.value.instanceType
+    name: Type
+    description: Type of instance
+    type: string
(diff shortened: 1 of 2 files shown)


derekwaynecarr:
It is inevitable that you will want printers for each provider.

The beta is targeting AWS, so this is good for now.  As we surface other providers, we may want to look at having the machine-api-operator itself install the CRDs that vary by provider so we can have pretty columns rendered with platform specific metadata.

We should do the same for machinedeployments.

When we support machine class, we want to revisit this.

/cc @enxebre @smarterclayton

smarterclayton:
I hate machinesets columns (which are really deployment columns) but it's fine.  I have open bugs upstream to make them not suck, but it's not your problem

enxebre:
thanks! I added a commit on your branch to make yaml lint happy

ingvagabund:
> It is inevitable that you will want printers for each provider.

Would be great to have a clean way how to do that.

> The beta is targeting AWS, so this is good for now. As we surface other providers, we may want to look at having the machine-api-operator itself install the CRDs that vary by provider so we can have pretty columns rendered with platform specific metadata.

As long as the CRDs are again managed by the machine-api-operator, it is acceptable. Though, it will cause troubles when (and if) we start supporting multiple instances of an actuator inside the same cluster (a.k.a hybrid clouds).

PR #183, opened 2019-01-22
Update links to different CRD's in README

praveenkumar:
Right now if someone wants to try out that `Dev` workflow he/she need to first go to the current links which are part of test directory but those contains the symlink to actual file which is not intuitive.


ingvagabund:
@enxebre do we need the crds under `test/integration/manifests` at all?

enxebre:
closed by https://github.com/openshift/machine-api-operator/commit/02a2e2687dbe49bbba2fada7192d4dafe23c95ee#diff-04c6e90faac2675aa89e2176d2eec7d8

PR #185, opened 2019-01-28
Include CRDs for openshift.io

enxebre:
Although we aim to consume directly from upstream repo, we are temporary mirroring upstream API under openshift.io https://github.com/openshift/cluster-api. This PR makes both available for a backward compatible migration


enxebre:

PR #188, opened 2019-01-31
Openshift machine api

enxebre:
Since we are migrating to use openshift.io for machine API. We want to do a backward compatible transition.
This runs controllers for both groups simultaneously
TODO:
- Add actuator controllers
- Update cluster-api vendored sha


enxebre:

cmd/nodelink-controller/main.go
```diff
@@ -61,15 +60,15 @@ const (
 	controllerName = "nodelink"
 
 	// machineAnnotationKey is the annotation storing a link between a node and it's machine. Should match upstream cluster-api machine controller. (node.go)
-	machineAnnotationKey = "cluster.k8s.io/machine"
+	machineAnnotationKey = "machine.k8s.io/machine"
```
bison: This seems a little weird. I feel like we should either keep this the same as upstream, or go ahead and change it to `openshift.io`.

owned-manifests/clusterapi-manager-controllers.yaml
```diff
@@ -61,6 +61,20 @@ spec:
         args:
           - --logtostderr=true
           - --v=3
+      - name: nodelink-controller-deprecated
+        image: quay.io/coreos/machine-api-operator:origin-v4.0-2019-01-31-041134
```
bison: I think it would be nice to template these in after having read them from the config like we do for the others. Even if it's temporary, it would be nice for these to come from the release image. I suppose this is not a total blocker though.

cmd/nodelink-controller/main.go
enxebre: sure, it was meant to be `machine.openshift.io`, thanks!

owned-manifests/clusterapi-manager-controllers.yaml
enxebre: this is never gonna come from the payload/release image. The ones coming from the payload are the ones built from master so adding support for the new types.
The payload CI build will replace every appearance of https://github.com/openshift/machine-api-operator/blob/master/install/image-references#L8 with the corresponding image Stream built by CI.
We could template in the configMap though but it wouldn't make any difference

frobware:
With the obvious exception of the Godep mirror location for cluster-api, LGTM.

pkg/operator/operator.go
```diff
@@ -210,10 +210,20 @@ func (optr *Operator) maoConfigFromInstallConfig() (*OperatorConfig, error) {
 		return nil, err
 	}
 
+	// TODO: Remove once we transition over machine.openshift.io group
+	var providerDreprecatedControllerImage string
+	switch provider {
+	case AWSProvider:
+		providerDreprecatedControllerImage = images.ClusterAPIControllerAWSDeprecated
```
frobware: Our first pass at landing this work continues to use the existing images, correct?

enxebre: That's correct. Once we merge this:
- k8s-e2e can go green for https://github.com/openshift/cluster-api-provider-aws/pull/147 since it needs the new nodelink controller included in this PR
e2e-aws/e2e-aws-operator can go green here https://github.com/openshift/cluster-api-provider-aws/pull/147 since the deprecated controllers are still run by this changes

Gopkg.toml
```diff
@@ -31,8 +31,8 @@ required = [
   branch = "release-1.10"
 
 [[override]]
-  name = "sigs.k8s.io/cluster-api"
-  branch = "master"
+  name = "github.com/openshift/cluster-api"
+  revision = "91fca585a85b163ddfd119fd09c128c9feadddca"
```
ingvagabund: Fine for now though let's use tags once we finish the mirroring

cmd/nodelink-controller/main.go
```diff
@@ -61,15 +61,15 @@ const (
 	controllerName = "nodelink"
 
 	// machineAnnotationKey is the annotation storing a link between a node and it's machine. Should match upstream cluster-api machine controller. (node.go)
-	machineAnnotationKey = "cluster.k8s.io/machine"
+	machineAnnotationKey = "machine.openshift.io/machine"
```
ingvagabund: Time to use https://github.com/openshift/cluster-api/blob/release-4.0/pkg/controller/node/node.go#L35 instead.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 const (
-	machineAnnotationKey = "cluster.k8s.io/machine"
+	machineAnnotationKey = "machine.openshift.io/machine"
```
ingvagabund: Time to use https://github.com/openshift/cluster-api/blob/release-4.0/pkg/controller/node/node.go#L35 instead.

enxebre: Agreed, but that belongs to a different PR. We need to go step by step. Let's switch the api group. Then do improvements

PR #191, opened 2019-02-01
add back controllers using latest provider image so using machine.openshift.io

enxebre:
- Add back controllers using latest provider image so using machine.openshift.io https://github.com/openshift/cluster-api-provider-aws/pull/147
- We have now all controllers for `cluster.k8s.io` and `machine.openshift.io`


enxebre:

ingvagabund:
From the design point of view accepted.
/approve

PR #192, opened 2019-02-04
Add OpenStack deprecated image for transitioning machine API groups

enxebre:
This adds OpenStack deprecated image for transitioning machine API groups.
Support for cluster.k8s.io will be dropped once all the providers in the installer point to the new group
e.g aws https://github.com/openshift/installer/pull/1175
For reference:
https://github.com/openshift/machine-api-operator/pull/188
https://github.com/openshift/machine-api-operator/pull/191

PR #193, opened 2019-02-04
openstack: Add the deprecated controller image

tomassedovic:
The `machine-api-operator` pod was failing on OpenStack because the
`.Controllers.ProviderDeprecated` value was set to an empty string.

This makes sure that value is set for OpenStack as well to fix the
actuator coming up.


tomassedovic:
@enxebre ah, d'oh! Yeah, I prefer #192, but I'd like to hear what @flaper87 thinks as well.

All of this deprecated stuff will be removed soon anyway, right?

enxebre:
@tomassedovic yes, support for cluster.k8s.io will be dropped asap. Once this gets https://github.com/openshift/installer/pull/1175 and your analogous PR for Openstack in the installer get merged

PR #201, opened 2019-02-06
Drop deprecated controllers, update e2e to machine.openshift.io

enxebre:
Drop deprecated controllers
Update e2e-operator test to run against machine.openshift.io
Leave cluster.k8s.io CRDs for now since they are still used by openstack


pkg/operator/operator.go
```diff
@@ -210,22 +210,10 @@ func (optr *Operator) maoConfigFromInstallConfig() (*OperatorConfig, error) {
...
-		providerDreprecatedControllerImage = images.ClusterAPIControllerOpenStackDeprecated
-	}
-
 	return &OperatorConfig{
 		optr.namespace,
 		Controllers{
 			providerControllerImage,
-			providerDreprecatedControllerImage,
 			machineAPIOperatorImage,
 			machineAPIOperatorImage,
```
frobware: How come we pass this twice?

enxebre: since both controllers rely on the same image:
```type Controllers struct
struct {
    Provider string
    NodeLink string
    MachineHealthCheck string
}```

enxebre:
e2e-aws-operator needs https://github.com/openshift/cluster-autoscaler-operator/pull/35 and https://github.com/openshift/kubernetes-autoscaler/pull/29

PR #208, opened 2019-02-13
Drop cluster.k8s crds

enxebre:
Drop cluster.k8s crds.
Leave cluster.cluster.k8s.io for now since openstack is using it


ingvagabund:
Looks good. Keeping just the cluster object in both groups.

/approve

PR #210, opened 2019-02-14
Move rbac up to CVO. Drop admin perms

enxebre:
Move rbac up to CVO. Drop admin perms


enxebre:
cc @bison

install/0000_30_machine-api-operator_08_rbac.yaml
```diff
@@ -1,13 +1,78 @@
...
+  name: cluster-api-manager-rolebinding
 roleRef:
-  kind: ClusterRole
-  name: cluster-admin
   apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-api-manager-role
+subjects:
+  - kind: ServiceAccount
+    name: default
```
bison: Can we create a dedicated service account? It's not much more trouble, and it means random pods started in the namespace don't get elevated permissions.

install/0000_30_machine-api-operator_08_rbac.yaml
```diff
@@ -1,13 +1,78 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: cluster-api-manager-role
```
bison: This is a nit, but I don't think these names should end in `-role` etc. You already know that by the type, and it looks weird in the console. I would just call this `machine-api` or something.

install/0000_30_machine-api-operator_08_rbac.yaml
```diff
@@ -1,13 +1,78 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
-apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
-  name: default-account-openshift-machine-api
-subjects:
-- kind: ServiceAccount
-  name: default
-  namespace: openshift-machine-api
+  name: cluster-api-manager-rolebinding
```
bison: Same nit on the name including the type here.

enxebre:
Before this merges, we need to make sure we add a validation for draining in this repo and we add the required permissions e.g manipulating daemonSets https://github.com/openshift/cluster-api-provider-aws/pull/158 cc @ingvagabund

enxebre:
Looking at the logs now perms list seems ok. I'll update for:
- Using specific sa
- Including draining permissions

ingvagabund:
let's wait until I introduce new test for the draining

enxebre:
@ingvagabund @bison @frobware let's get this in. I'll add a specific SA in a follow up

frobware:
The draining test seems orthogonal to the purpose of the PR (given the description). Can we move the test to a separate PR?

enxebre:
@frobware The perms has been made granular including the new draining featured added in aws actuator. This needs to be merged in one go, otherwise there is room for breaking draining. That's why they have a different commit

frobware:
> @frobware The perms has been made granular including the new draining featured added in aws actuator. This needs to be merged in one go, otherwise there is room for breaking draining. That's why they have a different commit

OK, thanks for the explanation.

ingvagabund:
/approve

Agree with Andrew that the node draining test should go into separate PR and merged first before this PR is. Though, given the current state of the CI I agree with Alberto it's better to keep it as it is.

test/e2e/operator_expectations.go
```diff
@@ -608,3 +613,481 @@ func (tc *testConfig) ExpectAutoscalerScalesOut() error {
...
+			glog.Errorf("error querying api for eventList object: %v, retrying...", err)
+			return false, nil
+		}
+
+		glog.Infof("Fetching delete machine and node drained events")
+		var nodeDrainedEvent *corev1.Event
+		var machineDeletedEvent *corev1.Event
+		for _, eventItem := range eventList.Items {
+			if eventItem.Reason == "Deleted" && eventItem.Message == fmt.Sprintf("Node %q drained", nodeName) {
+				nodeDrainedEvent = &eventItem
```
frobware: We're taking the address of the local loop variable here - is this intentional?

ingvagabund: It is not, it should be changed actually. Though, `ExpectNodeToBeDrainedBeforeDeletingMachine` was superseded by `ExpectNodeToBeDrainedBeforeMachineIsDeleted` so it can be deleted.

frobware:
> Agree with Andrew that the node draining test should go into separate PR and merged first before this PR is. Though, given the current state of the CI I agree with Alberto it's better to keep it as it is.

I think the key point is if you ever had to revert this PR would you also want to revert the test. If the answer is "no" then they should be separate commits.

test/e2e/operator_expectations.go
enxebre: Thanks! Seems ok since this is not storing appended values or similar
This is just bringing @ingvagabund draining tests from aws. To validate the new rbac does not break current coverage

Right after this we can merge https://github.com/openshift/machine-api-operator/pull/219
and https://github.com/openshift/cluster-api-actuator-pkg/pull/28
And start rewriting the e2e suite there addressing this specific concern and others

test/e2e/operator_expectations.go
```diff
@@ -608,3 +613,481 @@ func (tc *testConfig) ExpectAutoscalerScalesOut() error {
...
+	glog.Info("Get nodeList")
+	if err := wait.PollImmediate(1*time.Second, waitShort, func() (bool, error) {
+		nodeList := corev1.NodeList{}
+		if err := tc.client.List(context.TODO(), &listOptions, &nodeList); err != nil {
+			glog.Errorf("error querying api for nodeList object: %v, retrying...", err)
+			return false, nil
+		}
+		for _, nodeItem := range nodeList.Items {
+			if nodeItem.Name == nodeName {
+				node = &nodeItem
```
frobware: This takes the address of a local variable which seems Ok-ish here given we `break`. But it may be clearer and more future proof to use `&nodeList.Items[i]`.

frobware:
As long as we follow up and address the issues RSN...

/lgtm

PR #211, opened 2019-02-15
Adding new reviewers beekhof and cynepco3hahue for health check controller

enxebre:
@beekhof @cynepco3hahue on CNV team are using the machine API Operator domain for building solutions for machine health checking on bare metal.


enxebre:
since we bumped minikube on CI integratin will fail until https://github.com/openshift/machine-api-operator/pull/209 gets merged

PR #217, opened 2019-02-20
MAO: Stop setting statusProgressing=true when resyincing same version

michaelgugino:
Fixes: https://jira.coreos.com/browse/CLOUD-391


pkg/operator/status.go
```diff
@@ -38,18 +38,20 @@ func (optr *Operator) statusProgressing() error {
...
 	} else {
 		message = fmt.Sprintf("Running resync for %s", optr.printOperandVersions())
+		isProgressing = osconfigv1.ConditionFalse
 	}
 
 	conds := []osconfigv1.ClusterOperatorStatusCondition{
 		{
 			Type:    osconfigv1.OperatorProgressing,
-			Status:  osconfigv1.ConditionTrue,
+			Status:  isProgressing,
```
ingvagabund: Looks good :) can you also add a test here [1] to verify the condition status is true when `desiredVersions` != `currentVersions` and false when both are equal?

[1] https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/status_test.go

enxebre: +1, also including `failing=false` and `progressing=false` here would be nice https://github.com/openshift/machine-api-operator/blob/master/test/e2e/operator_expectations.go#L68

michaelgugino: @ingvagabund done

pkg/operator/status_test.go
```diff
@@ -25,3 +27,55 @@ func TestPrintOperandVersions(t *testing.T) {
...
+		optr.statusProgressing()
+		o, _ := optr.osClient.ConfigV1().ClusterOperators().Get(clusterOperatorName, metav1.GetOptions{})
+		var condition osconfigv1.ClusterOperatorStatusCondition
+		for _, coCondition := range o.Status.Conditions {
+			if coCondition.Type == osconfigv1.OperatorProgressing {
+				condition = coCondition
+				break
+			}
+		}
+		assert.Equal(t, tc.expected, condition.Status, "test-case %v expected condition %v to be %v, but got %v", i, condition.Type, tc.expected, condition.Status)
```
michaelgugino: Copied this from machine-config-operator.  It prints really nice output in case of test failure.  We can remove this if we don't want to add in this dependency.

test/e2e/operator_expectations.go
```diff
@@ -64,12 +64,25 @@ func (tc *testConfig) ExpectClusterOperatorStatusAvailable() error {
 			glog.Errorf("error querying api for OperatorStatus object: %v, retrying...", err)
 			return false, nil
 		}
+		ready := false
 		if available := cvoresourcemerge.FindOperatorStatusCondition(clusterOperator.Status.Conditions, osconfigv1.OperatorAvailable); available != nil {
```
ingvagabund: @enxebre do you want to block this PR until https://github.com/openshift/machine-api-operator/pull/219 is merged?

pkg/operator/status_test.go
ingvagabund: I don't see any machine-config-operator dep :) Did you update the PR meantime?

michaelgugino: @ingvagabund no, I mean the 'assert' library.  They are using this library, it seems nice.  But I can just do a normal test here if we don't want to pull this in.

michaelgugino: (This is what machine-config-operator is using)

ingvagabund: I see :) I am completely fine with the new dependency.

pkg/operator/status_test.go
```diff
@@ -25,3 +27,55 @@ func TestPrintOperandVersions(t *testing.T) {
 		t.Errorf("Expected: %s, got: %s", expectedOutput, got)
 	}
 }
+
+func TestOperatorStatusProgressing(t *testing.T) {
+	type tCase struct {
```
frobware: nit: turn this into an anonymous struct.

PR #219, opened 2019-02-21
E2e cluster api actuator pkg

enxebre:
This is part of an effort to abstract away the e2e tests duplicated across repos which validate holistic expected behaviour of the system which is provided by the interaction of one or more components of the machine API
https://github.com/openshift/cluster-api-actuator-pkg/pull/28


enxebre:
cc @ingvagabund @bison

PR #228, opened 2019-02-26
machine-api-operator should not set limits

derekwaynecarr:
do not set limits for the operator per no component that runs on masters should use limits.
update cpu requests to 10m align with other operators.


enxebre:
@bison integration needs https://github.com/openshift/machine-api-operator/pull/233

PR #229, opened 2019-02-27
Add system-critical priority class to controller

ravisantoshgudimetla:
Add system-cluster-critical priority class to controller,  without this there is a chance that scheduler evicts these pods when there is a resource crunch.

### DIFF
--- a/owned-manifests/clusterapi-manager-controllers.yaml
+++ b/owned-manifests/clusterapi-manager-controllers.yaml
@@ -21,6 +21,7 @@ spec:
         api: clusterapi
         k8s-app: controller
     spec:
+      priorityClassName: system-node-critical
       nodeSelector:
         node-role.kubernetes.io/master: ""
       tolerations:


enxebre:
/approve
needs https://github.com/openshift/machine-api-operator/pull/233

ravisantoshgudimetla:
@ingvagabund - Jan, I see that this PR failing tests continuously with error:

`time="2019-02-28T09:40:27Z" level=info msg="Waiting for all clusterapi-manager-controllers deployment pods to be ready, have 0, expecting 1"`

Any other changes that could have caused this issue?

ingvagabund:
@ravisantoshgudimetla lemme take a look

ingvagabund:
2 facts:
- `ci/jenkins/integration` is over over pure k8s
- controllers are deployed into `openshift-machine-api` namespace

At this point only `kube-system` can run critical pods. I will need to update the job to deploy into `kube-system` instead of `openshift-machine-api`.

ravisantoshgudimetla:
Thanks @ingvagabund. 

> At this point only kube-system can run critical pods.

That is right. Shouldn't we run CI over OpenShift 4.0 clusters?

ingvagabund:
> That is right. Shouldn't we run CI over OpenShift 4.0 clusters?

we do both, OpenShift 4.0 and Kubernetes, I will fix it ASAP.

ingvagabund:
@ravisantoshgudimetla we are about to remove `ci/jenkins/integration`

ingvagabund:
you don't need to do anything here, we will just remove the job and switch to a new one that deploys all controllers into `kube-system`. Just waiting until https://github.com/openshift/machine-api-operator/pull/220 lands. Then I can switch. Should not take more than 2hrs.

ravisantoshgudimetla:
@ingvagabund - Thank you :)

ravisantoshgudimetla:
/hold

Till #220 lands. I will remove the hold once we are good to go.

PR #231, opened 2019-02-27
Logging and events

enxebre:
Improve logging and add events for the operator cluster status
Conventions:
- https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md
- if an error message is just printed, use uppercase
- if an error message is return by a function/method and passed on, use lowercase
- NOUN VERB STATE format when possible

### DIFF
--- a/cmd/machine-api-operator/helpers.go
+++ b/cmd/machine-api-operator/helpers.go
@@ -7,7 +7,7 @@ import (
 
 	"github.com/golang/glog"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -41,7 +41,7 @@ func CreateResourceLock(cb *ClientBuilder, componentNamespace, componentName str
 
 	id, err := os.Hostname()
 	if err != nil {
-		glog.Fatalf("error creating lock: %v", err)
+		glog.Fatalf("Error creating lock: %v", err)
 	}
 
 	// add a uniquifier so that two processes on the same host don't accidentally both become active
--- a/cmd/machine-api-operator/start.go
+++ b/cmd/machine-api-operator/start.go
@@ -67,7 +67,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: run,
 			OnStoppedLeading: func() {
-				glog.Fatalf("leaderelection lost")
+				glog.Fatalf("Leader election lost")
 			},
 		},
 	})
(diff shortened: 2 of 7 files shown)


cmd/machine-api-operator/helpers.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	"github.com/golang/glog"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
ingvagabund: why?

cmd/machine-api-operator/start.go
```diff
@@ -67,7 +67,7 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: run,
 			OnStoppedLeading: func() {
-				glog.Fatalf("leaderelection lost")
+				glog.Fatalf("leader election lost")
```
ingvagabund: uppercase `Leader`

cmd/machine-api-operator/helpers.go
enxebre: goimports generates it

cmd/machine-api-operator/start.go
frobware: Awkward one, but I tend to go with: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

frobware:
I think the "rules" would be simpler if it was lower _always.. With that simple rule there's no ambiguity. Code reviews go faster because there's only 1 rule. Error message are consistently formatted. It is also idiomatic Go [1].

[1] - https://github.com/golang/go/wiki/CodeReviewComments#error-strings

enxebre:
/test integration
for debugging

enxebre:
fixes integration https://github.com/openshift/machine-api-operator/pull/233

pkg/operator/status.go
```diff
@@ -39,10 +41,18 @@ func (optr *Operator) statusProgressing() error {
...
+	if err != nil {
+		glog.Errorf("Failed to get or create Cluster Operator: %v", err)
+		return err
+	}
+
 	var message string
 	if !reflect.DeepEqual(desiredVersions, currentVersions) {
 		glog.V(2).Info("Syncing status: progressing")
 		message = fmt.Sprintf("Progressing towards %s", optr.printOperandVersions())
+		optr.eventRecorder.Eventf(co, v1.EventTypeNormal, "Status upgrade", message)
```
ingvagabund: `Status upgraded` or `Status upgraging` maybe?

ingvagabund:
Does it make sense to report `Status successfully updated` as well?

cmd/nodelink-controller/main.go
```diff
@@ -388,17 +388,17 @@ func (c *Controller) processNode(node *corev1.Node) error {
...
-		glog.V(3).Infof("copying label %s = %s", k, v)
+		glog.V(3).Infof("Copying label %s = %s", k, v)
 		modNode.Labels[k] = v
 	}
 
 	addTaintsToNode(modNode, matchingMachine)
 
 	if !reflect.DeepEqual(node, modNode) {
-		glog.V(3).Infof("node has changed, updating")
+		glog.V(3).Infof("Node has changed, updating")
```
frobware: If we're going to log we might as well indicate _which_ node.

PR #232, opened 2019-02-27
Remove resource limits in operator deployment

bison:
This was reported in Bugzilla. Second-level-operators are expected to set resource requests, but explicitly should _not_ set limits.


bison:
Closing in favor of https://github.com/openshift/machine-api-operator/pull/228.

PR #236, opened 2019-02-28
status: report cluster operator version from payload

derekwaynecarr:
Every operator shall report a version with name=operator whose value is equal to the RELEASE_VERSION provided by the cluster version operator.

will be enforced via [openshift/origin#22156](https://github.com/openshift/origin/pull/22156)


enxebre:
/approve
@derekwaynecarr thanks! is there any doc we can refer pointing this out?

PR #237, opened 2019-02-28
nodelink-controller: ensure node annotations is non nil before adding values

frobware:
I experienced two panics when testing/debugging the nodelink-controller using kubemark. In both cases the panic was because `node.Annotations` was nil.

Additional drive-by fixes to:
- make a logging statement more useful
- to remove a logging statement that was useless


frobware:
My testing was related to: https://jira.coreos.com/browse/CLOUD-214

ingvagabund:
This makes sense.

/approve

enxebre:
interesting, nice catch

PR #239, opened 2019-03-01
Rename /install to /manifests

ingvagabund:
Imagebuilder does not implemented COPY the same way as docker does.
So running `COPY --from=builder /go/src/github.com/openshift/machine-api-operator/install manifests`
will actually create `manifests/install` directory instead of copying content of
install under manifests.


ingvagabund:
Hmm, `COPY` command seems to be fixed with the latest `imagebuilder` binary.

PR #240, opened 2019-03-01
Remove deprecated test/integration in favor of ginkgo based conformance tests

ingvagabund:
New JJ: https://github.com/openshift/aos-cd-jobs/pull/1719

PR #251, opened 2019-03-20
Fix links to operator status and install config

ingvagabund:
`install-config.yaml` and `config/status-crd.yaml` were removed by https://github.com/openshift/machine-api-operator/pull/240. Adding the files back and updating links to them.

### DIFF
--- a/config/machineset.yaml
+++ b/config/machineset.yaml
@@ -1,0 +1,53 @@
+apiVersion: machine.openshift.io/v1beta1
+kind: MachineSet
+metadata:
+  labels:
+    machine.openshift.io/cluster-api-cluster: {{ .ClusterID }}
+    machine.openshift.io/cluster-api-machine-role: worker
+    machine.openshift.io/cluster-api-machine-type: worker
+  name: worker
+  namespace: openshift-machine-api
+spec:
+  replicas: 2
+  selector:
+    matchLabels:
+      machine.openshift.io/cluster-api-cluster: {{ .ClusterID }}
+      machine.openshift.io/cluster-api-machineset: worker
+  template:
+    metadata:
+      labels:
+        machine.openshift.io/cluster-api-cluster: {{ .ClusterID }}
+        machine.openshift.io/cluster-api-machine-role: worker
+        machine.openshift.io/cluster-api-machine-type: worker
+        machine.openshift.io/cluster-api-machineset: worker
+    spec:
+      providerSpec:
+        value:
+          credentialsSecret:
+            name: aws-credentials-secret
+          ami:
+            id: ami-05c79276d240b93bc
+          apiVersion: awsproviderconfig.openshift.io/v1beta1
+          iamInstanceProfile:
+            id: {{ .ClusterID }}-worker-profile
+          instanceType: t2.medium
+          kind: AWSMachineProviderConfig
+          placement:
+            region: us-east-1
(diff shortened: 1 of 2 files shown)


config/install-config.yaml
```diff
@@ -0,0 +1,32 @@
+admin:
+  email: test
+  password: test
+  sshKey: |
+    ssh-rsa
+baseDomain: test
+clusterID: a7265676-7dc3-4ff3-8759-f2d6e3934e76
```
frobware: This seems very specific; is it in fact needed?

config/machineset.yaml
```diff
@@ -0,0 +1,53 @@
...
+        machine.openshift.io/cluster-api-machine-role: worker
+        machine.openshift.io/cluster-api-machine-type: worker
+        machine.openshift.io/cluster-api-machineset: worker
+    spec:
+      providerSpec:
+        value:
+          credentialsSecret:
+            name: aws-credentials-secret
+          ami:
+            id: ami-05c79276d240b93bc
```
frobware: What is this? Will it always be around?

config/status-crd.yaml
```diff
@@ -0,0 +1,34 @@
...
+  names:
+    kind: ClusterOperator
+    listKind: ClusterOperatorList
+    plural: clusteroperators
+    singular: clusteroperator
+  scope: Cluster
+  subresources:
+    status: {}
+  version: v1
+
```
frobware: nit: drop blank/empty lines at end of file.

ingvagabund:
@frobware I only copy pasted the files, did not intend the modify their content. If the content needs to be changed, we might put it into a separate PR.

config/machineset.yaml
ingvagabund: most likely not

config/install-config.yaml
ingvagabund: Something has to be set it. In this context it's basically a random hash

config/machineset.yaml
ingvagabund: a user will have to substitute those values

config/status-crd.yaml
ingvagabund: copy-paste heresy. harmless

config/status-crd.yaml
```diff
@@ -0,0 +1,34 @@
+kind: CustomResourceDefinition
```
enxebre: this can point to https://github.com/openshift/machine-api-operator/blob/master/config/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml intead

config/install-config.yaml
enxebre: this can just point to https://github.com/openshift/machine-api-operator/blob/master/config/kubemark-install-config.yaml instead

ingvagabund: Would not be better to point to some more real install config? I still prefer the aws one since it's way closer to the production deployment.

enxebre: this for the case where your are deploying on your own env in a non OpenShift context, the only assumption made by the operator is https://github.com/openshift/machine-api-operator/blob/master/config/kubemark-install-config.yaml#L8 anything else is has no purpose in this context

ingvagabund:
@enxebre comments addressed, PTAL

PR #252, opened 2019-03-20
mao: add support for 'None' provider type

michaelgugino:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1691120

### DIFF
--- a/pkg/operator/config.go
+++ b/pkg/operator/config.go
@@ -33,6 +33,7 @@ const (
 	KubemarkProvider = Provider("kubemark")
 	// BareMetalPlatformType is used for install using managed Bare Metal
 	BareMetalProvider = Provider("baremetal")
+	NoneProvider      = Provider("none")
 )
 
 type Provider string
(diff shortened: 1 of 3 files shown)


pkg/operator/config.go
```diff
@@ -81,6 +82,9 @@ type InstallPlatform struct {
 
 	// BareMetal is the configuration used when running on managed Bare Metal
 	BareMetal interface{} `json:"baremetal,omitempty"`
+
+	// None is the configuration used when running on unmanaged UPI
+	None interface{} `json:"baremetal,omitempty"`
```
abhinavdahiya: `json:"none,omitempty"`

pkg/operator/operator.go
```diff
@@ -176,7 +176,14 @@ func (optr *Operator) sync(key string) error {
 		glog.Errorf("Failed getting operator config: %v", err)
 		return err
 	}
-
+	if operatorConfig.Controllers.Provider == "None" {
+		glog.V(1).Info("`None` provider specified, reporting available")
+		if err := optr.statusAvailable(); err != nil {
+			glog.Errorf("Error syncing ClusterOperatorStatus: %v", err)
+			return fmt.Errorf("error syncing ClusterOperatorStatus: %v", err)
```
enxebre: @michaelgugino should we `optr.statusFailing` here?

michaelgugino: I don't think so.  If we fail to set statusAvailable it's most likely due to temporary network outage / api unavailable, so we want to retry I think.

Generally, if provider is None type, I think we want to do a no-op and report available.

PR #254, opened 2019-03-25
Set operator status failing if sync operator fails

vikaschoudhary16:


pkg/operator/operator.go
```diff
@@ -174,7 +176,7 @@ func (optr *Operator) sync(key string) error {
 	operatorConfig, err := optr.maoConfigFromInstallConfig()
 	if err != nil {
 		glog.Errorf("Failed getting operator config: %v", err)
-		return err
+		return fmt.Errorf("Failed getting operator config: %v", err)
```
enxebre: passed errors should use lower case

pkg/operator/operator.go
```diff
@@ -158,7 +158,9 @@ func (optr *Operator) handleErr(err error, key interface{}) {
 		optr.queue.AddRateLimited(key)
 		return
 	}
-
+	if err := optr.statusFailing(err.Error()); err != nil {
```
frobware: I don't grok this: we call `statusFailing()` and log an error if this fails. And then, given the scoping rules, we continue to use/consume `err` as passed in. Where do we "set operator status failing if sync operator fails"?

frobware: Please can we add a unit test that asserts this behaviour.

ingvagabund: Are there any cases where `err != nil` and it's still fine not to set the status failing?

enxebre:
@vikaschoudhary16 can we please add a section on the README elaborating the specific cases when we report failing inline with https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md?

pkg/operator/operator.go
vikaschoudhary16: There could be any transient errors like because of network connectivity in such cases should be retried before setting operator status.

vikaschoudhary16:
Now we are already setting `StatusDegraded`, which superceded `StatusFailing`

PR #255, opened 2019-03-25
Drop unused labels

vikaschoudhary16:
Depends on: https://github.com/openshift/cluster-api-provider-aws/pull/182

### DIFF
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -331,21 +331,8 @@ func hasMatchingLabels(machineHealthCheck *healthcheckingv1alpha1.MachineHealthC
 }
 
 func isMaster(machine mapiv1.Machine, client client.Client) bool {
-	machineMasterLabels := []string{
-		"machine.openshift.io/cluster-api-machine-role",
-		"machine.openshift.io/cluster-api-machine-type",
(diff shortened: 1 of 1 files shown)


pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -331,19 +331,13 @@ func hasMatchingLabels(machineHealthCheck *healthcheckingv1alpha1.MachineHealthC
...
 
 func isMaster(machine mapiv1.Machine, client client.Client) bool {
-	machineMasterLabels := []string{
-		"machine.openshift.io/cluster-api-machine-role",
-		"machine.openshift.io/cluster-api-machine-type",
-		"sigs.k8s.io/cluster-api-machine-role",
-		"sigs.k8s.io/cluster-api-machine-type",
-	}
-	nodeMasterLabels := []string{
+	masterLabels := []string{
```
cynepco3hahue: Does it make sense to check machine labels at all?
IMHO the only proper way is to check node labels that machine reference to.

enxebre: I agree. @vikaschoudhary16 as part of this effort can we drop labels here https://github.com/openshift/installer/tree/master/pkg/asset/machines/aws
And update https://github.com/openshift/origin/pull/22090/files#r267287918 to fetch from node accordingly cc @ingvagabund @derekwaynecarr

PR #257, opened 2019-03-27
Generate crds

ingvagabund:
The process is still partially manual. I.e. after the CRDs are updated, one still needs to check if all the necessary fields are present. E.g. add back `additionalPrinterColumns` field.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,10 @@ nodelink-controller:
 machine-healthcheck:
 	$(DOCKER_CMD) ./hack/go-build.sh machine-healthcheck
 
+.PHONY: gen-crd
+gen-crd:
+	./hack/gen-crd.sh
+
 .PHONY: update-codegen
 update-codegen:
 	$(DOCKER_CMD) ./hack/update-codegen.sh
--- a/Makefile
+++ b/Makefile
@@ -42,9 +42,12 @@ nodelink-controller:
 machine-healthcheck:
 	$(DOCKER_CMD) ./hack/go-build.sh machine-healthcheck
 
+.PHONY: generate
+generate: gen-crd update-codegen
+
 .PHONY: gen-crd
 gen-crd:
-	./hack/gen-crd.sh
+	$(DOCKER_CMD) ./hack/gen-crd.sh
 
 .PHONY: update-codegen
 update-codegen:
--- a/hack/gen-crd.sh
+++ b/hack/gen-crd.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+set -eu
+
+echo "Building controller-gen tool..."
+go build -o bin/controller-gen github.com/openshift/machine-api-operator/vendor/sigs.k8s.io/controller-tools/cmd/controller-gen
+
+dir=$(mktemp -d -t XXXXXXXX)
+echo $dir
+mkdir -p $dir/src/github.com/openshift/machine-api-operator/pkg/apis
+mkdir -p $dir/src/github.com/openshift/machine-api-operator/vendor
+
+cp -r pkg/apis/healthchecking $dir/src/github.com/openshift/machine-api-operator/pkg/apis/.
+cp -r vendor/github.com/openshift/cluster-api/pkg/apis/machine $dir/src/github.com/openshift/machine-api-operator/pkg/apis
+# Some dependencies need to be coppied as well. Othwerwise, controller-gen will complain about non-existing kind Unsupported
+cp -r vendor/k8s.io $dir/src/github.com/openshift/machine-api-operator/vendor/.
+cp -r vendor/github.com $dir/src/github.com/openshift/machine-api-operator/vendor/.
+
+cwd=$(pwd)
+pushd $dir/src/github.com/openshift/machine-api-operator
+GOPATH=$dir ${cwd}/bin/controller-gen crd --domain openshift.io
+popd
+
+echo "Coping generated CRDs"
+cp $dir/src/github.com/openshift/machine-api-operator/config/crds/healthchecking_v1alpha1_machinehealthcheck.yaml install/0000_30_machine-api-operator_07_machinehealthcheck.crd.yaml
+cp $dir/src/github.com/openshift/machine-api-operator/config/crds/machine_v1beta1_machinedeployment.yaml install/0000_30_machine-api-operator_04_machinedeployment.crd.yaml
+cp $dir/src/github.com/openshift/machine-api-operator/config/crds/machine_v1beta1_machineset.yaml install/0000_30_machine-api-operator_03_machineset.crd.yaml
+cp $dir/src/github.com/openshift/machine-api-operator/config/crds/machine_v1beta1_machine.yaml install/0000_30_machine-api-operator_02_machine.crd.yaml
+
+rm -rf $dir
--- a/install/0000_30_machine-api-operator_07_machinehealthcheck.crd.yaml
+++ b/install/0000_30_machine-api-operator_07_machinehealthcheck.crd.yaml
@@ -1,22 +1,44 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
+  creationTimestamp: null
+  labels:
+    controller-tools.k8s.io: "1.0"
   name: machinehealthchecks.healthchecking.openshift.io
 spec:
   group: healthchecking.openshift.io
   names:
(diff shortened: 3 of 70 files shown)


install/0000_30_machine-api-operator_07_machinehealthcheck.crd.yaml
```diff
@@ -1,22 +1,44 @@
...
 metadata:
+  creationTimestamp: null
+  labels:
+    controller-tools.k8s.io: "1.0"
   name: machinehealthchecks.healthchecking.openshift.io
 spec:
   group: healthchecking.openshift.io
   names:
     kind: MachineHealthCheck
-    listKind: MachineHealthCheckList
```
ingvagabund: @enxebre not sure about this if it's really needed.

Makefile
```diff
@@ -42,6 +42,18 @@ nodelink-controller:
 machine-healthcheck:
 	$(DOCKER_CMD) ./hack/go-build.sh machine-healthcheck
 
+.PHONY: gen-crd
+gen-crd:
+	cp -r vendor/github.com/openshift/cluster-api/pkg/apis/machine pkg/apis/.
```
enxebre: could we may be run the cp, run and rm steps against a tmp folder that we add to our .gitignore list

michaelgugino: Maybe we should add these commands to a script in hack/ so we don't clutter up the make file.

ingvagabund: We can't unless we recreate separate GOPATH/src/... workspace. Lemme try it.

ingvagabund: Done. Copying files under temp dir and generating the CRDs there.

Makefile
```diff
@@ -42,6 +42,10 @@ nodelink-controller:
 machine-healthcheck:
 	$(DOCKER_CMD) ./hack/go-build.sh machine-healthcheck
 
+.PHONY: gen-crd
+gen-crd:
```
michaelgugino: I like it.  Perhaps we can add this script before as the first step in our make fmt, as that will detect any differences in git diff to ensure we don't have drift;  Anything that would change a CRD should generate before push.

ingvagabund: We can't do it right know since the generated CRDS for machine and machineset removes `additionalPrinterColumns`. And those are aws and openshift specific.

michaelgugino: @ingvagabund are you saying, after generating CRDs, someone needs to modify them?

ingvagabund: correct

hack/gen-crd.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
```
enxebre: can't we just modify the `pkg/apis` layout here to satisfy the crd command then just:
```
go run controller-gen/main.go crd --domain openshift.io --output-dir=~/my-new-crds/
pushd cluster-api
go run controller-gen/main.go crd --domain openshift.io --output-dir=~/my-new-crds/
```

ingvagabund:
```
$ pwd
/home/jchaloup/Projects/src/github.com/openshift/machine-api-operator/vendor/github.com/openshift/cluster-api
$ go run ../../../sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd --domain openshift.io
W0403 10:13:42.055714   16847 parse.go:224] Ignoring directory ./pkg/apis: No files for pkg "github.com/openshift/cluster-api/pkg/apis"
CRD files generated, files can be found under path /home/jchaloup/Projects/src/github.com/openshift/machine-api-operator/vendor/github.com/openshift/cluster-api/config/crds.
$ ls /home/jchaloup/Projects/src/github.com/openshift/machine-api-operator/vendor/github.com/openshift/cluster-api/config/crds
$ 
```

No CRDs got generated.

enxebre: :/ thanks for trying it! I still don't wanna give up on finding a nicer way, let us investigate a bit more before merging

Makefile
frobware: > correct

Oh - I would say this is bad. Somebody at some point will not realise and they things will go amiss.

enxebre: what about we drop the step for copying the files and only automate the generation

hack/gen-crd.sh
```diff
@@ -0,0 +1,31 @@
...
+# Some dependencies need to be coppied as well. Othwerwise, controller-gen will complain about non-existing kind Unsupported
+cp -r vendor/k8s.io $dir/src/github.com/openshift/machine-api-operator/vendor/.
+cp -r vendor/github.com $dir/src/github.com/openshift/machine-api-operator/vendor/.
+
+cwd=$(pwd)
+pushd $dir/src/github.com/openshift/machine-api-operator
+GOPATH=$dir ${cwd}/bin/controller-gen crd --domain openshift.io
+popd
+
+echo "Coping generated CRDs"
```
enxebre: given
> The process is still partially manual. I.e. after the CRDs are updated, one still needs to check if all the necessary fields are present. E.g. add back additionalPrinterColumns field.
I think we should leave the copying step manual. Also we woudn't want to copy machineclass

ingvagabund:
Generated CRDs are now moved under `install.generated` directory. It's not git ignored so a user can notice the new dir and copy relevant CRDs from it under `install` directory.

frobware:
> Generated CRDs are now moved under `install.generated` directory. It's not git ignored so a user can notice the new dir and copy relevant CRDs from it under `install` directory.

Is there a README (or other) explaining the expected steps?

frobware:
> Generated CRDs are now moved under `install.generated` directory. It's not git ignored so a user can notice the new dir and copy relevant CRDs from it under `install` directory.

Can we not add a Makefile target that does this?

ingvagabund:
@frobware Update README with generic `you might need to run make generate`. The generated CRDs no longer has to be manually updated.

Makefile
bison: Can't we just add the `kubebuilder:printcolumn` tags in the source so the `additionalPrinterColumns` is generated as well?

ingvagabund: Any example how to do it?

ingvagabund: I am pretty sure you mean https://github.com/openshift/machine-api-operator/pull/257/commits/0da2ff4ea36c70f002810b3d77f6a656595f1329

frobware:
I have no idea how to review 12K changes, so, intrinsically:

/lgtm

vendor/github.com/openshift/cluster-api/pkg/apis/machine/v1beta1/machine_types.go
```diff
@@ -39,6 +39,12 @@ const (
 // Machine is the Schema for the machines API
 // +k8s:openapi-gen=true
```
enxebre: This will be overwritten as a soon as we revendor and we don't want to put it in the source repo as it'll become obsolete as soon as we add a new provider

ingvagabund: We are not solving the right problem. It's wrong to put anything from `spec.providerSpec`/`.status.providerStatus` under `additionalPrinterColumns.[*].JSONPath` at the first place. Plus, the CRD specification has no conditional `if .field == <value> then print this` statement.

I am against dynamic reconciling of machine CRDs based on a config.

Instead, we SHOULD create a new field in the machine type specification that will get populated by a machine controller in a generic way and accessed by `additionalPrinterColumns.[*].JSONPath`. E.g.

```
newfieldname
aws:instanceID=<instanceID>:instanceState=<instanceState>:instanceType=<instanceType>:region=<region>:zone=<zone>
```

The column width can just get pretty long though.

bison: I agree with @ingvagabund here. Things like printer columns should not be reaching into raw-extension fields. We should define generic status fields on the CRD that providers can optionally set and use those.

I also really don't think we should go down the route of modifying our CRD definitions based on provider. Even just to add things like printer columns. I think that is a recipe for disaster when it comes to migrations and fixing issues with providers.

enxebre: yeh so I agree on https://github.com/openshift/machine-api-operator/pull/257#discussion_r273399114 I just don't see how is that related to merge that commit into the vendored folder.
We'll be introducing a new cloud provider soon. I think is more important solve the underlying problem you are stating than having the full automated command

PR #261, opened 2019-03-29
mao: add support for the vsphere platform

staebler:
The vsphere platform is a no-op for mao, like the none platform.

Adds a unit test to ensure that the mao reports available for no-op platforms.

### DIFF
--- a/pkg/operator/operator_test.go
+++ b/pkg/operator/operator_test.go
@@ -1,0 +1,127 @@
+package operator
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	configv1 "github.com/openshift/api/config/v1"
+	fakeconfigclientset "github.com/openshift/client-go/config/clientset/versioned/fake"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/record"
+)
+
+// TestOperatorSync_NoOp tests syncing to ensure that the mao reports available
+// for platforms that are no-ops.
+func TestOperatorSync_NoOp(t *testing.T) {
+	cases := []struct {
+		platform     configv1.PlatformType
+		expectedNoop bool
+	}{
+		{
+			platform:     configv1.AWSPlatformType,
+			expectedNoop: false,
+		},
+		{
+			platform:     configv1.LibvirtPlatformType,
+			expectedNoop: false,
+		},
+		{
+			platform:     configv1.OpenStackPlatformType,
+			expectedNoop: false,
+		},
+		{
+			platform:     configv1.AzurePlatformType,
+			expectedNoop: false,
+		},
+		{
+			platform:     bareMetalPlatform,
+			expectedNoop: false,
+		},
+		{
+			platform:     kubemarkPlatform,
+			expectedNoop: false,
+		},
+		{
+			platform:     configv1.VSpherePlatformType,
+			expectedNoop: true,
+		},
+		{
+			platform:     configv1.NonePlatformType,
+			expectedNoop: true,
+		},
+		{
+			platform:     "bad-platform",
+			expectedNoop: true,
+		},
+	}
+
+	tempDir, err := ioutil.TempDir("", "TestOperatorSync")
+	if err != nil {
+		t.Fatalf("could not create the temp dir: %v", err)
+	}
+	defer os.RemoveAll(tempDir)
+
+	images := Images{
+		MachineAPIOperator: "test-mao",
+	}
+	imagesAsJSON, err := json.Marshal(images)
+	if err != nil {
+		t.Fatalf("failed to marshal images: %v", err)
+	}
+	imagesFilePath := filepath.Join(tempDir, "test-images.json")
+	if err := ioutil.WriteFile(imagesFilePath, imagesAsJSON, 0666); err != nil {
+		t.Fatalf("could not write the images file: %v", err)
+	}
+
+	for _, tc := range cases {
+		t.Run(string(tc.platform), func(t *testing.T) {
+			infra := &configv1.Infrastructure{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "cluster",
+				},
+				Status: configv1.InfrastructureStatus{
+					Platform: tc.platform,
+				},
+			}
+
+			optr := Operator{
+				eventRecorder: record.NewFakeRecorder(5),
+				osClient:      fakeconfigclientset.NewSimpleClientset(infra),
+				imagesFile:    imagesFilePath,
+			}
+
+			err = optr.sync("test-key")
+			if !tc.expectedNoop {
+				if !assert.Error(t, err, "unexpected sync success") {
+					t.Fatal()
+				}
+			} else {
+				if !assert.NoError(t, err, "unexpected sync failure") {
+					t.Fatal()
+				}
+			}
+
+			o, err := optr.osClient.ConfigV1().ClusterOperators().Get(clusterOperatorName, metav1.GetOptions{})
+			if !assert.NoError(t, err, "failed to get clusteroperator") {
+				t.Fatal()
+			}
+			expectedConditions := map[configv1.ClusterStatusConditionType]configv1.ConditionStatus{
+				configv1.OperatorAvailable:   configv1.ConditionTrue,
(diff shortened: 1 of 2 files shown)


pkg/operator/config.go
```diff
@@ -166,6 +174,8 @@ func getProviderControllerFromImages(provider Provider, images Images) (string,
 		return images.ClusterAPIControllerOpenStack, nil
 	case KubemarkProvider:
 		return images.ClusterAPIControllerKubemark, nil
+	case VSphereProvider:
+		return "vSphere", nil
```
frobware: Any reason not to follow the existing convention of `images.ClusterAPIController<Type>`?

pkg/operator/operator.go
```diff
@@ -176,8 +176,9 @@ func (optr *Operator) sync(key string) error {
 		glog.Errorf("Failed getting operator config: %v", err)
 		return err
 	}
-	if operatorConfig.Controllers.Provider == "None" {
-		glog.V(1).Info("`None` provider specified, reporting available")
+	switch p := operatorConfig.Controllers.Provider; p {
+	case "vSphere", "None":
```
frobware: I think this would better expressed using the existing constants:

```diff
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -176,8 +176,8 @@ func (optr *Operator) sync(key string) error {
                glog.Errorf("Failed getting operator config: %v", err)
                return err
        }
-       switch p := operatorConfig.Controllers.Provider; p {
-       case "vSphere", "None":
+       switch p := Provider(operatorConfig.Controllers.Provider); p {
+       case VSphereProvider, NoneProvider:
                glog.V(1).Infof("%q provider specified, reporting available", p)
                if err := optr.statusAvailable(); err != nil {
                        glog.Errorf("Error syncing ClusterOperatorStatus: %v", err)
```

frobware: I also experimented by removing the two cases that are handled to see if the existing unit tests failed - they did not. Can we add a unit test for the vSphere case?

pkg/operator/config.go
frobware: This should also use the existing constant:

```diff
--- a/pkg/operator/config.go
+++ b/pkg/operator/config.go
@@ -175,7 +175,7 @@ func getProviderControllerFromImages(provider Provider, images Images) (string,
        case KubemarkProvider:
                return images.ClusterAPIControllerKubemark, nil
        case VSphereProvider:
-               return "vSphere", nil
+               return string(VSphereProvider), nil
        case BareMetalProvider:
                return images.ClusterAPIControllerBareMetal, nil
        case NoneProvider:
```

frobware: Oh - but this is "vSphere" and the Provider constant is "vsphere". Which wins?

pkg/operator/operator.go
frobware: And same question again regarding "vSphere" versus "vsphere".

pkg/operator/config.go
enxebre: the images values comes from a configMap which images are replaced by automation when building the release payload based on the mapping imageStream here https://github.com/openshift/machine-api-operator/blob/master/install/image-references
So hardcoding here for those not wanting machine API controllers/images is fine

pkg/operator/operator.go
enxebre: let's use VSphere everywhere for consistency, we can clean up and indeed unexport all those const in a follow up

frobware: > let's use VSphere everywhere for consistency, we can clean up and indeed unexport all those const in a follow up

So defined once as "VSphere" but referenced symbolically throughout. Agreed? Or am I missing the obvious?

pkg/operator/config.go
staebler: I was following the convention laid done by the None provider. That provider uses lowercase for the provider name and mixed case for the controller name. The ofiicial name of the vSphere product uses a lowercase for the v, so that is why I choose vSphere over VSphere.

pkg/operator/operator.go
staebler: It seems odd to me that we would use mxed case for VSphere while all of the other providers use strictly lowercase: aws, libvirt, openstack, kubemark, baremetal, and none.

staebler: > I also experimented by removing the two cases that are handled to see if the existing unit tests failed - they did not. Can we add a unit test for the vSphere case?

Sure. I'll write up some unit tests for the operator.

pkg/operator/config.go
staebler: Resolved by cd6ac6f...3844ccf

pkg/operator/operator.go
staebler: Resolved by cd6ac6f...3844ccf

enxebre:
since https://github.com/openshift/machine-api-operator/pull/264 got merged. Once this get's merged https://github.com/openshift/machine-api-operator/pull/266 vSphere will already no-op and we'll just need to include the relevant unit tests

staebler:
9114986...2dbe66c brings this PR in line with the changes from #264 and #266.

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,169 @@
...
+			}
+			imagesAsJSON, err := json.Marshal(images)
+			if err != nil {
+				t.Fatalf("failed to marshal images: %v", err)
+			}
+			imagesFilePath := filepath.Join(tempDir, "test-images.json")
+			if err := ioutil.WriteFile(imagesFilePath, imagesAsJSON, 0666); err != nil {
+				t.Fatalf("could not write the images file: %v", err)
+			}
+
```
enxebre: from 108 to this line could actually go outside the loop?

pkg/operator/operator_test.go
```diff
@@ -0,0 +1,169 @@
...
+			optr := Operator{
+				eventRecorder: record.NewFakeRecorder(5),
+				osClient:      fakeconfigclientset.NewSimpleClientset(infra),
+				imagesFile:    imagesFilePath,
+			}
+
+			err = optr.sync("test-key")
+			if tc.expectedSyncError {
+				if !assert.Error(t, err, "unexpected sync success") {
+					t.Fatal()
```
enxebre: why `Fatal`? from the docs: "it will call `failNow` and stops function execution", why not  just asset.Error so the loop keep going and we output for each case
cc @frobware

staebler: With sub-tests, the call to `Fatal` will only fail the one sub-test. The rest of the sub-tests will still execute.

pkg/operator/operator_test.go
staebler: Sure, it could. Is there a compelling reason why it would be better to introduce dependency between the sub-tests by moving setup code outside of the sub-tests? The only reason that I can think of is to save the time of creating the directory and creating the images file. That does not seem like a great reason to me.

staebler: Moved in 2dbe66c...d28b7cb.

staebler:
d28b7cb...67a8e84 simplified the test cases for the operator sync test.

pkg/operator/operator_test.go
enxebre: so what's the different over just `if !tc.expectedNoop { assert.Error(t, err, "unexpected sync success")}`

enxebre:
thanks a lot @staebler 
/approve

pkg/operator/operator_test.go
staebler: The difference is that with your code, the rest of the sub-test will be executed. So there will be more test errors later on in the sub-test when it comes to testing the conditions set on the clusteroperator. That is not terribly; it does make the test results harder to analyze due to more noise, however.

I am fine with removing the call to `Fatal` here, if you prefer. The call to `Fatal` at https://github.com/openshift/machine-api-operator/blob/67a8e84e4a5b7105e75f6d0cb4db777c61e3231d/pkg/operator/operator_test.go#L108-L111 needs to stay to prevent a panic when the clusteroperator does not exist.

enxebre:
needs rebase

staebler:
> needs rebase

67a8e84...12e66da

enxebre:
we are hitting https://bugzilla.redhat.com/show_bug.cgi?id=1698624 re-triggering for CI measurement

enxebre:
this should go green now https://github.com/openshift/machine-api-operator/pull/286
/retest

PR #256, opened 2019-03-26
operator: use infrastructure.config.openshfit.io for platformtype

abhinavdahiya:
cluster operators must use the config.openshit.io as source for their top level config.

/cc @derekwaynecarr @enxebre

PR #264, opened 2019-04-01
operator: use infrastructure.config.openshfit.io for platformtype

enxebre:
cluster operators must use the config.openshit.io as source for their top level config.
Rebase and follow up for https://github.com/openshift/machine-api-operator/pull/256

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,22 +161,36 @@ INFO: For development and testing purposes only
    $ kustomize build config | kubectl apply -f -
    ```
 
-3. Create `cluster-config-v1` configmap to tell the MAO to deploy `kubemark` provider:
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -161,22 +161,36 @@ INFO: For development and testing purposes only
...
+     versions:
+     - name: v1
+       served: true
+   storage: true
+   ---
+   apiVersion: config.openshift.io/v1
+   kind: Infrastructure
+   metadata:
+     name: cluster
+   status:
```
michaelgugino: Should this be status or spec?  Seems configurables should be in spec not status.

enxebre: we are just consumers here https://github.com/openshift/installer/blob/master/pkg/asset/manifests/infrastructure.go#L67

PR #265, opened 2019-04-01
Register ClusterOperator to properly generate event for the kind

ingvagabund:
```
E0401 08:06:27.851998       1 event.go:259] Could not construct reference to: '&v1.ClusterOperator{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"machine-api", GenerateName:"", Namespace:"", SelfLink:"/apis/config.openshift.io/v1/clusteroperators/machine-api", UID:"0cfdf4cd-5455-11e9-a124-0adb1443898a", ResourceVersion:"656", Generation:1, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:63689702787, loc:(*time.Location)(0x1d1cea0)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string(nil), OwnerReferences:[]v1.OwnerReference(nil), Initializers:(*v1.Initializers)(nil), Finalizers:[]string(nil), ClusterName:""}, Spec:v1.ClusterOperatorSpec{}, Status:v1.ClusterOperatorStatus{Conditions:[]v1.ClusterOperatorStatusCondition(nil), Versions:[]v1.OperandVersion(nil), RelatedObjects:[]v1.ObjectReference(nil), Extension:runtime.RawExtension{Raw:[]uint8(nil), Object:runtime.Object(nil)}}}' due to: 'no kind is registered for the type v1.ClusterOperator in scheme "github.com/openshift/cluster-api/pkg/client/clientset_generated/clientset/scheme/register.go:31"'. Will not report event: 'Normal' 'Status upgrade' 'Progressing towards operator: 0.0.1-snapshot'
E0401 08:13:09.777804       1 event.go:259] Could not construct reference to: '&v1.ClusterOperator{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"machine-api", GenerateName:"", Namespace:"", SelfLink:"/apis/config.openshift.io/v1/clusteroperators/machine-api", UID:"0cfdf4cd-5455-11e9-a124-0adb1443898a", ResourceVersion:"2856", Generation:1, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:63689702787, loc:(*time.Location)(0x1d1cea0)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string(nil), OwnerReferences:[]v1.OwnerReference(nil), Initializers:(*v1.Initializers)(nil), Finalizers:[]string(nil), ClusterName:""}, Spec:v1.ClusterOperatorSpec{}, Status:v1.ClusterOperatorStatus{Conditions:[]v1.ClusterOperatorStatusCondition{v1.ClusterOperatorStatusCondition{Type:"Progressing", Status:"False", LastTransitionTime:v1.Time{Time:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}}, Reason:"SyncingResources", Message:"Running resync for operator: 0.0.1-snapshot"}, v1.ClusterOperatorStatusCondition{Type:"Available", Status:"True", LastTransitionTime:v1.Time{Time:time.Time{wall:0x0, ext:63689702787, loc:(*time.Location)(0x1d1cea0)}}, Reason:"", Message:""}, v1.ClusterOperatorStatusCondition{Type:"Failing", Status:"False", LastTransitionTime:v1.Time{Time:time.Time{wall:0x0, ext:63689702787, loc:(*time.Location)(0x1d1cea0)}}, Reason:"", Message:""}}, Versions:[]v1.OperandVersion{v1.OperandVersion{Name:"operator", Version:"0.0.1-snapshot"}}, RelatedObjects:[]v1.ObjectReference(nil), Extension:runtime.RawExtension{Raw:[]uint8(nil), Object:runtime.Object(nil)}}}' due to: 'no kind is registered for the type v1.ClusterOperator in scheme "github.com/openshift/cluster-api/pkg/client/clientset_generated/clientset/scheme/register.go:31"'. Will not report event: 'Warning' 'Status failing' 'timed out waiting for the condition'
E0401 08:13:09.781746       1 sync.go:36] Error syncing cluster api controller: timed out waiting for the condition
```

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -11,6 +11,7 @@ import (
 	osconfigv1 "github.com/openshift/api/config/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	appsinformersv1 "k8s.io/client-go/informers/apps/v1"
@@ -20,8 +21,6 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
-
-	"github.com/openshift/cluster-api/pkg/client/clientset_generated/clientset/scheme"
 )
 
 const (
@@ -77,13 +76,16 @@ func New(
 		operandVersions = append(operandVersions, osconfigv1.OperandVersion{Name: "operator", Version: releaseVersion})
 	}
 
+	eventRecorderScheme := runtime.NewScheme()
+	osconfigv1.Install(eventRecorderScheme)
+
 	optr := &Operator{
 		namespace:       namespace,
 		name:            name,
 		imagesFile:      imagesFile,
 		kubeClient:      kubeClient,
 		osClient:        osClient,
-		eventRecorder:   eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: "machineapioperator"}),
+		eventRecorder:   eventBroadcaster.NewRecorder(eventRecorderScheme, v1.EventSource{Component: "machineapioperator"}),
 		queue:           workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "machineapioperator"),
 		operandVersions: operandVersions,
 	}


cmd/machine-api-operator/main.go
```diff
@@ -24,6 +27,8 @@ var (
 )
 
 func init() {
+	osconfigv1.AddToScheme(scheme.Scheme)
```
bison: Outside of tests where it's a quick-fix, I'm not sure `init()` is the right place for this. Shouldn't it happen in `operator.New()` or something?

enxebre: this won't be picked here https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/operator.go#L23
and https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/operator.go#L85

ingvagabund: Danke

ingvagabund: `init()` is called only once, `operator.New` can be called multiple times. Imho, `AddToSheme` will fail if the same type is registered twice.

pkg/operator/operator.go
```diff
@@ -33,6 +33,10 @@ const (
 	ownedManifestsDir = "owned-manifests"
 )
 
+func init() {
+	osconfigv1.AddToScheme(scheme.Scheme)
```
bison: Right package, but I still don't think this belongs in `init()`. Why not just call this in `New()`?

bison: > init() is called only once, operator.New can be called multiple times. Imho, AddToSheme will fail if the same type is registered twice.

I really don't like putting stuff in `init()` here personally. We could set up a new scheme to add these types to and pass that when creating the event recorder, right? That would solve the problem of being called more than once.

ingvagabund: Moved under `New()`.

PR #270, opened 2019-04-02
Remove not yet supported CRDs

enxebre:
As we go more stable we want to reduce the surface area for not yet supported resources until there's a solid e2e story for them.
Removing classes, ~~deployments~~ and clusters CRDs


enxebre:
cc @derekwaynecarr @bison @ingvagabund @frobware

ingvagabund:
/approve

not breaking anything afaik

frobware:
Pretty sure this will break cluster-autoscaler.

ingvagabund:
@frobware for the cluster-autoscaler + machine deployments

ingvagabund:
> Pretty sure this will break cluster-autoscaler.

Machine deployments should be optional-ish, right?

frobware:
> Machine deployments should be optional-ish, right?

They are and we disable looking for them by default. But we do create a machine deployment informer which will presumably now fail.

frobware:
/hold

Pretty sure this will now break the autoscaler.

ingvagabund:
/approve cancel

until properly resolved in CA

enxebre:
keeping deployments for now
/hold cancel

PR #272, opened 2019-04-04
Update LastTransitionTime at OperatorStatus Condition

vikaschoudhary16:
bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1695197

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -61,20 +61,9 @@ func (optr *Operator) statusProgressing() error {
 	}
 
 	conds := []osconfigv1.ClusterOperatorStatusCondition{
-		{
-			Type:    osconfigv1.OperatorProgressing,
-			Status:  isProgressing,
-			Reason:  string(ReasonSyncing),
-			Message: message,
-		},
-		{
-			Type:   osconfigv1.OperatorAvailable,
-			Status: osconfigv1.ConditionTrue,
-		},
-		{
-			Type:   osconfigv1.OperatorFailing,
-			Status: osconfigv1.ConditionFalse,
-		},
+		newClusterOperatorStatusCondition(osconfigv1.OperatorProgressing, isProgressing, string(ReasonSyncing), message),
+		newClusterOperatorStatusCondition(osconfigv1.OperatorAvailable, osconfigv1.ConditionTrue, "", ""),
+		newClusterOperatorStatusCondition(osconfigv1.OperatorFailing, osconfigv1.ConditionFalse, "", ""),
 	}
 
 	return optr.syncStatus(co, conds)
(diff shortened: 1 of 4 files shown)


bison:
Maybe this is fine for now, but it means `LastTransitionTime` isn't exactly what it says, right? This updates the timestamp regardless of whether the condition transitioned to a new state. Here's how we're handling that in the cluster-autoscaler-operator:

https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/operator/status.go#L124-L145

vikaschoudhary16:
@bison [SetOperatorStatusCondition](https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/status.go#L163), updates time only when condition state transitions.
https://github.com/openshift/machine-api-operator/blob/master/vendor/github.com/openshift/cluster-version-operator/lib/resourcemerge/os.go#L52

Also this has been covered in this unit test as well, https://github.com/openshift/machine-api-operator/blob/master/vendor/github.com/openshift/cluster-version-operator/lib/resourcemerge/os.go#L52

bison:
Ah, sorry. I misread because `EnsureClusterOperatorStatus()` does not handle that.

So, is there a case where we report available after an upgrade without ever transitioning into a "Progressing" state? I think that might be the case with a no-op provider, right? My read of the bug is that we're still required to reset the transition time on the Progressing condition in that case -- even though it never transitioned to `True` status.

vikaschoudhary16:
@bison good catch!! PTAL https://github.com/openshift/machine-api-operator/pull/272/commits/7b6c0d02bad76d41f3ff5358de2e31666e9de977

pkg/operator/status.go
```diff
@@ -157,6 +125,18 @@ func (optr *Operator) statusFailing(error string) error {
 	return optr.syncStatus(co, conds)
 }
 
+func newClusterOperatorStatusCondition(conditionType osconfigv1.ClusterStatusConditionType,
+	conditionStatus osconfigv1.ConditionStatus, reason string,
+	message string) osconfigv1.ClusterOperatorStatusCondition {
+	return osconfigv1.ClusterOperatorStatusCondition{
+		Type:               conditionType,
+		Status:             conditionStatus,
+		LastTransitionTime: metav1.Now(),
```
michaelgugino: Since we're updating the LastTransitionTime on these statuscondition objects all the time now, we need to ensure we don't update the LastTransitionTime when we don't desire.  EG, we don't want to update this on available every time the loop runs.  We should add some unit testing around this.

vikaschoudhary16: @michaelgugino it will not update when when the condition status is not changing. I already added a unit test for the same.
https://github.com/openshift/machine-api-operator/blob/master/vendor/github.com/openshift/cluster-version-operator/lib/resourcemerge/os.go#L52

PR #273, opened 2019-04-07
Integrate machine disruption with machine health check

cynepco3hahue:
This PR integrates `MachineDisruptionBudget` check into the machine health check controller, it makes possible to prevent machine deletion if the cluster has not enough healthy machines.

### DIFF
--- a/pkg/controller/disruption/disruption.go
+++ b/pkg/controller/disruption/disruption.go
@@ -9,14 +9,16 @@ import (
 	mapiv1 "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	healthcheckingv1alpha1 "github.com/openshift/machine-api-operator/pkg/apis/healthchecking/v1alpha1"
 	machineutil "github.com/openshift/machine-api-operator/pkg/util/machines"
-	apiequality "k8s.io/apimachinery/pkg/api/equality"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/retry"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -26,16 +28,31 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-// DeletionTimeout sets maximum time from the moment a machine is added to DisruptedMachines in MDB.Status
-// to the time when the machine is expected to be seen by MDB controller as having been marked for deletion.
-// If the machine was not marked for deletion during that time it is assumed that it won't be deleted at
-// all and the corresponding entry can be removed from mdb.Status.DisruptedMachines. It is assumed that
-// machine/mdb apiserver to controller latency is relatively small (like 1-2sec) so the below value should
-// be more than enough.
-// If the controller is running on a different node it is important that the two nodes have synced
-// clock (via ntp for example). Otherwise MachineDisruptionBudget controller may not provide enough
-// protection against unwanted pod disruptions.
-const DeletionTimeout = 2 * time.Minute
+const (
+	// DeletionTimeout sets maximum time from the moment a machine is added to DisruptedMachines in MDB.Status
+	// to the time when the machine is expected to be seen by MDB controller as having been marked for deletion.
+	// If the machine was not marked for deletion during that time it is assumed that it won't be deleted at
+	// all and the corresponding entry can be removed from mdb.Status.DisruptedMachines. It is assumed that
+	// machine/mdb apiserver to controller latency is relatively small (like 1-2sec) so the below value should
+	// be more than enough.
+	// If the controller is running on a different node it is important that the two nodes have synced
+	// clock (via ntp for example). Otherwise MachineDisruptionBudget controller may not provide enough
+	// protection against unwanted machine disruptions.
+	DeletionTimeout = 2 * time.Minute
+	// maxDisruptedMachinSize is the max size of MachineDisruptionBudgetStatus.DisruptedMachines.
+	// MachineHealthCheck will refuse to delete machine covered by the corresponding MDB
+	// if the size of the map exceeds this value.
+	maxDisruptedMachinSize = 50
+)
+
+// updateMDBRetry is the retry for a conflict where multiple clients
+// are making changes to the same resource.
+var updateMDBRetry = wait.Backoff{
+	Steps:    20,
+	Duration: 500 * time.Millisecond,
+	Factor:   1.0,
+	Jitter:   0.1,
+}
 
 // Add creates a new MachineDisruption Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and start it when the Manager is started.
@@ -326,7 +343,7 @@ func (r *ReconcileMachineDisruption) updateMachineDisruptionBudgetStatus(
 		ObservedGeneration:        mdb.Generation,
 	}
 
-	return r.client.Update(context.TODO(), newMdb)
+	return r.client.Status().Update(context.TODO(), newMdb)
 }
 
 // failSafe is an attempt to at least update the MachineDisruptionsAllowed field to
@@ -336,7 +353,7 @@ func (r *ReconcileMachineDisruption) updateMachineDisruptionBudgetStatus(
 func (r *ReconcileMachineDisruption) failSafe(mdb *healthcheckingv1alpha1.MachineDisruptionBudget) error {
 	newMdb := mdb.DeepCopy()
 	mdb.Status.MachineDisruptionsAllowed = 0
-	return r.client.Update(context.TODO(), newMdb)
+	return r.client.Status().Update(context.TODO(), newMdb)
 }
 
 func (r *ReconcileMachineDisruption) getMachineDisruptionBudgetForMachine(machine *mapiv1.Machine) *healthcheckingv1alpha1.MachineDisruptionBudget {
@@ -440,4 +457,73 @@ func (r *ReconcileMachineDisruption) machineToMachineDisruptionBudget(o handler.
 
 	name := client.ObjectKey{Namespace: mdb.Namespace, Name: mdb.Name}
 	return []reconcile.Request{{NamespacedName: name}}
+}
+
+// isMachineDisruptionAllowed returns true if the provided MachineDisruptionBudget allows any disruption
+func isMachineDisruptionAllowed(mdb *healthcheckingv1alpha1.MachineDisruptionBudget, maxDisruptedMachinSize int) bool {
+	if mdb.Status.ObservedGeneration < mdb.Generation {
+		glog.Warningf("The machine disruption budget %s is still being processed by the server", mdb.Name)
+		return false
+	}
+	if mdb.Status.MachineDisruptionsAllowed < 0 {
+		glog.Warningf("The machine disruption budget %s MachineDisruptionsAllowed is negative", mdb.Name)
+		return false
+	}
+	if len(mdb.Status.DisruptedMachines) > maxDisruptedMachinSize {
+		glog.Warningf("The machine disruption budget %s DisruptedMachines map too big - too many deletions not confirmed by MDB controller", mdb.Name)
+		return false
+	}
+	if mdb.Status.MachineDisruptionsAllowed == 0 {
+		glog.Warningf("Cannot remediate machine as it would violate the machine's disruption budget %s", mdb.Name)
+		return false
+	}
+
+	return true
+}
+
+func decrementMachineDisruptionsAllowed(c client.Client, machineName string, mdb *healthcheckingv1alpha1.MachineDisruptionBudget) error {
+	mdb.Status.MachineDisruptionsAllowed--
+	if mdb.Status.DisruptedMachines == nil {
+		mdb.Status.DisruptedMachines = make(map[string]metav1.Time)
+	}
+	// MachineHealthCheck controller needs to inform the MDB controller that it is about to remediate a machine
+	// so it should not consider it as available in calculations when updating MachineDisruptions allowed.
+	// If the machine is not remediated within a reasonable time limit MDB controller will assume that it won't
+	// be remediated at all and remove it from DisruptedMachines map.
+	mdb.Status.DisruptedMachines[machineName] = metav1.Time{Time: time.Now()}
+	return c.Status().Update(context.TODO(), mdb)
+}
+
+// RetryDecrementMachineDisruptionsAllowed validates if the disruption is allowed, when it allowed it will decrement
+// MDB MachineDisruptionsAllowed parameter and update the status of the MDB, in case when update failed
+// on the conflict it will try again with the backoff
+func RetryDecrementMachineDisruptionsAllowed(c client.Client, machine *mapiv1.Machine) error {
+	var mdb *healthcheckingv1alpha1.MachineDisruptionBudget
+	err := retry.RetryOnConflict(updateMDBRetry, func() error {
+		mdbs, err := machineutil.GetMachineMachineDisruptionBudgets(c, machine)
+		if err != nil {
+			return err
+		}
+
+		if len(mdbs) > 1 {
+			return fmt.Errorf("machine %q has more than one MachineDisruptionBudget, which is not supported", machine.Name)
+		}
+
+		if len(mdbs) == 1 {
+			mdb = mdbs[0]
+
+			if !isMachineDisruptionAllowed(mdb, maxDisruptedMachinSize) {
+				return fmt.Errorf("machine disruption is not allowed")
+			}
+
+			return decrementMachineDisruptionsAllowed(c, machine.Name, mdb)
+		}
+		return nil
+	})
+
+	if err == wait.ErrWaitTimeout {
+		err = fmt.Errorf("couldn't update MachineDisruptionBudget %q due to conflicts", mdb.Name)
+	}
+
+	return err
 }
--- a/pkg/controller/disruption/disruption.go
+++ b/pkg/controller/disruption/disruption.go
@@ -102,7 +102,7 @@ type ReconcileMachineDisruption struct {
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileMachineDisruption) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	glog.Infof("Reconciling MachineDisruption triggered by %s/%s\n", request.Namespace, request.Name)
+	glog.V(4).Infof("Reconciling MachineDisruption triggered by %s/%s\n", request.Namespace, request.Name)
 
 	// Get machine from request
 	mdb := &healthcheckingv1alpha1.MachineDisruptionBudget{}
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -2,13 +2,13 @@ package machinehealthcheck
 
 import (
 	"context"
-
 	golangerrors "errors"
 	"time"
 
 	"github.com/golang/glog"
 	mapiv1 "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	healthcheckingv1alpha1 "github.com/openshift/machine-api-operator/pkg/apis/healthchecking/v1alpha1"
+	"github.com/openshift/machine-api-operator/pkg/controller/disruption"
 	"github.com/openshift/machine-api-operator/pkg/util/conditions"
 
 	corev1 "k8s.io/api/core/v1"
@@ -200,20 +200,27 @@ func remediate(r *ReconcileMachineHealthCheck, remediationStrategy *healthchecki
 		// apply remediation logic, if at least one condition last more than specified timeout
 		// specific remediation logic goes here
 		if unhealthyForTooLong(nodeCondition, conditionTimeout) {
+			// do not fail immediatlty, but try again if the method fails because of the update conflict
+			if err = disruption.RetryDecrementMachineDisruptionsAllowed(r.client, machine); err != nil {
+				// if the error appears here it means that machine healthcheck operation restricted by machine
+				// disruption budget, in this case we want to re-try after one minute
+				glog.Warning(err)
+				return reconcile.Result{Requeue: true, RequeueAfter: time.Minute}, nil
+			}
+
 			if remediationStrategy != nil && *remediationStrategy == remediationStrategyReboot {
 				return r.remediationStrategyReboot(machine, node)
-			} else {
-				if isMaster(*machine, r.client) {
-					glog.Infof("The machine %s is a master node, skipping remediation", machine.Name)
-					return reconcile.Result{}, nil
-				}
-				glog.Infof("Machine %s has been unhealthy for too long, deleting", machine.Name)
-				if err := r.client.Delete(context.TODO(), machine); err != nil {
-					glog.Errorf("Failed to delete machine %s, requeuing referenced node", machine.Name)
-					return reconcile.Result{}, err
-				}
+			}
+			if isMaster(*machine, r.client) {
+				glog.Infof("The machine %s is a master node, skipping remediation", machine.Name)
 				return reconcile.Result{}, nil
 			}
+			glog.Infof("Machine %s has been unhealthy for too long, deleting", machine.Name)
+			if err := r.client.Delete(context.TODO(), machine); err != nil {
+				glog.Errorf("Failed to delete machine %s, requeuing referenced node", machine.Name)
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{}, nil
 		}
 
 		now := time.Now()
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
@@ -11,6 +11,7 @@ import (
 	healthcheckingv1alpha1 "github.com/openshift/machine-api-operator/pkg/apis/healthchecking/v1alpha1"
 	"github.com/openshift/machine-api-operator/pkg/util/conditions"
 	maotesting "github.com/openshift/machine-api-operator/pkg/util/testing"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
--- a/pkg/util/machines/machines.go
+++ b/pkg/util/machines/machines.go
@@ -19,7 +19,7 @@ import (
 // IsMachineHealthy returns true if the the machine is running and machine node is healthy
 func IsMachineHealthy(c client.Client, machine *mapiv1.Machine) bool {
 	if machine.Status.NodeRef == nil {
-		glog.Infof("machine %s does not have NodeRef", machine.Name)
+		glog.V(4).Infof("machine %s does not have NodeRef", machine.Name)
 		return false
 	}
 
@@ -33,12 +33,12 @@ func IsMachineHealthy(c client.Client, machine *mapiv1.Machine) bool {
 
 	readyCond := conditions.GetNodeCondition(node, v1.NodeReady)
 	if readyCond == nil {
-		glog.Infof("node %s does have 'Ready' condition", machine.Name)
+		glog.V(4).Infof("node %s does have 'Ready' condition", machine.Name)
 		return false
 	}
 
 	if readyCond.Status != v1.ConditionTrue {
-		glog.Infof("node %s does have has 'Ready' condition with the status %s", machine.Name, readyCond.Status)
+		glog.V(4).Infof("node %s does have has 'Ready' condition with the status %s", machine.Name, readyCond.Status)
 		return false
 	}
 	return true
--- a/install/0000_30_machine-api-operator_07_machinehealthcheck.crd.yaml
+++ b/install/0000_30_machine-api-operator_07_machinehealthcheck.crd.yaml
@@ -14,6 +14,8 @@ spec:
     - mhc
     - mhcs
   scope: Namespaced
+  subresources:
+    status: {}
   validation:
     openAPIV3Schema:
       properties:
--- a/install/0000_30_machine-api-operator_08_machinedisruptionbudget.crd.yaml
+++ b/install/0000_30_machine-api-operator_08_machinedisruptionbudget.crd.yaml
@@ -27,6 +27,8 @@ spec:
     - mdb
     - mdbs
   scope: Namespaced
+  subresources:
+    status: {}
   validation:
     openAPIV3Schema:
       properties:
@@ -67,7 +69,15 @@ spec:
         status:
           description: Most recently observed status of the MachineDisruptionBudget.
           properties:
-            DisruptedMachines:
+            currentHealthy:
+              description: current number of healthy machines
+              format: int32
+              type: integer
+            desiredHealthy:
+              description: minimum desired number of healthy machines
+              format: int32
+              type: integer
+            disruptedMachines:
               description: DisruptedMachines contains information about machines whose
                 deletion was processed by the API server but has not yet been observed
                 by the MachineDisruptionBudget controller. A machine will be in this
@@ -81,14 +91,6 @@ spec:
                 goes smooth this map should be empty for the most of the time. Large
                 number of entries in the map may indicate problems with machines deletions.
               type: object
-            currentHealthy:
-              description: current number of healthy machines
-              format: int32
-              type: integer
-            desiredHealthy:
-              description: minimum desired number of healthy machines
-              format: int32
-              type: integer
             disruptionsAllowed:
               description: Number of machines disruptions that are currently allowed.
               format: int32
--- a/pkg/apis/healthchecking/v1alpha1/machinedisruptionbudget_types.go
+++ b/pkg/apis/healthchecking/v1alpha1/machinedisruptionbudget_types.go
@@ -44,7 +44,7 @@ type MachineDisruptionBudgetStatus struct {
 	// If everything goes smooth this map should be empty for the most of the time.
 	// Large number of entries in the map may indicate problems with machines deletions.
 	// +optional
-	DisruptedMachines map[string]metav1.Time `json:"DisruptedMachines,omitempty" protobuf:"bytes,2,rep,name=DisruptedMachines"`
+	DisruptedMachines map[string]metav1.Time `json:"disruptedMachines,omitempty" protobuf:"bytes,2,rep,name=DisruptedMachines"`
 
 	// Number of machines disruptions that are currently allowed.
 	MachineDisruptionsAllowed int32 `json:"disruptionsAllowed" protobuf:"varint,3,opt,name=disruptionsAllowed"`
@@ -63,7 +63,7 @@ type MachineDisruptionBudgetStatus struct {
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // MachineDisruptionBudget is an object to define the max disruption that a collection of machines can experience
-// kubebuilder:subresource:status
+// +kubebuilder:subresource:status
 // +kubebuilder:resource:shortName=mdb;mdbs
 // +kubebuilder:printcolumn:name="Healthy",type="integer",JSONPath=".status.currentHealthy",description="The number of healthy machines"
 // +kubebuilder:printcolumn:name="Total",type="integer",JSONPath=".status.expectedMachines",description="The total number of machines"
--- a/pkg/apis/healthchecking/v1alpha1/machinehealthcheck_types.go
+++ b/pkg/apis/healthchecking/v1alpha1/machinehealthcheck_types.go
@@ -14,7 +14,7 @@ type RemediationStrategyType string
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // MachineHealthCheck is the Schema for the machinehealthchecks API
-// kubebuilder:subresource:status
+// +kubebuilder:subresource:status
 // +kubebuilder:resource:shortName=mhc;mhcs
 // +k8s:openapi-gen=true
 type MachineHealthCheck struct {
--- a/pkg/controller/disruption/disruption_test.go
+++ b/pkg/controller/disruption/disruption_test.go
@@ -645,3 +645,97 @@ func TestReconcile(t *testing.T) {
 		}
 	}
 }
+
+func TestRepeatCheckAndDecrement(t *testing.T) {
+	node := maotesting.NewNode("node", true)
+	machine := maotesting.NewMachine("machine", node.Name)
+	machineWithoutLabels := maotesting.NewMachine("machineWithoutLabels", node.Name)
+	machineWithoutLabels.Labels = map[string]string{}
+
+	mdbObserverGeneration := maotesting.NewMinAvailableMachineDisruptionBudget(1)
+	mdbObserverGeneration.Generation = 2
+	mdbObserverGeneration.Status.ObservedGeneration = 1
+
+	mdbDisruptionAllowedZero := maotesting.NewMinAvailableMachineDisruptionBudget(1)
+	mdbDisruptionAllowedZero.Status.MachineDisruptionsAllowed = 0
+
+	mdbDisruptionAllowedLessThanZero := maotesting.NewMinAvailableMachineDisruptionBudget(1)
+	mdbDisruptionAllowedLessThanZero.Status.MachineDisruptionsAllowed = -1
+
+	mdb1 := maotesting.NewMinAvailableMachineDisruptionBudget(1)
+	mdb1.Name = "mdb1"
+	mdb1.Status.MachineDisruptionsAllowed = 1
+
+	mdb2 := maotesting.NewMinAvailableMachineDisruptionBudget(1)
+	mdb1.Name = "mdb2"
+	mdb2.Status.MachineDisruptionsAllowed = 1
+
+	testsCases := []struct {
+		testName string
+		machine  *mapiv1.Machine
+		mdbs     []*healthcheckingv1alpha1.MachineDisruptionBudget
+		error    bool
+	}{
+		{
+			testName: "With the machine without labels",
+			machine:  machineWithoutLabels,
+			mdbs:     []*healthcheckingv1alpha1.MachineDisruptionBudget{mdb1},
+			error:    true,
+		},
+		{
+			testName: "Without MDB",
+			machine:  machine,
+			mdbs:     []*healthcheckingv1alpha1.MachineDisruptionBudget{},
+			error:    false,
+		},
+		{
+			testName: "With two MDB's",
+			machine:  machine,
+			mdbs:     []*healthcheckingv1alpha1.MachineDisruptionBudget{mdb1, mdb2},
+			error:    true,
+		},
+		{
+			testName: "With the MDB that has wrong observedGeneration",
+			machine:  machine,
+			mdbs:     []*healthcheckingv1alpha1.MachineDisruptionBudget{mdbObserverGeneration},
+			error:    true,
+		},
+		{
+			testName: "With the MDB with zero DisruptionAllowed",
+			machine:  machine,
+			mdbs:     []*healthcheckingv1alpha1.MachineDisruptionBudget{mdbDisruptionAllowedZero},
+			error:    true,
+		},
+		{
+			testName: "With the MDB with less than zero DisruptionAllowed",
+			machine:  machine,
+			mdbs:     []*healthcheckingv1alpha1.MachineDisruptionBudget{mdbDisruptionAllowedLessThanZero},
+			error:    true,
+		},
+		{
+			testName: "With the correct MDB",
+			machine:  machine,
+			mdbs:     []*healthcheckingv1alpha1.MachineDisruptionBudget{mdb1},
+			error:    false,
+		},
+	}
+
+	for _, tc := range testsCases {
+		objects := []runtime.Object{}
+		objects = append(objects, tc.machine)
+
+		for _, mdb := range tc.mdbs {
+			objects = append(objects, mdb)
+		}
+
+		r := newFakeReconciler(nil, objects...)
+		err := RetryDecrementMachineDisruptionsAllowed(r.client, tc.machine)
+		if tc.error != (err != nil) {
+			var errorExpectation string
+			if !tc.error {
+				errorExpectation = "no"
+			}
+			t.Errorf("Test case: %s. Expected: %s error, got: %v", tc.testName, errorExpectation, err)
+		}
+	}
+}


pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +378,62 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
 	}
 	return false
 }
+
+// checkAndDecrement checks if the provided MachineDisruptionBudget allows any disruption
+func (r *ReconcileMachineHealthCheck) checkAndDecrement(machineName string, mdb *healthcheckingv1alpha1.MachineDisruptionBudget, maxDisruptedMachinSize int) error {
+	if mdb.Status.ObservedGeneration < mdb.Generation {
```
beekhof: Suggestion, be pessimistic and respect the  .MachineDisruptionsAllowed < 0 and .DisruptedMachines > max conditions even if the values we are checking are old

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +378,62 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
+
+	mdb.Status.MachineDisruptionsAllowed--
+	if mdb.Status.DisruptedMachines == nil {
+		mdb.Status.DisruptedMachines = make(map[string]metav1.Time)
+	}
+	// MachineHealthCheck controller needs to inform the MDB controller that it is about to delete a machine
+	// so it should not consider it as available in calculations when updating MachineDisruptions allowed.
+	// If the machine is not deleted within a reasonable time limit MDB controller will assume that it won't
+	// be deleted at all and remove it from DisruptedMachines map.
+	mdb.Status.DisruptedMachines[machineName] = metav1.Time{Time: time.Now()}
```
beekhof: I'm less keen about this part but perhaps it unavoidable when not using the Ready condition.
It would have been nice if MDB could automagically exclude "unhealthy" nodes from its calculations.

Could/should we have MDB consume the fencing config map?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
cynepco3hahue: if this method returns an error, the MHC will skip fencing anyway

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
cynepco3hahue: I think we have here two questions:

- all this mdb.Status.DisruptedMachines does not regarding health and unhealthy machines, but regarding the delay in the deletion of the machine so we do want to take into account machines that will soon be deleted, and if it does not be deleted after the specified timeout we start to take them in the account again
- I will use the node conditions config map under the MDB in the next iteration

pkg/controller/disruption/disruption.go
```diff
@@ -98,7 +98,7 @@ type ReconcileMachineDisruption struct {
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileMachineDisruption) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	glog.Infof("Reconciling MachineDisruption triggered by %s/%s\n", request.Namespace, request.Name)
+	glog.V(4).Infof("Reconciling MachineDisruption triggered by %s/%s\n", request.Namespace, request.Name)
```
enxebre: this does not seems related to the title of this commit

pkg/controller/disruption/disruption.go
```diff
@@ -311,11 +311,13 @@ func (r *ReconcileMachineDisruption) updateMachineDisruptionBudgetStatus(
 	expectedCount int32,
 	disruptedMachines map[string]metav1.Time) error {
 
+	// we add one because we do not want to respect disruption budget when expected and healthy are equal
+	disruptionsAllowed := currentHealthy - desiredHealthy + 1
```
enxebre: this does not seems related to the title of this commit

pkg/controller/disruption/disruption.go
```diff
@@ -336,7 +338,8 @@ func (r *ReconcileMachineDisruption) updateMachineDisruptionBudgetStatus(
 		ExpectedMachines:          expectedCount,
 		MachineDisruptionsAllowed: disruptionsAllowed,
 		DisruptedMachines:         disruptedMachines,
-		ObservedGeneration:        mdb.Generation,
+		// We want that the ObservedGeneration and Generation fields will be the same after the MDB update
+		ObservedGeneration: mdb.Generation + 1,
```
enxebre: this does not seems related to the title of this commit

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -29,10 +32,24 @@ import (
 )
 
 const (
+	// MaxDisruptedMachinSize is the max size of MachineDisruptionBudgetStatus.DisruptedMachines.
+	// MachineHealthCheck will refuse to delete machine covered by the corresponding MDB
+	// if the size of the map exceeds this value.
+	MaxDisruptedMachinSize = 50
```
enxebre: why 50?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -211,6 +228,15 @@ func remediate(r *ReconcileMachineHealthCheck, machine *mapiv1.Machine) (reconci
 			return reconcile.Result{}, err
 		}
 
+		// do not fail immediatlty, but try again if the method fails because of the update conflict
+		err = r.retryCheckAndDecrement(machine)
+		if err != nil {
```
enxebre: this could be inlined

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +378,62 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
 	}
 	return false
 }
+
+// checkAndDecrement checks if the provided MachineDisruptionBudget allows any disruption
```
enxebre: the func is called `checkAndDecrement` but the comment mentions nothing about decrement. Also what is it checking..? Probably should split in two functions `isMachineDisruptionAllowed` or so which does just that.

pkg/util/machines/machines.go
```diff
@@ -19,7 +19,7 @@ import (
 // IsMachineHealthy returns true if the the machine is running and machine node is healthy
 func IsMachineHealthy(c client.Client, machine *mapiv1.Machine) bool {
 	if machine.Status.NodeRef == nil {
-		glog.Infof("machine %s does not have NodeRef", machine.Name)
+		glog.V(4).Infof("machine %s does not have NodeRef", machine.Name)
```
enxebre: this does not seems related to the title of this commit

pkg/util/machines/machines.go
```diff
@@ -33,12 +33,12 @@ func IsMachineHealthy(c client.Client, machine *mapiv1.Machine) bool {
 
 	readyCond := conditions.GetNodeCondition(node, v1.NodeReady)
 	if readyCond == nil {
-		glog.Infof("node %s does have 'Ready' condition", machine.Name)
+		glog.V(4).Infof("node %s does have 'Ready' condition", machine.Name)
```
enxebre: this does not seems related to the title of this commit

pkg/util/machines/machines.go
```diff
@@ -33,12 +33,12 @@ func IsMachineHealthy(c client.Client, machine *mapiv1.Machine) bool {
...
 		return false
 	}
 
 	if readyCond.Status != v1.ConditionTrue {
-		glog.Infof("node %s does have has 'Ready' condition with the status %s", machine.Name, readyCond.Status)
+		glog.V(4).Infof("node %s does have has 'Ready' condition with the status %s", machine.Name, readyCond.Status)
```
enxebre: this does not seems related to the title of this commit

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
cynepco3hahue: I assume if it has more than 50 machines under the `DisruptedMachines` list it says that the cluster has some problem, I  just assume that under the big cluster it can be around 500 nodes, so 50, in this case, looks fine

pkg/controller/disruption/disruption.go
cynepco3hahue: I will move all logs changes under the different commit, but it relatively small change, so I thought it is ok

pkg/controller/disruption/disruption.go
cynepco3hahue: @enxebre  I encounter the problem with it when I tested integration, can I just add a better description under the commit instead of creating a new one?

pkg/controller/disruption/disruption.go
cynepco3hahue: @enxebre I encounter the problem with it when I tested integration, can I just add a better description under the commit instead of creating a new one?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
cynepco3hahue: will update the comment

enxebre: we should split and rename

pkg/util/machines/machines.go
cynepco3hahue: done

pkg/util/machines/machines.go
cynepco3hahue: done

pkg/util/machines/machines.go
cynepco3hahue: done

cynepco3hahue:
@enxebre Fixed all comments, can you please review again?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +377,71 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
+
+func (r *ReconcileMachineHealthCheck) decrementMachineDisruptionsAllowed(machineName string, mdb *healthcheckingv1alpha1.MachineDisruptionBudget) error {
+	mdb.Status.MachineDisruptionsAllowed--
+	if mdb.Status.DisruptedMachines == nil {
+		mdb.Status.DisruptedMachines = make(map[string]metav1.Time)
+	}
+	// MachineHealthCheck controller needs to inform the MDB controller that it is about to delete a machine
+	// so it should not consider it as available in calculations when updating MachineDisruptions allowed.
+	// If the machine is not deleted within a reasonable time limit MDB controller will assume that it won't
+	// be deleted at all and remove it from DisruptedMachines map.
```
ingvagabund: It's up to MHC to decide when a machine is healthy and when it's not. Thus, MDB controller can not make that decision and remove a machine from `mdb.Status.DisruptedMachines`. Also, object's status reflects object's recociler's opinion about object's current state. Thus, updating object status with two different controllers is not recommended. I would rather have MDB to be the only controller allowed to update MDB object's status. But again, MHC is responsible for deciding which machine is unhealthy and which machine needs to be re-mediated. So it's only up to MHC to modify `mdb.Status.DisruptedMachines` field.

pkg/controller/disruption/disruption.go
ingvagabund: Question to this particular line. Does it mean that when there is at least the same number of currently healthy machines as desired machines, we will no allow mhc to fix the remaining unhealthy machines? `isMachineDisruptionAllowed` has the second condition pretty clear:

```go
if mdb.Status.MachineDisruptionsAllowed < 0 {
	glog.Warningf("the machine disruption budget %s MachineDisruptionsAllowed is negative", mdb.Name)
	return false
}
```

I still want the MHC to be able to re-mediate machines even when `currentHealthy > desiredHealthy`.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
cynepco3hahue: It does not the limit on disruption machines it is a limit on the number of disrupted machines under the `DisruptedMachines map[string]metav1.Time` map, I think the comment describes pretty well the purpose of this map
https://github.com/openshift/machine-api-operator/blob/8b8563a8e9fba4b13babbfd679fc9ab0ed73da0b/pkg/apis/healthchecking/v1alpha1/machinedisruptionbudget_types.go#L36

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -29,10 +32,24 @@ import (
...
 	ownerControllerKind  = "MachineSet"
 )
 
+// UpdateMDBRetry is the retry for a conflict where multiple clients
+// are making changes to the same resource.
+var UpdateMDBRetry = wait.Backoff{
+	Steps:    20,
+	Duration: 500 * time.Millisecond,
+	Factor:   1.0,
+	Jitter:   0.1,
```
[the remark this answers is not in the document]

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -211,6 +228,14 @@ func remediate(r *ReconcileMachineHealthCheck, machine *mapiv1.Machine) (reconci
 			return reconcile.Result{}, err
 		}
 
+		// do not fail immediatlty, but try again if the method fails because of the update conflict
+		if err = r.retryCheckAndDecrement(machine); err != nil {
+			// if the error appears here it means that machine healthcheck operation restricted by machine
+			// disruption budget, in this case we want to re-try after one minute
+			glog.Warning(err)
+			return reconcile.Result{Requeue: true, RequeueAfter: time.Minute}, nil
```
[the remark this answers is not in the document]
cynepco3hahue: I do not sure I understand, what do you mean by `continue processing other machines`? The `remediate` method works only on a single machine.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +377,71 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
+
+	return true
+}
+
+func (r *ReconcileMachineHealthCheck) decrementMachineDisruptionsAllowed(machineName string, mdb *healthcheckingv1alpha1.MachineDisruptionBudget) error {
+	mdb.Status.MachineDisruptionsAllowed--
+	if mdb.Status.DisruptedMachines == nil {
+		mdb.Status.DisruptedMachines = make(map[string]metav1.Time)
+	}
+	// MachineHealthCheck controller needs to inform the MDB controller that it is about to delete a machine
```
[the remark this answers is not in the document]
cynepco3hahue: will change the `delete` -> `remediate`

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +377,71 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
+func (r *ReconcileMachineHealthCheck) retryCheckAndDecrement(machine *mapiv1.Machine) error {
+	var mdb *healthcheckingv1alpha1.MachineDisruptionBudget
+	err := retry.RetryOnConflict(UpdateMDBRetry, func() error {
+		mdbs, err := machineutil.GetMachineMachineDisruptionBudgets(r.client, machine)
+		if err != nil {
+			return err
+		}
+
+		if len(mdbs) > 1 {
+			return fmt.Errorf("machine %s has more than one MachineDisruptionBudget, which not supported", machine.Name)
```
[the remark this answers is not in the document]

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
cynepco3hahue: I can see that `DisruptedMachines` field raises a lot of questions, to be honest, for me, it also hard to imagine the situation when:
```
	// DisruptedMachines contains information about machines whose deletion was
	// processed by the API server but has not yet been observed by the MachineDisruptionBudget controller.
	// A machine will be in this map from the time when the API server processed the
	// deletion request to the time when the machine is seen by MDB controller
	// as having been marked for deletion (or after a timeout).
```
but I made this controller similar to pod distribution one, so I believe k8s guys had some issues with it. But we can consider removing `DisruptedMachines` in the next iteration, but for now, I prefer to leave it close as possible to the pod disruption controller to avoid some corner cases that we still did not encounter.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +377,71 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
+				return fmt.Errorf("mahcine disruption is not allowed")
+			}
+
+			return r.decrementMachineDisruptionsAllowed(machine.Name, mdbs[0])
+		}
+		return nil
+	})
+
+	if err == wait.ErrWaitTimeout {
+		err = fmt.Errorf("couldn't update MachineDisruptionBudget %q due to conflicts", mdb.Name)
```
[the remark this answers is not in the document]
cynepco3hahue: updated previous code to use `mdb` variable instead of `mdbs[0]`

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -327,3 +327,97 @@ func TestUnhealthyForTooLong(t *testing.T) {
...
+
+	mdb2 := maotesting.NewMinAvailableMachineDisruptionBudget(1)
+	mdb1.Name = "mdb2"
+	mdb2.Status.MachineDisruptionsAllowed = 1
+
+	testsCases := []struct {
+		testName string
+		machine  *mapiv1alpha1.Machine
+		mdbs     []*healthcheckingv1alpha1.MachineDisruptionBudget
+		expected bool
```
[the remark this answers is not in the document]
cynepco3hahue: yep if do we expect an error, changed `expected -> error`

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +377,71 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
+				return fmt.Errorf("mahcine disruption is not allowed")
+			}
+
+			return r.decrementMachineDisruptionsAllowed(machine.Name, mdb)
+		}
+		return nil
+	})
+
+	if err == wait.ErrWaitTimeout {
+		err = fmt.Errorf("couldn't update MachineDisruptionBudget %q due to conflicts", mdb.Name)
```
vikaschoudhary16:
```suggestion
		return fmt.Errorf("Couldn't update MachineDisruptionBudget %q due to conflicts", mdb.Name)
```

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +377,71 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
+	}
+
+	return err
```
vikaschoudhary16:
```suggestion
	return nil
```

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +377,71 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
 	}
 	return false
 }
+
+// isMachineDisruptionAllowed returns true if the provided MachineDisruptionBudget allows any disruption
+func isMachineDisruptionAllowed(mdb *healthcheckingv1alpha1.MachineDisruptionBudget, maxDisruptedMachinSize int) bool {
```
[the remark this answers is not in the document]
cynepco3hahue: I think `isMachineDisruptionAllowed` better describes the purpose of this method, because it checks if remediation allowed only in the context of the machine disruption budget, it possible that in future we will have additional conditions when we want to prevent remediation and in this case, we can introduce `isMachineRemediationAllowed` and call to `isMachineDisruptionAllowed` as part of it.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +377,71 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
+	if mdb.Status.ObservedGeneration < mdb.Generation {
+		glog.Warningf("the machine disruption budget %s is still being processed by the server", mdb.Name)
```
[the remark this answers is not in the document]

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +377,71 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
+	if mdb.Status.MachineDisruptionsAllowed < 0 {
+		glog.Warningf("the machine disruption budget %s MachineDisruptionsAllowed is negative", mdb.Name)
+		return false
+	}
+	if len(mdb.Status.DisruptedMachines) > maxDisruptedMachinSize {
+		glog.Warningf("the machine disruption budget %s DisruptedMachines map too big - too many deletions not confirmed by MDB controller", mdb.Name)
+		return false
+	}
+	if mdb.Status.MachineDisruptionsAllowed == 0 {
+		glog.Warningf("cannot delete machine as it would violate the machine's disruption budget %s", mdb.Name)
```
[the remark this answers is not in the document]

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +377,71 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
+
+		if len(mdbs) > 1 {
+			return fmt.Errorf("machine %q has more than one MachineDisruptionBudget, which is not supported", machine.Name)
+		}
+
+		if len(mdbs) == 1 {
+			mdb = mdbs[0]
+
+			if !isMachineDisruptionAllowed(mdb, MaxDisruptedMachinSize) {
+				return fmt.Errorf("mahcine disruption is not allowed")
```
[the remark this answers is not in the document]

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
cynepco3hahue: we want to update an error message just in cases when, `err == wait.ErrWaitTimeout`, but it also possible that we have other errors

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
cynepco3hahue: see comment above

pkg/controller/disruption/disruption.go
bison: Can you explain this? This doesn't make sense to me. You'd typically use the observed generation to detect changes to the spec, but in that case you'd set it to whatever generation you acted on, not just increment it.

cynepco3hahue: Sure, `Generation` field updated each time on the object update, status can be updated by MHC or by MDB controllers, but in the case when it updated by MDB, I want to have `ObservedGeneration` equals to `Generation` otherwise on each update of status by MDB `ObservedGeneration` will be different from `Generation` and in this case check https://github.com/openshift/machine-api-operator/pull/273/files#diff-614d58186947ca2e4e215d42c496d72eR383 will always fail, but we want to fail it only when MDB CR updated by MHC controller and not by MDB.

ingvagabund: This is definitely wrong since the MDB CR is updated 8 times a second (it can be even higher depending on your CPU). Everytime you update the CR, it triggers another reconciling loop since `ObservedGeneration` is always different.

ingvagabund: Even with this change the controller reconciles the CR all over again since the CR is always different. It always gets updated and reconciled again.

cynepco3hahue: It is the good point will check the code again

cynepco3hahue: @ingvagabund ah but we have this check before
```
if mdb.Status.CurrentHealthy == currentHealthy &&
		mdb.Status.DesiredHealthy == desiredHealthy &&
		mdb.Status.ExpectedMachines == expectedCount &&
		mdb.Status.MachineDisruptionsAllowed == disruptionsAllowed &&
		reflect.DeepEqual(mdb.Status.DisruptedMachines, disruptedMachines) &&
		mdb.Status.ObservedGeneration == mdb.Generation {
		return nil
	}
```
so it should be ok

ingvagabund: It's not ok. I can record a video that shows you mdb is reconciling like crazy. Though, no matter if it's `ObservedGeneration: mdb.Generation` or `ObservedGeneration: mdb.Generation + 1`, it's still looping all over again. So, this actually does not break/fix anything since it's already broken.

cynepco3hahue: it fixes the problem with the check under https://github.com/openshift/machine-api-operator/pull/273/files#diff-614d58186947ca2e4e215d42c496d72eR383 I wasn't aware that it reconciles like crazy, can you give a link on the video I will take a look

ingvagabund: This fixes the case: https://github.com/openshift/machine-api-operator/pull/350

pkg/controller/disruption/disruption.go
enxebre: > we add one because we do not want to respect disruption budget when expected and healthy are equal

I don't understand this ^^ expected=desired, healthy=current? I'm confused

Also if  `disruptionsAllowed := currentHealthy - desiredHealthy`
if currentHealthy=3 and desiredHealthy=5 why wouldn't that allow disruptions??
This needs to clarified  and cleaned along with `getExpectedMachineCount` and `getExpectedScale which names are confusing
Also why `getExpectedScale` look at the Replicas at all, that's pretty odd since we are talking about machines here, how does replicas comes in to the game?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +377,71 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
+	}
+	// MachineHealthCheck controller needs to inform the MDB controller that it is about to remediate a machine
+	// so it should not consider it as available in calculations when updating MachineDisruptions allowed.
+	// If the machine is not remediated within a reasonable time limit MDB controller will assume that it won't
+	// be remediated at all and remove it from DisruptedMachines map.
+	mdb.Status.DisruptedMachines[machineName] = metav1.Time{Time: time.Now()}
+	return r.client.Update(context.TODO(), mdb)
+}
+
+// retryCheckAndDecrement tries to run checkAndDecrement command and if it fails because of the
```
enxebre: Can you please move this to the MDB controller
Also what is command? what is check? what is decrementing? Naming is confusing here

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +377,71 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
+	}
+	if mdb.Status.MachineDisruptionsAllowed == 0 {
+		glog.Warningf("Cannot remediate machine as it would violate the machine's disruption budget %s", mdb.Name)
+		return false
+	}
+
+	return true
+}
+
+func (r *ReconcileMachineHealthCheck) decrementMachineDisruptionsAllowed(machineName string, mdb *healthcheckingv1alpha1.MachineDisruptionBudget) error {
```
enxebre: can you please move this to the MDB controller

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +377,71 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
 	}
 	return false
 }
+
```
enxebre: can you please move this to the MDB controller

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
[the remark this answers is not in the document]
enxebre: can we please move this to the MDB controller

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
enxebre: can we please put this in the MDB controller instead of here

enxebre:
please see https://github.com/openshift/machine-api-operator/pull/273#discussion_r302895421

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -211,6 +228,14 @@ func remediate(r *ReconcileMachineHealthCheck, machine *mapiv1.Machine) (reconci
 			return reconcile.Result{}, err
 		}
 
+		// do not fail immediatlty, but try again if the method fails because of the update conflict
+		if err = r.retryCheckAndDecrement(machine); err != nil {
```
enxebre: I'm confused. It's difficult to say what this function is meant to do by the name but it seems to be meant to update and decrement the number of disruptions allowed, but the remediation is not decided to take place until line 240 so why trying to decrease here?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +377,71 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
+		return false
+	}
+	if len(mdb.Status.DisruptedMachines) > maxDisruptedMachinSize {
+		glog.Warningf("The machine disruption budget %s DisruptedMachines map too big - too many deletions not confirmed by MDB controller", mdb.Name)
+		return false
+	}
+	if mdb.Status.MachineDisruptionsAllowed == 0 {
+		glog.Warningf("Cannot remediate machine as it would violate the machine's disruption budget %s", mdb.Name)
+		return false
+	}
```
ingvagabund: I have a cluster with the following nodes:
```
NAME                                   STATUS     ROLES     AGE       VERSION
0bfe7aa3-000c-4f20-b75c-558effd3ecf0   Ready	  worker    3h45m     v1.13.7-beta.0-1
243b4181-1505-4ae2-999b-6b10224840a5   NotReady   worker    11m       v1.13.7-beta.0-1
595dce88-c9ca-484e-a44d-c5919f5b7868   NotReady   worker    11m       v1.13.7-beta.0-1
63141c47-5892-4ea2-ba09-c9c976831689   NotReady   worker    11m       v1.13.7-beta.0-1
80aedfb8-3c2a-43fe-9f65-9c9219a13e07   NotReady   worker    11m       v1.13.7-beta.0-1
8fa00196-9d60-4496-9373-df2743e10b89   NotReady   worker    11m       v1.13.7-beta.0-1
a17d1e3f-b632-43c4-99b3-3fc8c30f73e7   NotReady   worker    11m       v1.13.7-beta.0-1
e158da13-8a14-41bc-a4c0-171254b40ec0   NotReady   worker    11m       v1.13.7-beta.0-1
fcf4ebbd-6bb3-4716-8175-2af362d8dfe3   NotReady   worker    11m       v1.13.7-beta.0-1
fd6bdee7-1057-4134-884c-12d58746a733   NotReady   worker    11m       v1.13.7-beta.0-1
ip-10-0-101-95.ec2.internal            Ready	  master    4h10m     v1.13.1
```

MHC and MDB (with `maxUnavailable` set to 1) are deployed. MHC is continuously reporting:
```
W0712 15:02:26.733652   14737 machinehealthcheck_controller.go:385] cannot delete machine as it would violate the machine's disruption budget mdb-test
W0712 15:02:26.733675   14737 machinehealthcheck_controller.go:224] mahcine disruption is not allowed
```

Given the disruption budget is already violated even before MHC starts to remediate, none of the NotReady nodes will ever get healed. Is that the expected behavior? I would assume at least one node at a time will be allowed to be  healed.

ingvagabund: Ok, I get it now. I had different assumption about how the disruption budget is used. There are two perspectives actually:
1) Cluster is assumed to be healthy. There is upper bound to a number of node that can be unhealthy so in case the bound is set e.g. to 4 and at most 4 nodes switch to NotReady state, MHC controller can remediate those nodes.
2) In case I create the disruption budget for a cluster where too many nodes are unhealthy, I expect the MHC controller to do absolutely nothing about it.

So, this behavior is actually correct.

ingvagabund: My original though was another scenario that is currently not handled by the disruption budget. In case there are more nodes unhealthy than the disruption budget allows I was expecting the MHC controller to select a subset of at most `maxUnavailable` nodes to heal (so it would kinda respect the fact that only at most `maxUnavailable` machines are healed). The confusion hear is semantics of the word disruption. My first impression was it means at most `maxUnavailable` nodes can be healed. Not the fact, that if a set of nodes has more than `maxUnavailable` nodes unhealthy, entire set of the nodes is considered too fragile to be touched.

pkg/controller/disruption/disruption.go
ingvagabund: No need to `ObservedGeneration: mdb.Generation + 1` at all. The reason for the endless reconciliation loop is the call to `client.Update(...)`, which always bumps object's `Generation` field. The right way to update the status is to call `client.Status().Update(...)`, which will just updated the status without changing the object's `Generation` field.

ingvagabund:
The most noticeable changes I made:

```
diff --git a/install/0000_30_machine-api-operator_07_machinehealthcheck.crd.yaml b/install/0000_30_machine-api-operator_07_machinehealthcheck.crd.yaml
index 8ad7563..0d0662b 100644
--- a/install/0000_30_machine-api-operator_07_machinehealthcheck.crd.yaml
+++ b/install/0000_30_machine-api-operator_07_machinehealthcheck.crd.yaml
@@ -14,6 +14,8 @@ spec:
     - mhc
     - mhcs
   scope: Namespaced
+  subresources:
+    status: {}
   validation:
     openAPIV3Schema:
       properties:
diff --git a/install/0000_30_machine-api-operator_08_machinedisruptionbudget.crd.yaml b/install/0000_30_machine-api-operator_08_machinedisruptionbudget.crd.yaml
index 8f49b3c..641a23b 100644
--- a/install/0000_30_machine-api-operator_08_machinedisruptionbudget.crd.yaml
+++ b/install/0000_30_machine-api-operator_08_machinedisruptionbudget.crd.yaml
@@ -27,6 +27,8 @@ spec:
     - mdb
     - mdbs
   scope: Namespaced
+  subresources:
+    status: {}
   validation:
     openAPIV3Schema:
       properties:
@@ -67,7 +69,15 @@ spec:
         status:
           description: Most recently observed status of the MachineDisruptionBudget.
           properties:
-            DisruptedMachines:
+            currentHealthy:
+              description: current number of healthy machines
+              format: int32
+              type: integer
+            desiredHealthy:
+              description: minimum desired number of healthy machines
+              format: int32
+              type: integer
+            disruptedMachines:
               description: DisruptedMachines contains information about machines whose
                 deletion was processed by the API server but has not yet been observed
                 by the MachineDisruptionBudget controller. A machine will be in this
@@ -81,14 +91,6 @@ spec:
                 goes smooth this map should be empty for the most of the time. Large
                 number of entries in the map may indicate problems with machines deletions.
               type: object
-            currentHealthy:
-              description: current number of healthy machines
-              format: int32
-              type: integer
-            desiredHealthy:
-              description: minimum desired number of healthy machines
-              format: int32
-              type: integer
             disruptionsAllowed:
               description: Number of machines disruptions that are currently allowed.
               format: int32
diff --git a/pkg/apis/healthchecking/v1alpha1/machinedisruptionbudget_types.go b/pkg/apis/healthchecking/v1alpha1/machinedisruptionbudget_types.go
index 0960809..9c91a30 100644
--- a/pkg/apis/healthchecking/v1alpha1/machinedisruptionbudget_types.go
+++ b/pkg/apis/healthchecking/v1alpha1/machinedisruptionbudget_types.go
@@ -44,7 +44,7 @@ type MachineDisruptionBudgetStatus struct {
 	// If everything goes smooth this map should be empty for the most of the time.
 	// Large number of entries in the map may indicate problems with machines deletions.
 	// +optional
-	DisruptedMachines map[string]metav1.Time `json:"DisruptedMachines,omitempty" protobuf:"bytes,2,rep,name=DisruptedMachines"`
+	DisruptedMachines map[string]metav1.Time `json:"disruptedMachines,omitempty" protobuf:"bytes,2,rep,name=disruptedMachines"`
 
 	// Number of machines disruptions that are currently allowed.
 	MachineDisruptionsAllowed int32 `json:"disruptionsAllowed" protobuf:"varint,3,opt,name=disruptionsAllowed"`
@@ -63,7 +63,7 @@ type MachineDisruptionBudgetStatus struct {
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // MachineDisruptionBudget is an object to define the max disruption that a collection of machines can experience
-// kubebuilder:subresource:status
+// +kubebuilder:subresource:status
 // +kubebuilder:resource:shortName=mdb;mdbs
 // +kubebuilder:printcolumn:name="Healthy",type="integer",JSONPath=".status.currentHealthy",description="The number of healthy machines"
 // +kubebuilder:printcolumn:name="Total",type="integer",JSONPath=".status.expectedMachines",description="The total number of machines"
diff --git a/pkg/apis/healthchecking/v1alpha1/machinehealthcheck_types.go b/pkg/apis/healthchecking/v1alpha1/machinehealthcheck_types.go
index 5c2e2c8..fa8c927 100644
--- a/pkg/apis/healthchecking/v1alpha1/machinehealthcheck_types.go
+++ b/pkg/apis/healthchecking/v1alpha1/machinehealthcheck_types.go
@@ -7,11 +7,13 @@ import (
 // ConfigMapNodeUnhealthyConditions contains the name of the unhealthy conditions config map
 const ConfigMapNodeUnhealthyConditions = "node-unhealthy-conditions"
 
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // MachineHealthCheck is the Schema for the machinehealthchecks API
-// kubebuilder:subresource:status
+// +kubebuilder:subresource:status
 // +kubebuilder:resource:shortName=mhc;mhcs
 // +k8s:openapi-gen=true
 type MachineHealthCheck struct {
diff --git a/pkg/controller/disruption/disruption.go b/pkg/controller/disruption/disruption.go
index 807caf7..8a7fec9 100644
--- a/pkg/controller/disruption/disruption.go
+++ b/pkg/controller/disruption/disruption.go
@@ -3,14 +3,13 @@ package disruption
 import (
 	"context"
 	"fmt"
-	"reflect"
 	"time"
 
 	"github.com/golang/glog"
 	mapiv1 "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	healthcheckingv1alpha1 "github.com/openshift/machine-api-operator/pkg/apis/healthchecking/v1alpha1"
-	"github.com/openshift/machine-api-operator/pkg/util"
 	machineutil "github.com/openshift/machine-api-operator/pkg/util/machines"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -326,7 +313,7 @@ func (r *ReconcileMachineDisruption) updateMachineDisruptionBudgetStatus(
 		mdb.Status.DesiredHealthy == desiredHealthy &&
 		mdb.Status.ExpectedMachines == expectedCount &&
 		mdb.Status.MachineDisruptionsAllowed == disruptionsAllowed &&
-		reflect.DeepEqual(mdb.Status.DisruptedMachines, disruptedMachines) &&
+		apiequality.Semantic.DeepEqual(mdb.Status.DisruptedMachines, disruptedMachines) &&
 		mdb.Status.ObservedGeneration == mdb.Generation {
 		return nil
 	}
@@ -338,11 +325,10 @@ func (r *ReconcileMachineDisruption) updateMachineDisruptionBudgetStatus(
 		ExpectedMachines:          expectedCount,
 		MachineDisruptionsAllowed: disruptionsAllowed,
 		DisruptedMachines:         disruptedMachines,
-		// We want that the ObservedGeneration and Generation fields will be the same after the MDB update
-		ObservedGeneration: mdb.Generation + 1,
+		ObservedGeneration:        mdb.Generation,
 	}
 
-	return r.client.Update(context.TODO(), newMdb)
+	return r.client.Status().Update(context.TODO(), newMdb)
 }
 
 // failSafe is an attempt to at least update the MachineDisruptionsAllowed field to
diff --git a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
index cb67a13..c660009 100644
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -103,7 +93,7 @@ type ReconcileMachineHealthCheck struct {
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileMachineHealthCheck) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	glog.Infof("Reconciling MachineHealthCheck triggered by %s/%s\n", request.Namespace, request.Name)
+	glog.Infof("Reconciling MachineHealthCheck triggered by %q\n", request.Name)
 
 	// Get node from request
 	node := &corev1.Node{}
@@ -410,7 +424,7 @@ func (r *ReconcileMachineHealthCheck) decrementMachineDisruptionsAllowed(machine
 	// If the machine is not remediated within a reasonable time limit MDB controller will assume that it won't
 	// be remediated at all and remove it from DisruptedMachines map.
 	mdb.Status.DisruptedMachines[machineName] = metav1.Time{Time: time.Now()}
-	return r.client.Update(context.TODO(), mdb)
+	return r.client.Status().Update(context.TODO(), mdb)
 }
 
 // retryCheckAndDecrement tries to run checkAndDecrement command and if it fails because of the
```

- call `client.Status().Update` instead of `client.Update` to update status
- update CRDs to query `status` subresource
- dropping status generation increment

Still testing the integration though so far the integration of mhc controller with the disruption budget seems to work as expected. I.e. with `maxUnavailable = 1` only one machine/node at a time is deleted and replaced with a new one. Not more.

pkg/controller/disruption/disruption.go
cynepco3hahue: done

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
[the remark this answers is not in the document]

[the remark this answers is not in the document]

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go

pkg/controller/disruption/disruption.go
cynepco3hahue: Good catch and thanks for the fix.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
cynepco3hahue: Updated the name and the comment and moved under MDB controller

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
cynepco3hahue: The good point moved under `if unhealthyForTooLong(nodeCondition, conditionTimeout)`, so it will decrease the `MachineDisruptionsAllowed` only just before the remediation of the machine and also updated the name of the method.

pkg/controller/disruption/disruption.go
cynepco3hahue: > Also if `disruptionsAllowed := currentHealthy - desiredHealthy`
> if currentHealthy=3 and desiredHealthy=5 why wouldn't that allow disruptions?? naming is very confusing.

When `currentHealthy` < `desiredHealthy` it means that you do not have enough quorum to remediate machines, as a good example you can see CEPH quorum, you do not want to delete unhealthy machines until you will have a quorum again, because you do not really know if some connectivity problem or maybe data corrupted on machines, so we want to prevent remediation in this case until we will have quorum again.

> This needs to clarified and cleaned along with `getExpectedMachineCount` and `getExpectedScale` which names are confusing

I can agree that `expectedCount` does not suit well, maybe something like `totalMachinesNumber` will suit much better, but regarding all other names, I think it explains well the purpose of the method.

> Also why `getExpectedScale` look at the Replicas at all, that's pretty odd since we are talking about machines here, how does replicas comes in to the game?

To calculate the total number of machines and number of desired healthy machines we need to get `replicas` from the owner machine controller(if it has some), because of it possible that we have `machineset` with replica 2, but it has only one machine object because `machineset`  failed to create the second machine from some reason, but we want to take this machine into account when we calculate the `desiredHealthy` parameter.
```
if mdb.Spec.MaxUnavailable != nil {
		expectedCount = r.getExpectedScale(mdb, machines)
		desiredHealthy = expectedCount - int32(*mdb.Spec.MaxUnavailable)
		if desiredHealthy < 0 {
			desiredHealthy = 0
		}
	} else if mdb.Spec.MinAvailable != nil {
		desiredHealthy = *mdb.Spec.MinAvailable
		expectedCount = int32(len(machines))
	}
```

> Also this seems to be relying on `IsMachineHealthy` which use static conditions which might or might not be the same than the ones defined as the node-problem

Yep it should be fixed, will do it under the separate PR.

cynepco3hahue:
@enxebre I created two additional PR
https://github.com/openshift/machine-api-operator/pull/351 - that holds all renaming changes
https://github.com/openshift/machine-api-operator/pull/352 - use of config map to decide if a machine is unhealthy

pkg/controller/disruption/disruption.go
ingvagabund: > When currentHealthy < desiredHealthy it means that you do not have enough quorum to remediate machines, as a good example you can see CEPH quorum, you do not want to delete unhealthy machines until you will have a quorum again, because you do not really know if some connectivity problem or maybe data corrupted on machines, so we want to prevent remediation in this case until we will have quorum again.

That line is inside `updateMachineDisruptionBudgetStatus` method. So, I still don't see why the decision about respecting or not respecting the disruption budget should be made inside this method. If you want to have the MDB signal the MHC can not delete any additional machine since the disruption budget would be violated otherwise, the only way to do it is to set `mdb.Status.MachineDisruptionsAllowed` to zero.

ingvagabund: > Also if disruptionsAllowed := currentHealthy - desiredHealthy
if currentHealthy=3 and desiredHealthy=5 why wouldn't that allow disruptions?? naming is very confusing. This needs to clarified and cleaned along with getExpectedMachineCount and getExpectedScale which names are confusing

@enxebre's point is valid. If you want to make sure you will not start healing machines when the quorum is not reached, you set `MaxUnavailable` to be 1 or 2. That's not something that can be handled on the code level.

ingvagabund: If any, computation of `disruptionsAllowed` value does not belong inside `updateMachineDisruptionBudgetStatus`. `reconcile` itself might be the good place for it. Possibly inside another local method that will compute the value and provide justification why it's done this way.

ingvagabund: Just to be clear, I am commenting on the `+1` case. If the disruption budget is violated (currentlyHealthy < desiredHealthy), no disruption is allowed as expected.

ingvagabund: > `// we add one because we do not want to respect disruption budget when expected and healthy are equal`

If you allow to disrupt another machine when `currentHealthy == desiredHealthy`, you will violate the disruption budget. That's not right. Can you undo the change and open another PR we can discuss this particular case?

pkg/controller/disruption/disruption.go
```diff
@@ -26,16 +28,31 @@ import (
...
+	// DeletionTimeout sets maximum time from the moment a machine is added to DisruptedMachines in MDB.Status
+	// to the time when the machine is expected to be seen by MDB controller as having been marked for deletion.
+	// If the machine was not marked for deletion during that time it is assumed that it won't be deleted at
+	// all and the corresponding entry can be removed from mdb.Status.DisruptedMachines. It is assumed that
+	// machine/mdb apiserver to controller latency is relatively small (like 1-2sec) so the below value should
+	// be more than enough.
+	// If the controller is running on a different node it is important that the two nodes have synced
+	// clock (via ntp for example). Otherwise MachineDisruptionBudget controller may not provide enough
+	// protection against unwanted machine disruptions.
+	DeletionTimeout = 2 * time.Minute
```
enxebre: nit: does this need to be exported?

ingvagabund:
Wrt. `disruptionsAllowed := currentHealthy - desiredHealthy` equation. It makes perfect sense in context of pods since there is no pod healer (besides kubelet restarting failed containers). So the rule the PDB is guided by is: `I want at least so many pods to be healthy (whatever that means) so I will not allow to delete a pod in case the number of healthy pods might decreased (no matter which pod is to be deleted).` The `no matter which pod is to be deleted` is crucial here. The PDB does not care if the delete operation is trying to remove a pod that is not healthy. It just assumes the worst. Thus in case where I have 10 pods expected, 5 of them are healthy and `MaxUnavailable` is 5, the PDB will not allow to delete any pod. Even unhealthy one since it just does not care.

Assume the same scenario for machines (10 machines expected, 5 of them are healthy, `MaxUnavailable` is 5). Do we expect the same behavior from MDB? Or, do we want the MHC to remediate those 5 unhealthy machines? Well, my expectation was we want. Since it will not violate the disruption budget since there's still going to be those 5 healthy machines. If that's to be true, MDB needs to know which machines are considered healthy so it can properly calculate how many machines can be disrupted. And, from that point, `disruptionsAllowed` no longer makes sense since it's calculated over all machines. I would rather expect the MDB to tell me "Hey, `disruptionsAllowed` is 5 but you can delete only unhealthy machines". Which assumes tight coupling between MHC and MDB.

ingvagabund:
Apart from
- deciding when a disruption budget is violated (where we would really need a solid document explaining why's)
- having MDB to have the same ability to decide when a machine is (un)healthy
it's good to go:
- MDB no longer loops endlessly
- MDB only updates status, not entire object (which lead to generation number bump)
- MHC is not allowed to delete a machine in case the disruption budget will be violated

MDB is understood as MDB controller, MHC as MHC controller (keeping shorter forms for better readibility)

/approve

ingvagabund:
I also used the following steps to validate functionality of the integration:
- deploy cluster with one real node in AWS
- deploy a bunch of kubemark nodes (through machineset + kubemark actuator) on the real node in AWS
- use the following CRs:
```yaml
---
apiVersion: healthchecking.openshift.io/v1alpha1
kind: MachineDisruptionBudget
metadata:
  name: mdb-test
  namespace: kubemark-provider
spec:
  selector:
    matchLabels:
      machine.openshift.io/cluster-api-cluster: kubemark
      sigs.k8s.io/cluster-api-machineset: kubemark-scale-group-2a
  maxUnavailable: 2
---
apiVersion: healthchecking.openshift.io/v1alpha1
kind: MachineHealthCheck
metadata:
  name: example
  namespace: default
spec:
  remediationStrategy: ""
  selector:
    matchLabels:
      machine.openshift.io/cluster-api-cluster: kubemark
      sigs.k8s.io/cluster-api-machineset: kubemark-scale-group-2a
```

`.spec.selector.machineLabels` can be arbitrary as long as it targets machines owned by a machineset (or a set of machinesets).

I was able to observe successful healing while still respecting the disruption budget following the same politics as PDB does. I.e. disruption budget getting violated when there is a chance a healthy machine will get deleted no matter if the deleted machine is actually unhealthy thus keeping the same number of `MaxUnavailable` healthy machines constant.

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -11,6 +11,7 @@ import (
 	healthcheckingv1alpha1 "github.com/openshift/machine-api-operator/pkg/apis/healthchecking/v1alpha1"
 	"github.com/openshift/machine-api-operator/pkg/util/conditions"
 	maotesting "github.com/openshift/machine-api-operator/pkg/util/testing"
+
```
enxebre: why this extraspace?

enxebre:
lgtm PTAL @bison

ingvagabund:
> I think there's an issue with multiple calls to RetryDecrementMachineDisruptionsAllowed() decreasing the allowed disruption count even if the same machine is passed. We should fix that and test for it, but I'm fine with doing that in a follow-up -- especially given that this isn't really in-use at the moment.

Follow-up PR: https://github.com/openshift/machine-api-operator/pull/358

PR #277, opened 2019-04-10
WIP: mao: Add more logging

michaelgugino:
This commit adds additional logging to make e2e diagnosis
easier.

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/golang/glog"
 	appsv1 "k8s.io/api/apps/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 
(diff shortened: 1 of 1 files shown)


pkg/operator/sync.go
```diff
@@ -55,9 +55,11 @@ func (optr *Operator) syncClusterAPIController(config OperatorConfig) error {
 	controller := resourceread.ReadDeploymentV1OrDie(controllerBytes)
 	_, updated, err := resourceapply.ApplyDeployment(optr.kubeClient.AppsV1(), controller)
 	if err != nil {
+		glog.Error("Failed to apply deployment")
```
frobware: You might as well log the underlying error here too.
However, do we know why the return is not logged further up the call chain?

michaelgugino: The underlying error is logged after this function returns.  Unfortunately, the class of error can be 'timeout' and that's not specific to any particular thing this function does, so there's no way to see what 'actually' failed in this function.

This is basically just a bread crumb, I don't actually care what the error was just that I couldn't apply the deployment.

enxebre:
closing due to inactivity. Please reopen if still relevant.

PR #279, opened 2019-04-11
Feature gate support

cynepco3hahue:
Machine-api-operator controller will watch `FeatureGate` resources and update the deployment accordingly.
Currently, the only feature that can be enabled via the `FeatureGate` object is machine-health-check. To enable it you will need to create FeatureGate object with the content:
```yaml
kind: FeatureGate
metadata:
  name: machine-api-operator
spec:
  featureSet: "TechPreviewNoUpgrade"
```

TODO:
- [ ] add unittests
- [ ] adjust `MachineHealthCheck` e2e tests to create feature gate before the test and wait for controllers
- [ ] codify creation of the MHC CRD, can be done under another PR that will follow this one


ingvagabund:
@cynepco3hahue thanks for the PR.
Can you break it into two? First one generating the deployment in the code and removing the yaml manifest. And second introducing the feature gate? Those two changes can be done separately and are easier to review.

pkg/operator/sync.go
```diff
@@ -90,3 +86,126 @@ func (optr *Operator) waitForDeploymentRollout(resource *appsv1.Deployment) erro
 		return false, nil
 	})
```
enxebre: this is nice, thanks! can we get it to a different PR?

ingvagabund:
Depends on https://github.com/openshift/machine-api-operator/pull/280

cynepco3hahue:
Not relevant anymore.

PR #282, opened 2019-04-12
Machine-Healthcheck: Set Requeue if error in reading object

vikaschoudhary16:


pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -100,7 +100,8 @@ func (r *ReconcileMachineHealthCheck) Reconcile(request reconcile.Request) (reco
 			return reconcile.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
-		return reconcile.Result{}, err
+		glog.Warningf("Requeuing the key. Error reading the object: %v", err)
```
enxebre: why is this needed? by default any requests will requeue when returning error != nil

ingvagabund:
@vikaschoudhary16 can you update the description explaining why it is needed and what happens if not fixed? Any upstream issue or bug that you are fixing?

vikaschoudhary16:
Agree @enxebre @ingvagabund. I was mistaken. Closing.

PR #284, opened 2019-04-15
mhc: re-queue after the unhealthy node with the remaining duration

cynepco3hahue:
Before we re-queue the unhealthy node with the condition timeout duration, that can delay
the node remediation, to improve it now we re-queue the unhealthy node with the duration
equals to `conditionTimeout - nodeDurationUnhealthy + second`.

### DIFF
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -232,9 +232,13 @@ func remediate(r *ReconcileMachineHealthCheck, machine *mapiv1.Machine) (reconci
 			durationUnhealthy.String(),
 		)
 
+		// calculate the duration until the node will be unhealthy for too long
+		// and re-queue after with this timeout, add one second just to be sure
+		// that we will not enter this loop again before the node unhealthy for too long
+		unhealthyTooLongTimeout := conditionTimeout - durationUnhealthy + time.Second
 		// be sure that we will use timeout with the minimal value for the reconcile.Result
-		if minimalConditionTimeout == 0 || minimalConditionTimeout > conditionTimeout {
-			minimalConditionTimeout = conditionTimeout
+		if minimalConditionTimeout == 0 || minimalConditionTimeout > unhealthyTooLongTimeout {
+			minimalConditionTimeout = unhealthyTooLongTimeout
 		}
 		result = &reconcile.Result{Requeue: true, RequeueAfter: minimalConditionTimeout}
 	}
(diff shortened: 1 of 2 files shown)


pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -232,9 +232,10 @@ func remediate(r *ReconcileMachineHealthCheck, machine *mapiv1.Machine) (reconci
 			durationUnhealthy.String(),
 		)
 
+		unhealthyTimeout := conditionTimeout - durationUnhealthy + time.Second
```
ingvagabund: Can you more elaborate on why this is the right formula this time?

cynepco3hahue:
@enxebre @ingvagabund Can you please review again?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -232,9 +232,13 @@ func remediate(r *ReconcileMachineHealthCheck, machine *mapiv1.Machine) (reconci
...
+		// and re-queue after with this timeout, add one second just to be sure
+		// that we will not enter this loop again before the node unhealthy for too long
+		unhealthyTooLongTimeout := conditionTimeout - durationUnhealthy + time.Second
 		// be sure that we will use timeout with the minimal value for the reconcile.Result
-		if minimalConditionTimeout == 0 || minimalConditionTimeout > conditionTimeout {
-			minimalConditionTimeout = conditionTimeout
+		if minimalConditionTimeout == 0 || minimalConditionTimeout > unhealthyTooLongTimeout {
+			minimalConditionTimeout = unhealthyTooLongTimeout
 		}
 		result = &reconcile.Result{Requeue: true, RequeueAfter: minimalConditionTimeout}
```
ingvagabund: I see. So the previous lines of code were setting `RequeueAfter` to fixed `conditionTimeout`. So instead of checking how much is left before a machine is considered too unhealthy and setting `RequeueAfter` to the time remaining, `RequeueAfter` was always set to 5 minutes (the default timeout). Which is actually in accordance to logs [1]:
```
W0415 11:22:17.099490       1 machinehealthcheck_controller.go:226] Machine ci-op-q918js07-c6aca-qqmql-worker-us-east-1a-slrxb has unhealthy node ip-10-0-133-87.ec2.internal with the condition Ready and the timeout 300s for 2m57.099482011s. Requeuing...
I0415 11:27:17.100341       1 machinehealthcheck_controller.go:216] machine ci-op-q918js07-c6aca-qqmql-worker-us-east-1a-slrxb has been unhealthy for too long, deleting
```

So the new formula does the same. Just, instead of taking the fixed `conditionTimeout`, it subtracts the time for which the machine was in unhealthy state (`durationUnhealthy`).

[1] https://github.com/openshift/machine-api-operator/pull/285#issuecomment-483230161

ingvagabund:
/approve
/lgtm

Fixes flakiness in HC tests.

PR #287, opened 2019-04-16
Use deployment client

vikaschoudhary16:
For listing deployments, ideally deployment lister based on shared informers should have been used because of security aspects. But in the past there was an issue happening and as a stop gap kubeclient was used directly. 
Now that issue is not happening with shared informers and deployLister is being used.

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -8,7 +8,6 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


pkg/operator/sync.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/golang/glog"
 	appsv1 "k8s.io/api/apps/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	//metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
enxebre: can we removed comment?

pkg/operator/sync.go
```diff
@@ -65,17 +65,14 @@ func (optr *Operator) syncClusterAPIController(config OperatorConfig) error {
...
-		d, err := optr.kubeClient.AppsV1().Deployments(resource.Namespace).Get(resource.Name, metav1.GetOptions{})
+		d, err := optr.deployLister.Deployments(resource.Namespace).Get(resource.Name)
 		if apierrors.IsNotFound(err) {
 			return false, nil
 		}
 		if err != nil {
 			// Do not return error here, as we could be updating the API Server itself, in which case we
 			// want to continue waiting.
-			glog.Errorf("Error getting Deployment %q during rollout: %v", resource.Name, err)
+			glog.Errorf("Error vikas getting Deployment %q during rollout: %v", resource.Name, err)
```
enxebre: Can we don refer to Vikas specifically? :joy:

ingvagabund:
```
// TODO(vikas): When using deployLister, an issue is happening related to the apiVersion of cluster-api objects.
// This will be debugged later on to find out the root cause. For now, working aound is to use kubeClient.AppsV1
// d, err := optr.deployLister.Deployments(resource.Namespace).Get(resource.Name)
```
@vikaschoudhary16 can you update the PR description explaining why we can go back to `optr.deployLister`?

vikaschoudhary16:
@ingvagabund done!

vendor/github.com/openshift/cluster-api/pkg/apis/machine/v1beta1/machine_types.go
```diff
@@ -39,12 +39,6 @@ const (
 // Machine is the Schema for the machines API
 // +k8s:openapi-gen=true
 // +kubebuilder:subresource:status
-// +kubebuilder:printcolumn:name="Instance",type="string",JSONPath=".status.providerStatus.instanceId",description="Instance ID of machine created in AWS"
```
ingvagabund: Please, do not remove these lines. There are needed so we can generate CRDs with https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_02_machine.crd.yaml#L9-L33

cynepco3hahue:
> Please, put all `+kubebuilder:printcolumn` lines back.

Why we do not add it under the cluster-api repository?

pkg/operator/sync.go
frobware: Fame and fortune... :1st_place_medal:

PR #291, opened 2019-04-17
Allow to enable/disable machine healtch-check controller

ingvagabund:
So we can use feature gate mechanism to enable/disable the controller.

One needs to deploy:
```yaml
kind: FeatureGate
metadata:
  name: machine-api
spec:
  featureSet: "TechPreviewNoUpgrade"
```
to enable the HC deployment. Tested through unit tests where a fake clientset with the feature gate is provided.

Feature gate building pieces took from https://github.com/openshift/machine-api-operator/pull/279


ingvagabund:
@cynepco3hahue PTAL, I have integrated your feature gate PR, modified it a bit, added unit tests.

PR #292, opened 2019-04-17
Change operator failing to degraded

michaelgugino:
This commit renames statusFailing to statusDegraded
to conform with other openshift components.


michaelgugino:
/hold

Depends on: https://github.com/openshift/cluster-autoscaler-operator/pull/93

PR #293, opened 2019-04-22
fix feature gate

enxebre:
Follow up for https://github.com/openshift/machine-api-operator/pull/291 which left machine health check enable by default https://github.com/openshift/machine-api-operator/pull/291/files#diff-554de5523753fda8c93e7008c9bd947fR229

### DIFF
--- a/config/0000_05_config-operator_01_featuregate.crd.yaml
+++ b/config/0000_05_config-operator_01_featuregate.crd.yaml
@@ -1,0 +1,49 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: featuregates.config.openshift.io
+spec:
+  group: config.openshift.io
+  version: v1
+  scope: Cluster
+  names:
+    kind: FeatureGate
+    singular: featuregate
+    plural: featuregates
+    listKind: FeatureGateList
+  versions:
+  - name: v1
+    served: true
+    storage: true
+  validation:
+    openAPIV3Schema:
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          description: Standard object's metadata.
+          type: object
+        spec:
+          description: spec holds user settable values for configuration
+          properties:
+            featureSet:
+              description: featureSet changes the list of features in the cluster.  The
+                default is empty.  Be very careful adjusting this setting. Turning
+                on or off features may cause irreversible changes in your cluster
+                which cannot be undone.
+              type: string
+          type: object
+        status:
+          description: status holds observed values from the cluster. They may not
+            be overridden.
+          type: object
+      required:
+      - spec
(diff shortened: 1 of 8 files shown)


pkg/operator/operator_test.go
```diff
@@ -176,21 +176,34 @@ func deploymentHasContainer(d *appsv1.Deployment, containerName string) bool {
 	return false
 }
 
-func TestOperatorSyncClusterAPIControllerHealthCheckControllerNotDeployed(t *testing.T) {
```
ingvagabund: I put the tests under two separete functions so it's obvious which case is tested. Though, if you feel it's better this way, I am fine with that.

install/0000_30_machine-api-operator_08_rbac.yaml
```diff
@@ -50,6 +50,16 @@ rules:
     verbs:
       - get
 
+  - apiGroups:
```
ingvagabund: Please, extend https://github.com/openshift/machine-api-operator/blob/master/config/machine-api-operator-patch.yaml as well.

ingvagabund: Addressed in https://github.com/openshift/machine-api-operator/pull/295

ingvagabund:
Blocked by https://github.com/openshift/machine-api-operator/pull/295

config/0000_05_config-operator_01_featuregate.crd.yaml
```diff
@@ -0,0 +1,49 @@
...
+  - name: v1
+    served: true
+    storage: true
+  validation:
+    openAPIV3Schema:
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
```
vikaschoudhary16: s/value, and/value and

pkg/operator/sync.go
```diff
@@ -58,7 +58,7 @@ func (optr *Operator) syncClusterAPIController(config OperatorConfig) error {
 		if !errors.IsNotFound(err) {
 			return err
 		}
-		glog.V(2).Infof("failed to find feature gate %s, will use default feature set", MachineAPIFeatureGateName)
+		glog.V(2).Infof("Failed to find feature gate %q, will use default feature set", MachineAPIFeatureGateName)
```
vikaschoudhary16: should it be more verbose. may be v(4)?

pkg/operator/operator_test.go
vikaschoudhary16: table driven tests is the recommended approach i guess, https://github.com/golang/go/wiki/TableDrivenTests

config/0000_05_config-operator_01_featuregate.crd.yaml
ingvagabund: I copy-pasted the CRD, cant' fix that.

pkg/operator/sync.go
ingvagabund: V(4) is actually less verbose than V(2)

pkg/operator/operator_test.go
frobware: I find that it depends. Quite often table driven tests end up adding complexity and sometimes I find it easier to do:

```Go
func TestFooDoesX

func TestFooErrorsWithBar

func TestFooSucceedsWithKnownInputs
```

pkg/operator/operator_test.go
```diff
@@ -206,59 +219,24 @@ func TestOperatorSyncClusterAPIControllerHealthCheckControllerNotDeployed(t *tes
...
-			Provider:           "controllers-provider",
-			NodeLink:           "controllers-nodelink",
-			MachineHealthCheck: "controllers-machinehealthcheck",
-		},
-	}
+	for _, tc := range tests {
+		kubeclientSet := fakekube.NewSimpleClientset()
+		op, err := initOperator(tc.featureGate, kubeclientSet)
+		if err != nil {
+			t.Errorf("Unable to init operator: %v", err)
```
frobware: If this errors, what is the value of `op` a few lines below? Presumably it can be `nil` - so should this be `Fatalf`?

PR #294, opened 2019-04-22
Simplify Gopkg.toml and update vendor

michaelgugino:
This commit removes competing dependencies from Gopkg.toml
to allow us to utilize cluster-api-actuator-pkg as the
source of truth for these dependencies.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -45,16 +45,8 @@ required = [
   source = "github.com/openshift/kubernetes-kube-aggregator"
 
 [[override]]
-  name = "github.com/openshift/cluster-api"
-  branch = "openshift-4.0-cluster-api-0.0.0-alpha.4"
-
-[[override]]
   name = "github.com/openshift/cluster-api-actuator-pkg"
   branch = "master"
-
-[[override]]
-  name = "github.com/openshift/api"
-  branch = "rebase-1.13.4"
 
 [[override]]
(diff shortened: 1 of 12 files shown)


Gopkg.toml
```diff
@@ -44,18 +44,10 @@ required = [
...
-[[override]]
-  name = "github.com/openshift/cluster-api"
-  branch = "openshift-4.0-cluster-api-0.0.0-alpha.4"
-
 [[override]]
   name = "github.com/openshift/cluster-api-actuator-pkg"
   branch = "master"
 
-[[override]]
-  name = "github.com/openshift/api"
```
ingvagabund: I don't see `github.com/openshift/api` contained in any of actuator-pkg, cluster-api repos. Are you sure we can remove this override as well?

michaelgugino: It's a constraint in github.com/openshift/cluster-api-actuator-pkg .

ingvagabund: I see now [1]. We than need to change the constraint in `actuator-pkg` to be:
```
[[constraint]]
  name = "github.com/openshift/api"
  branch = "rebase-1.13.4"
```
so we don't accidentally sync with older or newer version.

[1] https://github.com/openshift/cluster-api-actuator-pkg/blob/master/Gopkg.toml#L44-L46

PR #295, opened 2019-04-23
kubemark: Deploy feature gate CRDs

ingvagabund:
So the kubemark-based deployment of machine API stack can test new features in technical preview.


ingvagabund:
Integrated into https://github.com/openshift/machine-api-operator/pull/293

PR #299, opened 2019-04-25
Unit test: Verify hc feature-gate disable after enable scenario

vikaschoudhary16:
This unit test verifies that if hc feature-gate is disabled after enabling it, hc-controller gets removed.

### DIFF
--- a/pkg/operator/operator_test.go
+++ b/pkg/operator/operator_test.go
@@ -249,3 +249,61 @@ func TestOperatorSyncClusterAPIControllerHealthCheckController(t *testing.T) {
 		}
 	}
 }
+
+func TestEnableDisableOperatorSyncClusterAPIControllerHealthCheckController(t *testing.T) {
+	infra := &v1.Infrastructure{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "cluster",
+		},
+		Status: v1.InfrastructureStatus{
+			Platform: v1.AWSPlatformType,
+		},
+	}
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+
+	optr := newFakeOperator(nil, []runtime.Object{newFeatureGate(v1.TechPreviewNoUpgrade), infra}, stopCh)
+	go optr.Run(2, stopCh)
+
+	if err := wait.PollImmediate(1*time.Second, 5*time.Second, func() (bool, error) {
+		d, err := optr.deployLister.Deployments(targetNamespace).Get(deploymentName)
+		if err != nil {
+			t.Logf("Failed to get %q deployment: %v", deploymentName, err)
+			return false, nil
+		}
+		if !deploymentHasContainer(d, hcControllerName) {
+			t.Logf("Failed to verify that deployment %q has container %q", deploymentName, hcControllerName)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		t.Fatalf("Failed to verify that deployment %q has container %q", deploymentName, hcControllerName)
+	}
+
+	t.Logf("Found %q container in %q deployment", hcControllerName, deploymentName)
+
+	hcFeatureGate, err := optr.osClient.ConfigV1().FeatureGates().Get(MachineAPIFeatureGateName, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("Unable to get %q feature gate: %v", MachineAPIFeatureGateName, err)
+	}
+	hcFeatureGate.Spec.FeatureSet = v1.Default
+	if _, err := optr.osClient.ConfigV1().FeatureGates().Update(hcFeatureGate); err != nil {
+		t.Fatalf("Unable to update %q feature gate and set default feature set: %v", MachineAPIFeatureGateName, err)
+	}
+
+	if err := wait.PollImmediate(1*time.Second, 15*time.Second, func() (bool, error) {
+		d, err := optr.deployLister.Deployments(targetNamespace).Get(deploymentName)
+		if err != nil {
+			t.Logf("Failed to get %q deployment: %v", deploymentName, err)
+			return false, nil
+		}
+		if deploymentHasContainer(d, hcControllerName) {
+			t.Logf("Failed to verify that deployment %q has no container with name %q", deploymentName, hcControllerName)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		t.Errorf("Failed to verify that deployment %q has no container with name %q ", deploymentName, hcControllerName)
+	}
+
+}
--- a/vendor/github.com/openshift/cluster-version-operator/lib/resourcemerge/core.go
+++ b/vendor/github.com/openshift/cluster-version-operator/lib/resourcemerge/core.go
@@ -24,38 +24,8 @@ func ensurePodTemplateSpec(modified *bool, existing *corev1.PodTemplateSpec, req
 }
 
 func ensurePodSpec(modified *bool, existing *corev1.PodSpec, required corev1.PodSpec) {
-	// any container we specify, we require.
-	for _, required := range required.InitContainers {
-		var existingCurr *corev1.Container
-		for j, curr := range existing.InitContainers {
-			if curr.Name == required.Name {
-				existingCurr = &existing.InitContainers[j]
-				break
-			}
-		}
-		if existingCurr == nil {
-			*modified = true
-			existing.Containers = append(existing.InitContainers, corev1.Container{})
-			existingCurr = &existing.InitContainers[len(existing.InitContainers)-1]
-		}
-		ensureContainer(modified, existingCurr, required)
-	}
-
-	for _, required := range required.Containers {
-		var existingCurr *corev1.Container
-		for j, curr := range existing.Containers {
-			if curr.Name == required.Name {
-				existingCurr = &existing.Containers[j]
-				break
-			}
-		}
-		if existingCurr == nil {
-			*modified = true
-			existing.Containers = append(existing.Containers, corev1.Container{})
-			existingCurr = &existing.Containers[len(existing.Containers)-1]
-		}
-		ensureContainer(modified, existingCurr, required)
-	}
+	ensureContainers(modified, &existing.InitContainers, required.InitContainers)
+	ensureContainers(modified, &existing.Containers, required.Containers)
 
 	// any volume we specify, we require.
 	for _, required := range required.Volumes {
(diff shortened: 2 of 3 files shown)


pkg/operator/operator_test.go
```diff
@@ -208,6 +208,9 @@ func TestOperatorSyncClusterAPIControllerHealthCheckController(t *testing.T) {
 			},
 		},
 		expectedMachineHealthCheckController: true,
+	}, {
```
ingvagabund: How does this verify the feature is disable after it's enabled? The operator is initialized from scratch for every test case.

enxebre: indeed, this is duplicating the tc https://github.com/openshift/machine-api-operator/pull/299/files#diff-d5d9cf6e8104e5231787b5cc89fb7b52R198

pkg/operator/operator.go
```diff
@@ -196,37 +206,45 @@ func (optr *Operator) sync(key string) error {
...
+		TargetNamespace: optr.namespace,
+		Controllers: Controllers{
+			Provider:           providerControllerImage,
+			NodeLink:           machineAPIOperatorImage,
+			MachineHealthCheck: machineAPIOperatorImage,
+		},
+	}, nil
+}
+
+func (optr *Operator) getImages() (string, string, error) {
```
ingvagabund: No need to implement and mock `getImages()`. You can create `Configv1.Infrastructure` object and have the fake client return it.

vendor/github.com/openshift/cluster-version-operator/lib/resourcemerge/core.go
```diff
@@ -56,6 +56,15 @@ func ensurePodSpec(modified *bool, existing *corev1.PodSpec, required corev1.Pod
 		}
 		ensureContainer(modified, existingCurr, required)
 	}
+	var requiredContainers []corev1.Container
```
ingvagabund: How is this change related to the unit tests?

pkg/operator/operator.go
```diff
@@ -47,7 +49,11 @@ type Operator struct {
 	osClient      osclientset.Interface
 	eventRecorder record.EventRecorder
 
-	syncHandler func(ic string) error
+	syncHandler           func(ic string) error
+	imageGetterFunc       imageGetterFuncType
+	statusProgressingFunc func() error
```
ingvagabund: No need to mock `optr.statusProgressing` given you can have fake clients return expected objects. Please, give fake client `osconfigv1.ClusterOperator` object with pre-populated testing values.

vendor/github.com/openshift/cluster-version-operator/lib/resourcemerge/core.go
vikaschoudhary16: https://github.com/openshift/cluster-version-operator/pull/178

pkg/operator/operator.go
ingvagabund: Implemented here: https://github.com/openshift/machine-api-operator/pull/304

enxebre:
@vikaschoudhary16 needs rebase

vikaschoudhary16:
This PR still needs https://github.com/openshift/cluster-version-operator/pull/216/files

pkg/operator/operator_test.go
```diff
@@ -217,22 +229,27 @@ func TestOperatorSyncClusterAPIControllerHealthCheckController(t *testing.T) {
...
 
 	for _, tc := range tests {
-		infra := &v1.Infrastructure{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: "cluster",
-			},
-			Status: v1.InfrastructureStatus{
-				Platform: v1.AWSPlatformType,
-			},
+		err := initFeatureGateInformer([]runtime.Object{tc.featureGate}, optr)
```
ingvagabund: This is still wrong. All tests in the list (`tests`) are supposed to assume clean/vanilla environment. Initializing fg informer only once leaves previous fg object in the cache. Which might (or might not) influence result of each test in the last. If you want to assume dependency between fg objects, you need to create separate unit test for it.

pkg/operator/operator_test.go
```diff
@@ -228,11 +240,59 @@ func TestOperatorSyncClusterAPIControllerHealthCheckController(t *testing.T) {
...
+		},
+		Status: v1.InfrastructureStatus{
+			Platform: v1.AWSPlatformType,
+		},
+	}
+
+	stopCh := make(<-chan struct{})
+	optr := newFakeOperator(nil, []runtime.Object{infra}, stopCh)
+
+	for _, tc := range tests {
```
ingvagabund: Given the unit test will always have two steps (first enable fg, then disable fg), there is no need to iterate through both tests. It's sufficient to initialize `newFakeOperator` as it was before, remove `initFeatureGateInformer` and just call `optr.sync` two times. After the first call validating the fg is enable, after the second call validating it is not.

pkg/operator/operator_test.go
```diff
@@ -228,11 +240,59 @@ func TestOperatorSyncClusterAPIControllerHealthCheckController(t *testing.T) {
...
 			},
 		}
-
 		stopCh := make(<-chan struct{})
-		optr := newFakeOperator(nil, []runtime.Object{tc.featureGate, infra}, stopCh)
-		go optr.Run(2, stopCh)
+		optr := newFakeOperator(nil, []runtime.Object{infra}, stopCh)
+		err := initFeatureGateInformer([]runtime.Object{tc.featureGate}, optr)
+		if err != nil {
+			t.Errorf("Failed to set featuregate %v : %v", tc.featureGate, err)
```
frobware: Isn't this fatal?

pkg/operator/operator_test.go
vikaschoudhary16: >  remove initFeatureGateInformer

This is not possible because modifying the `FeatureGate` object which is passed to the fake clientset, will not work.  Eventually it makes a copy of the object and therefor any changes to the object will not be reflected back in the object returned by informer in response to the listing operation.

Updated the PR to address other parts. PTAL!

pkg/operator/operator_test.go
vikaschoudhary16: with `Errorf` this test will be reported failed but others will be executed. Is not that we want?

pkg/operator/operator_test.go
ingvagabund:
```

Comment (ingvagabund) [reply pkg/operator/operator_test.go]: @vikaschoudhary16 also please improve the messages to be more human readable

File: pkg/operator/operator_test.go [hunk 8]
```diff
@@ -74,7 +75,8 @@ func newFakeOperator(kubeObjects []runtime.Object, osObjects []runtime.Object, s
 	}
 
 	configSharedInformer.Start(stopCh)
-	kubeNamespacedSharedInformer.Start(stopCh)
+	kubeInformerstopCh := make(<-chan struct{})
```
frobware: Why do we need this separate channel? Shouldn't the following line use `stopCh`?

pkg/operator/operator_test.go
```diff
@@ -52,7 +53,7 @@ func newOperatorConfig() *OperatorConfig {
 	}
 }
 
-func newFakeOperator(kubeObjects []runtime.Object, osObjects []runtime.Object, stopCh <-chan struct{}) *Operator {
+func newFakeOperator(kubeObjects []runtime.Object, osObjects []runtime.Object, stopCh chan struct{}) *Operator {
```
ingvagabund: No need for `s/<-chan/chan/`.

pkg/operator/operator_test.go
vikaschoudhary16: updated, thanks!

pkg/operator/operator_test.go
ingvagabund: Agree, please undo the change.

pkg/operator/operator_test.go
```diff
@@ -157,7 +158,7 @@ func TestOperatorSync_NoOp(t *testing.T) {
 				},
 			}
 
-			stopCh := make(<-chan struct{})
+			stopCh := make(chan struct{})
```
ingvagabund: Please undo

pkg/operator/operator_test.go
```diff
@@ -229,7 +230,7 @@ func TestOperatorSyncClusterAPIControllerHealthCheckController(t *testing.T) {
 			},
 		}
 
-		stopCh := make(<-chan struct{})
+		stopCh := make(chan struct{})
```
ingvagabund: Please undo

pkg/operator/operator_test.go
```diff
@@ -249,3 +250,61 @@ func TestOperatorSyncClusterAPIControllerHealthCheckController(t *testing.T) {
...
+	go optr.Run(2, stopCh)
+
+	if err := wait.PollImmediate(1*time.Second, 5*time.Second, func() (bool, error) {
+		d, err := optr.deployLister.Deployments(targetNamespace).Get(deploymentName)
+		if err != nil {
+			t.Logf("Failed to get %q deployment: %v", deploymentName, err)
+			return false, nil
+		}
+		if !deploymentHasContainer(d, hcControllerName) {
+			t.Logf("Expected deploymentHasContainer for %q container to be true", hcControllerName)
```
ingvagabund: `Expected to found no %q container in %q deployment, got one.`

pkg/operator/operator_test.go
```diff
@@ -249,3 +250,61 @@ func TestOperatorSyncClusterAPIControllerHealthCheckController(t *testing.T) {
...
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		t.Fatalf("Failed to verify %q deployment", deploymentName)
```
ingvagabund: "Failed to verify %q deployment has %q container"

pkg/operator/operator_test.go
```diff
@@ -249,3 +250,61 @@ func TestOperatorSyncClusterAPIControllerHealthCheckController(t *testing.T) {
...
+	}
+
+	if err := wait.PollImmediate(1*time.Second, 15*time.Second, func() (bool, error) {
+		d, err := optr.deployLister.Deployments(targetNamespace).Get(deploymentName)
+		if err != nil {
+			t.Logf("Failed to get %q deployment: %v", deploymentName, err)
+			return false, nil
+		}
+		if deploymentHasContainer(d, hcControllerName) {
+			t.Logf("Expected deploymentHasContainer for %q container to be false", hcControllerName)
```
ingvagabund: `Expected to found %q container in %q deployment, found none.`

pkg/operator/operator_test.go
```diff
@@ -249,3 +250,61 @@ func TestOperatorSyncClusterAPIControllerHealthCheckController(t *testing.T) {
...
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		t.Errorf("Failed to verify %q deployment", deploymentName)
```
ingvagabund: "Failed to verify %q deployment has no %q container"

pkg/operator/operator_test.go
```diff
@@ -52,14 +52,17 @@ func newOperatorConfig() *OperatorConfig {
 	}
 }
 
-func newFakeOperator(kubeObjects []runtime.Object, osObjects []runtime.Object, stopCh <-chan struct{}) *Operator {
+func newFakeOperator(kubeObjects []runtime.Object, osObjects []runtime.Object, stopCh <-chan struct{}, wq string) *Operator {
```
frobware: what is `wq`?

ingvagabund: If this is the way to fix the flake, please open another PR for it. This is irrelevant for this PR.

vikaschoudhary16: will it be ok if i create a new commit for this flake fix. Actually with the changes in this PR, this flake has become almost 100% reproducible and thus will be difficult to merge this PR without this fix.

pkg/operator/operator_test.go
vikaschoudhary16: why should not the stop channel be closed?

ingvagabund: undo = `s/<-//`. Channel only for reading is all that is needed.

pkg/operator/operator_test.go
ingvagabund: How can different name for a working queue be a fix for the flake?

frobware: How does the name, or a different name, fix the flake?

ingvagabund: Still, it's unrelated to this PR. Please, open another one.

PR #303, opened 2019-04-29
Set operator controller recorder via variable

cynepco3hahue:
Makes possible to unittest operator events by passing fake recorder to the operator.


enxebre:
thanks! looks good to me. Let's hold until we open to merging after code freeze next week
/approve
/hold

PR #298, opened 2019-04-25
Run make test-e2e from vendor cluster-api-actuator-pkg

michaelgugino:
This commit enables running make script directly from
cluster-api-actuator-pkg to reduce drift.

PR #307, opened 2019-04-29
[WIP] Make from vendor

frobware:
This is a hack/variation on https://github.com/openshift/machine-api-operator/pull/298


frobware:
Fixed in https://github.com/openshift/machine-api-operator/pull/308

PR #310, opened 2019-05-01
Add toleration

ravisantoshgudimetla:
As of now, we are tolerating only `node-role.kubernetes.io/master` taint during scheduling stage. If we don't have the blanket toleration, there is a very good chance that these pods won't tolerate other taints that got added to the node, for example `disk-pressure` etc. The side-effect is that we will tolerate `NoExecute` taints as well

Please feel free to close this PR, if you think, you just need to tolerate the current taint(s) you have.

/cc @sjenning @derekwaynecarr

### DIFF
--- a/install/0000_30_machine-api-operator_09_deployment.yaml
+++ b/install/0000_30_machine-api-operator_09_deployment.yaml
@@ -46,9 +46,7 @@ spec:
         runAsNonRoot: true
         runAsUser: 65534
       tolerations:
-      - key: "node-role.kubernetes.io/master"
-        operator: "Exists"
-        effect: "NoSchedule"
+      - operator: "Exists"
       volumes:
       - name: images
         configMap:


install/0000_30_machine-api-operator_09_deployment.yaml
```diff
@@ -46,9 +46,7 @@ spec:
         runAsNonRoot: true
         runAsUser: 65534
       tolerations:
-      - key: "node-role.kubernetes.io/master"
         operator: "Exists"
-        effect: "NoSchedule"
```
ingvagabund: CI complains about:
```
error: error validating "STDIN": error validating data: ValidationError(Deployment.spec.template.spec.tolerations): invalid type for io.k8s.api.core.v1.PodSpec.tolerations: got "map", expected "array"; if you choose to ignore these errors, turn validation off with --validate=false
```
It should be at least:
```
tolerations:
- operator: "Exists"
```

ravisantoshgudimetla: Ohh yes, thanks for catching it. Fixed it :)

ingvagabund:
@ravisantoshgudimetla is there a list of currently available node taints we might face? I would still rather make an explicit list of tolerations rather than tolerating all taints. Though, at the end we just want the mao to be running so we can reconcile the machine api stack any time it's change or we need to update the deployment.

> The side-effect is that we will tolerate NoExecute taints as well

That might be a trouble when a master node is temporarily put aside yet still part of the cluster. E.g. to allow pods running on the master to be eventually scheduled on another master node so the affected master node can be analyzed.

ravisantoshgudimetla:
closing in favour of #311

PR #314, opened 2019-05-14
nodelink-controller: Support multiple internal IPs.

russellb:
In a baremetal environment, we may have servers with multiple network
interfaces.  In that case, the Machine may have multiple internal IPs
listed for the different network interfaces.

This patch makes the nodelink-controller keep track of all Machine
internal IPs so that it can successfully match a Node's internal IP
against whichever IP is being used by the kubelet.

### DIFF
--- a/cmd/nodelink-controller/controller_test.go
+++ b/cmd/nodelink-controller/controller_test.go
@@ -20,7 +20,7 @@ import (
 	"reflect"
 	"testing"
 
-	mapiv1alpha1 "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
+	mapiv1 "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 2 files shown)


enxebre:
can we have a unit test for this?

enxebre:
@russellb would you mind putting a unit for this?

russellb:
@enxebre sure, sorry I missed the first time you asked that

russellb:
@enxebre added tests.

While writing the tests, I noticed that the behavior of `updateMachine()` looks wrong.  It doesn't look at the old copy of the `Machine`, so if the addresses change, the cache is going to have stale entries in it mapping IPs to an incorrect `Machine`.  I can send a follow-up PR for that, since it's not related to this change.

russellb:
I don't see that unit test failure locally, and it looks unrealted to these changes ...

/test unit

enxebre:
@russellb sorry I overlapped with this. Multiple IPs will be now included via https://github.com/openshift/machine-api-operator/pull/321 so closing this avoid rebasing hassle

PR #321, opened 2019-06-05
Nodelink

enxebre:
Today the logic that handles the tangential points between machine lifecycle and node lifecycle
which infers information between machine/node is split between mao node link controller and cluster-api node controller. This has been historical this way since it was initially the proposed flow upstream for consumers opting-in to get the nodeRef via annotation. This is unnecessary and overcomplicated. This is a first PR to simplify and unify all the same domain behaviour in one single controller, aka node link controller and removing the unnecessary split.
- There's a follow up to drop cluster-api node controller and consolidate that behaviour here.

This PR replicate our current node link controller and:
- Fix the re-sync period (default to 10h), previously set to 5 seconds. Controller is only driven on events.
- Move logic to controller runtime consistently with all the machine api controllers
- Find machines/nodes by providerID
- Fallback to find machines/nodes by internalIP
- Add at least 3 cases for test coverage for every existing function


pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+
+	//Watch for changes to Node
+	err = c.Watch(&source.Kind{Type: &corev1.Node{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to Machines and enqueue if it exists the backed node
+	err = c.Watch(&source.Kind{Type: &mapiv1beta1.Machine{}}, &handler.EnqueueRequestsFromMapFunc{ToRequests: mapFn})
```
ingvagabund: I don't think this is actually needed. Any node object is periodically updated (by kubelet updating node status). So, even if a machine is changed (e.g. labels added, taints added), referenced node will get eventually updated. By watching machines you are increasing the rate in which node requests are processed.

I was more interesting if there is a way to put two different kinds into the same queue and process them in either the same `Reconcile` method or in two distinct.

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,394 @@
...
+
+	for i := range machineList.Items {
+		for _, a := range machineList.Items[i].Status.Addresses {
+			// Use the internal IP to look for matches:
+			if a.Type == corev1.NodeInternalIP {
+				if a.Address == nodeInternalAddress {
+					klog.V(3).Infof("Found machine %q for node %q with internal IP %q", machineList.Items[i].GetName(), node.GetName(), a.Address)
+					return &machineList.Items[i], nil
+				}
+				break
```
russellb: Can you drop this `break` here?  See https://github.com/openshift/machine-api-operator/pull/314 which would be dropped if this PR goes in first

enxebre: sure, I'll make sure to cover #314 if this merges first

pkg/controller/nodelink/nodelink_controller.go
enxebre: Given the node heartbeat frequency this won't make a substantial difference, there might be cases where heartbeat fails and this pursues correctness where reconciliation wants to be triggered by a machine change, e.g taints/labels.
This is pretty much doing what you state in the last paragraph https://github.com/kubernetes-sigs/controller-runtime/blob/master/FAQ.md#q-how-do-i-know-which-type-of-object-a-controller-references

pkg/controller/nodelink/nodelink_controller.go
enxebre: done https://github.com/openshift/machine-api-operator/pull/321/files#diff-69885f1e6e1f6b4cb889ce80021c95a1R82

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,394 @@
...
+		if nodeList.Items[i].Spec.ProviderID == *machine.Spec.ProviderID {
+			klog.V(3).Infof("Found node %q for machine %q with providerID %q", nodeList.Items[i].GetName(), machine.GetName(), nodeList.Items[i].Spec.ProviderID)
+			r.providerIDToNode[*machine.Spec.ProviderID] = &nodeList.Items[i]
+			return r.providerIDToNode[*machine.Spec.ProviderID], nil
+		}
+	}
+	return nil, nil
+}
+
+func (r *ReconcileNodeLink) findNodeFromMachineByIP(machine mapiv1beta1.Machine) (*corev1.Node, error) {
```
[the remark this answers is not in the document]
enxebre: done

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,430 @@
...
+		},
+	); err != nil {
+		return nil, fmt.Errorf("error setting index fields: %v", err)
+	}
+
+	if err := mgr.GetCache().IndexField(&mapiv1beta1.Machine{},
+		"ProviderID",
+		func(object runtime.Object) []string {
+			if machine, ok := object.(*mapiv1beta1.Machine); ok {
+				if machine.Spec.ProviderID != nil {
```
frobware: I think we should also check that `*machine.Spec.ProviderID != ""`. That case is invalid.

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,430 @@
...
+		},
+	); err != nil {
+		return nil, fmt.Errorf("error setting index fields: %v", err)
+	}
+
+	if err := mgr.GetCache().IndexField(&corev1.Node{},
+		"InternalIP",
+		func(object runtime.Object) []string {
+			var keys []string
+			if node, ok := object.(*corev1.Node); ok {
```
frobware: I think from a debugging standpoint it may be useful to log when this doesn't hold true. something somewhere went terribly wrong.

vikaschoudhary16:
`Fallback to find machines/nodes by internalDNSName` --- IIUC, it fallbacks to internalIP not dns name.

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,430 @@
...
+			},
+		}
+	}
+
+	klog.V(3).Infof("No-op: Node for machine %q not found", machine.GetName())
+	return []reconcile.Request{}
+}
+
+// findNodeFromMachine tries to a node from by providerIP and fallback to find by IP
+func (r *ReconcileNodeLink) findNodeFromMachine(machine mapiv1beta1.Machine) (*corev1.Node, error) {
```
frobware: Any reason not to pass a pointer here? `machine` is not mutated.

frobware: And the same comment goes to the other find() methods.

pkg/controller/nodelink/nodelink_controller_test.go
```diff
@@ -0,0 +1,495 @@
...
+}
+
+const (
+	ownerControllerKind = "MachineSet"
+	namespace           = "openshift-machine-api"
+)
+
+var (
+	knownDate          = metav1.Time{Time: time.Date(1985, 06, 03, 0, 0, 0, 0, time.Local)}
+	fakeNodeIndexer    = make(map[string]corev1.Node)
```
frobware: It would be better if these were not globals. Create them in the tests that need them. Modify build() funcs to return a map. Otherwise it is not clear from Test to Test function whether they should/should not have any content.

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+	return add(mgr, reconciler, reconciler.nodeRequestFromMachine)
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager) (*ReconcileNodeLink, error) {
+	// set convenient indexers
+	if err := mgr.GetCache().IndexField(&corev1.Node{},
+		"ProviderID",
+		func(object runtime.Object) []string {
```
ingvagabund: Please name the anonymous functions:
- so they can be unit tested
- to get:

```go
const (
	machineInternalIPIndex = "machineInternalIPIndex"
	machineProviderIDIndex = "machineProviderIDIndex"
	nodeInternalIPIndex = "nodeInternalIPIndex"
	nodeProviderIDIndex = "nodeProviderIDIndex"
)
...
if err := mgr.GetCache().IndexField(&corev1.Node{}, nodeProviderIDIndex, indexNodeByProviderID); err != nil {
	return nil, fmt.Errorf("error setting index fields: %v", err)
}
if err := mgr.GetCache().IndexField(&mapiv1beta1.Machine{}, machineProviderIDIndex, indexMachineByProviderID); err != nil {
	return nil, fmt.Errorf("error setting index fields: %v", err)
}
if err := mgr.GetCache().IndexField(&corev1.Node{}, nodeInternalIPIndex, indexNodeByInternalIP); err != nil {
	return nil, fmt.Errorf("error setting index fields: %v", err)
}
if err := mgr.GetCache().IndexField(&mapiv1beta1.Machine{}, machineInternalIPIndex, indexMachineByInternalIP); err != nil {
	return nil, fmt.Errorf("error setting index fields: %v", err)
}

```

pkg/controller/nodelink/nodelink_controller_test.go
```diff
@@ -0,0 +1,495 @@
...
+		machine.Spec.Taints = taints
+	}
+	return machine
+}
+
+func newFakeReconciler(client client.Client, machine *mapiv1beta1.Machine, node *corev1.Node) ReconcileNodeLink {
+	r := ReconcileNodeLink{
+		client: client,
+	}
+	r.listNodesByFieldFunc = func(_, value string) ([]corev1.Node, error) {
```
ingvagabund: Did you try to use the fake client itself? It should work as any other client where you just populated all node/machine objects on your own.

pkg/controller/nodelink/nodelink_controller.go
enxebre: why would you pass a pointer? that'd mean the machine is mutable within the function, why?

frobware: Asking because it is not mutated and would prevent the shallow copy.

frobware: The shallow copy doesn't prevent other pointer values being mutated: 

```go
package main

import (
	"fmt"
)

type A struct {
	b *B
}

type B struct {
	name string
}

func bar(a A) {
	a.b.name = "oops"
}

func main() {
	a := A{&B{name: "hello world"}}
	fmt.Println(a.b.name)

	bar(a)

	if a.b.name != "hello world" {
		panic(a.b.name)
	}
}
```

pkg/controller/nodelink/nodelink_controller_test.go
enxebre: yes, the fakeClient does not support fake indexers

pkg/controller/nodelink/nodelink_controller.go
enxebre: Done and created unit test for each indexer function

pkg/controller/nodelink/nodelink_controller_test.go
enxebre: Done in newFakeReconciler

pkg/controller/nodelink/nodelink_controller.go
enxebre: done

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,488 @@
...
+				},
+			},
+		}
+	}
+
+	klog.V(3).Infof("No-op: Node for machine %q not found", machine.GetName())
+	return []reconcile.Request{}
+}
+
+// findNodeFromMachine tries to a node from by providerIP and fallback to find by IP
```
vikaschoudhary16: s/tries to find
s/providerID

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,488 @@
...
+
+	node, err = r.findNodeFromMachineByIP(machine)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find node from machine %q by internal IP: %v", machine.GetName(), err)
+	}
+	return node, nil
+}
+
+func (r *ReconcileNodeLink) findNodeFromMachineByProviderID(machine *mapiv1beta1.Machine) (*corev1.Node, error) {
+	if machine == nil {
```
vikaschoudhary16: Should this be an error case?

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,488 @@
...
+	}
+
+	nodes, err := r.listNodesByFieldFunc(nodeProviderIDIndex, *machine.Spec.ProviderID)
+	if err != nil {
+		return nil, fmt.Errorf("failed getting node list: %v", err)
+	}
+
+	if len(nodes) > 0 {
+		klog.V(3).Infof("Found node %q for machine %q with providerID %q", nodes[0].GetName(), machine.GetName(), nodes[0].Spec.ProviderID)
+		return nodes[0].DeepCopy(), nil
```
vikaschoudhary16: should we return the most recently created node?

pkg/controller/nodelink/nodelink_controller.go
frobware: I think we should avoid these checks. It masks other problems. We shouldn't be trying to find a node  on a machine that we don't have (i.e., machine is `nil`).

vikaschoudhary16: If this func is called, it should be called with a non-nil machine. a nil machine is unexpected and a sign of some problem,so should be raised as an error. If there was no machine, it should have been handled outside of this func and shuold not have called this.

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,488 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed getting node list: %v", err)
+	}
+
+	if len(nodes) > 0 {
+		klog.V(3).Infof("Found node %q for machine %q with internal IP %q", nodes[0].GetName(), machine.GetName(), machineInternalAddress)
+		return nodes[0].DeepCopy(), nil
+	}
+
+	klog.V(3).Infof("not found node match for machine %q with internal IP %q", machine.GetName(), machineInternalAddress)
```
vikaschoudhary16: s/Matching node not found

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,488 @@
...
+	klog.Warningf("Expected a machine for indexing field, got: %T", object)
+	return []string{}
+}
+
+func indexNodeByInternalIP(object runtime.Object) []string {
+	var keys = []string{}
+	node, ok := object.(*corev1.Node)
+	if !ok {
+		klog.Warningf("expected a node for indexing field, got: %T", object)
+		return keys
```
frobware: We should be explicit with the return type and only return non-nil when we actually have some keys: 

```go
func indexNodeByInternalIP(object runtime.Object) []string {
	node, ok := object.(*corev1.Node)
	if !ok {
		klog.Warningf("expected a node for indexing field, got: %T", object)
		return nil
	}

	var keys = []string

	for _, a := range node.Status.Addresses {
		if a.Type == corev1.NodeInternalIP {
			keys = append(keys, a.Address)
			klog.V(3).Infof("Adding internal IP %q for node %q to indexer", a.Address, node.GetName())
		}
	}

	return keys
}
```

frobware: And the same comment https://github.com/openshift/machine-api-operator/pull/321/commits/d687214415f30eb550cc75daa2b5325b677b47e4#r291497921 applies to the other indexing functions.

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,488 @@
...
+	machine, err = r.findMachineFromNodeByIP(node)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find node from machine %q by internal IP: %v", machine.GetName(), err)
+	}
+	return machine, nil
+}
+
+func (r *ReconcileNodeLink) findMachineFromNodeByProviderID(node *corev1.Node) (*mapiv1beta1.Machine, error) {
+	if node == nil {
+		klog.V(3).Infof("Finding machine from node nil, returning...")
```
vikaschoudhary16: same as above

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,488 @@
...
+		return nil, nil
+	}
+
+	machines, err := r.listMachinesByFieldFunc(machineProviderIDIndex, node.Spec.ProviderID)
+	if err != nil {
+		return nil, fmt.Errorf("failed getting node list: %v", err)
+	}
+	if len(machines) > 0 {
+		klog.V(3).Infof("Found machine %q for node %q with providerID %q", machines[0].GetName(), node.GetName(), node.Spec.ProviderID)
+		return machines[0].DeepCopy(), nil
```
vikaschoudhary16: should we error if more than one machine?

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,488 @@
...
+	}
+	return nil, nil
+}
+
+func (r *ReconcileNodeLink) findMachineFromNodeByIP(node *corev1.Node) (*mapiv1beta1.Machine, error) {
+	if node == nil {
+		klog.V(3).Infof("Finding machine from node nil, returning...")
```
vikaschoudhary16: should we error?

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,488 @@
...
+	for _, a := range node.Status.Addresses {
+		if a.Type == corev1.NodeInternalIP {
+			nodeInternalAddress = a.Address
+			klog.V(3).Infof("Found internal IP for node %q: %q", node.GetName(), nodeInternalAddress)
+			break
+		}
+	}
+
+	if nodeInternalAddress == "" {
+		klog.Warningf("not found internal IP for node %q", node.GetName())
```
vikaschoudhary16: s/not/Not

vikaschoudhary16: Should we error here?

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,488 @@
...
+	}
+
+	machines, err := r.listMachinesByFieldFunc(machineInternalIPIndex, nodeInternalAddress)
+	if err != nil {
+		return nil, fmt.Errorf("failed getting node list: %v", err)
+	}
+
+	if len(machines) > 0 {
+		klog.V(3).Infof("Found machine %q for node %q with internal IP %q", machines[0].GetName(), node.GetName(), nodeInternalAddress)
+		return machines[0].DeepCopy(), nil
```
vikaschoudhary16: Should we error if more than one machine?

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,488 @@
...
+	}
+	klog.V(3).Infof("not found node match for node %q with internal IP %q", node.GetName(), nodeInternalAddress)
```
vikaschoudhary16: s/Matching node not found

pkg/controller/nodelink/nodelink_controller.go
frobware: nit: "Node %q has no internal IP address"

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,488 @@
...
+	klog.V(3).Infof("Finding machine from node %q", node.GetName())
+	machine, err := r.findMachineFromNodeByProviderID(node)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find machine from node %q by ProviderID: %v", node.GetName(), err)
+	}
+	if machine != nil {
+		return machine, nil
+	}
+
+	machine, err = r.findMachineFromNodeByIP(node)
```
vikaschoudhary16: Should we set node.Spec.ProviderID from machine.Spec.ProviderID?

pkg/controller/nodelink/nodelink_controller_test.go
```diff
@@ -0,0 +1,649 @@
...
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"testing"
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime"
+
```
frobware: nit: drop blank line.

pkg/controller/nodelink/nodelink_controller.go
enxebre: Let's error if len>1, that's an anomaly and error should eventually reconcile towards success

pkg/controller/nodelink/nodelink_controller_test.go
```diff
@@ -0,0 +1,649 @@
...
+				},
+			}, nil),
+			node: node("matchingInternalIP", "test", []corev1.NodeAddress{
+				{
+					Type:    corev1.NodeInternalIP,
+					Address: "matchingInternalIP",
+				},
+			}, nil),
+			expected: machine("matchingInternalIP", "test", []corev1.NodeAddress{
+				{
```
vikaschoudhary16: test case for non-matching internalIP?

pkg/controller/nodelink/nodelink_controller_test.go
```diff
@@ -0,0 +1,649 @@
...
+					Address: "matchingInternalIP",
+				},
+			}, nil),
+			node: node("matchingInternalIP", "test", []corev1.NodeAddress{
+				{
+					Type:    corev1.NodeInternalIP,
+					Address: "matchingInternalIP",
+				},
+			}, nil),
+			expected: node("matchingInternalIP", "test", []corev1.NodeAddress{
```
vikaschoudhary16: non-matching InternalIP?

pkg/controller/nodelink/nodelink_controller.go
frobware: > Should we set node.Spec.ProviderID from machine.Spec.ProviderID?

Not here - this function is about find.

pkg/controller/nodelink/nodelink_controller_test.go
```diff
@@ -0,0 +1,649 @@
...
+				Address: "matchingInternalIP",
+			}}, nil),
+			node: node("matchInternalIP", "", []corev1.NodeAddress{{
+				Type:    corev1.NodeInternalIP,
+				Address: "matchingInternalIP",
+			}}, nil),
+			expected: []reconcile.Request{
+				{
+					NamespacedName: client.ObjectKey{
+						Namespace: metav1.NamespaceNone,
```
vikaschoudhary16: test case with both, providerID and IP, matching?
test case with both, providerID and IP, non-empty and neither matching?

pkg/controller/nodelink/nodelink_controller.go
frobware: > Should we error here?

Isn't the state transient here? Next time it may have addresses.

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,500 @@
...
+		}
+	}
+
+	klog.V(3).Infof("No-op: Node for machine %q not found", machine.GetName())
+	return []reconcile.Request{}
+}
+
+// findNodeFromMachine find a node from by providerID and fallback to find by IP
+func (r *ReconcileNodeLink) findNodeFromMachine(machine *mapiv1beta1.Machine) (*corev1.Node, error) {
+	if machine == nil {
```
frobware: I still think we should drop this; calling this with nil is broken elsewhere.

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,500 @@
...
+	if len(nodes) == 1 {
+		klog.V(3).Infof("Found node %q for machine %q with providerID %q", nodes[0].GetName(), machine.GetName(), nodes[0].Spec.ProviderID)
+		return nodes[0].DeepCopy(), nil
+	}
+
+	return nil, nil
+}
+
+func (r *ReconcileNodeLink) findNodeFromMachineByIP(machine *mapiv1beta1.Machine) (*corev1.Node, error) {
+	if machine == nil {
```
frobware: I still think we should drop this; calling this with nil is broken elsewhere.

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,500 @@
...
+		klog.V(3).Infof("Found node %q for machine %q with internal IP %q", nodes[0].GetName(), machine.GetName(), machineInternalAddress)
+		return nodes[0].DeepCopy(), nil
+	}
+
+	klog.V(3).Infof("Matching node not found for machine %q with internal IP %q", machine.GetName(), machineInternalAddress)
+	return nil, nil
+}
+
+func (r *ReconcileNodeLink) findMachineFromNode(node *corev1.Node) (*mapiv1beta1.Machine, error) {
+	if node == nil {
```
frobware: I still think we should drop this; calling this with nil is broken elsewhere.

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,500 @@
...
+
+	machine, err = r.findMachineFromNodeByIP(node)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find node from machine %q by internal IP: %v", machine.GetName(), err)
+	}
+	return machine, nil
+}
+
+func (r *ReconcileNodeLink) findMachineFromNodeByProviderID(node *corev1.Node) (*mapiv1beta1.Machine, error) {
+	if node == nil {
```
frobware: I still think we should drop this; calling this with nil is broken elsewhere.

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -0,0 +1,500 @@
...
+
+	if len(machines) > 0 {
+		klog.V(3).Infof("Found machine %q for node %q with providerID %q", machines[0].GetName(), node.GetName(), node.Spec.ProviderID)
+		return machines[0].DeepCopy(), nil
+	}
+	return nil, nil
+}
+
+func (r *ReconcileNodeLink) findMachineFromNodeByIP(node *corev1.Node) (*mapiv1beta1.Machine, error) {
+	if node == nil {
```
frobware: I still think we should drop this; calling this with nil is broken elsewhere.

frobware:
/lgtm

Thanks for all the changes.

PR #330, opened 2019-06-17
Fix related objects to be included in status

enxebre:
As related objects belong to status subresource they were not being udpated as part of the client `Create` request.
This ensures they are included. In a follow up we'll include machines and machineSets
https://bugzilla.redhat.com/show_bug.cgi?id=1717633

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -5,12 +5,12 @@ import (
 	"reflect"
 	"strings"
 
-	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 
 	"github.com/golang/glog"
-
 	osconfigv1 "github.com/openshift/api/config/v1"
 	cvoresourcemerge "github.com/openshift/cluster-version-operator/lib/resourcemerge"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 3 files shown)


pkg/operator/status.go
```diff
@@ -148,37 +148,37 @@ func (optr *Operator) syncStatus(co *osconfigv1.ClusterOperator, conds []osconfi
...
+	co.Status.RelatedObjects = []osconfigv1.ObjectReference{
+		{
+			Group:    "",
+			Resource: "namespaces",
+			Name:     optr.namespace,
+		},
+	}
+	co, err = optr.osClient.ConfigV1().ClusterOperators().UpdateStatus(co)
+	if err != nil {
+		fmt.Errorf("error updating cluster operator status")
```
frobware: Where do we return an err if there's a genuine problem? The last line is `return co, nil`, should this be `return co, err`?

pkg/operator/status_test.go
```diff
@@ -102,3 +104,84 @@ func TestOperatorStatusProgressing(t *testing.T) {
...
+			osClient = fakeconfigclientset.NewSimpleClientset()
+		}
+		optr := Operator{
+			osClient:  osClient,
+			namespace: namespace,
+		}
+
+		co, err := optr.getOrCreateClusterOperator()
+		if err != nil {
+			t.Errorf("unexpected error: %v", err)
```
frobware: This should probably be a call to `Fatal`. Can the test continue of `co` is nil?

pkg/operator/status_test.go
```diff
@@ -102,3 +104,84 @@ func TestOperatorStatusProgressing(t *testing.T) {
 		assert.True(t, condition.LastTransitionTime.Equal(&conditionAfterAnotherSync.LastTransitionTime), "test-case %v expected LastTransitionTime not to be updated if condition state is same", i)
 	}
 }
+
+func TestGetOrCreateClusterOperator(t *testing.T) {
+	var namespace = "openshift-machine-api"
```
frobware: nit: I would chose something else; something else because we genuinely use this value and we want to assert we can plug in any value.

pkg/operator/status.go
```diff
@@ -148,38 +148,33 @@ func (optr *Operator) syncStatus(co *osconfigv1.ClusterOperator, conds []osconfi
...
+	// Related objects lets openshift/must-gather collect diagnostic content
+	co.Status.RelatedObjects = []osconfigv1.ObjectReference{
+		{
+			Group:    "",
+			Resource: "namespaces",
+			Name:     optr.namespace,
+		},
 	}
-	return co, nil
+	return optr.osClient.ConfigV1().ClusterOperators().UpdateStatus(co)
```
bison: This *always* calls `UpdateStatus()`? That seems like a weird side effect of retrieving the ClusterOperator object. I could see this spamming updates in the future if it's not already.

pkg/operator/status.go
```diff
@@ -148,36 +148,35 @@ func (optr *Operator) syncStatus(co *osconfigv1.ClusterOperator, conds []osconfi
...
+
+	// Related objects lets openshift/must-gather collect diagnostic content
+	relatedObjects := []osconfigv1.ObjectReference{
+		{
+			Group:    "",
+			Resource: "namespaces",
+			Name:     optr.namespace,
+		},
+	}
+	if !reflect.DeepEqual(co.Status.RelatedObjects, relatedObjects) {
```
bison: This all looks good, but I think we should be using the Kubernetes equality checking here:
https://godoc.org/k8s.io/apimachinery/pkg/api/equality

enxebre: thanks!

PR #332, opened 2019-06-24
Disable a machine from being fenced

emesika:
Documentation for the requirement to disable fencing using a special label set on the machine object

### DIFF
--- a/docs/proposals/disable-fencing.txt
+++ b/docs/proposals/disable-fencing.txt
@@ -1,0 +1,39 @@
+Temporarily disable machine fencing 
+===================================
+
+Summary
+----------------
+
+The Machine Healthcheck Controller (MHC) looks for failed nodes and deletes them after giving them grace time to recover (5 min by default). This feature provides a mechanism to prevent that from happening on a per-node basis.
+
+
+Motivation 
+----------------
+
+During upgrade, maintenance, and triage activities a node status may be not Ready for a while and we want a mechanism to exclude such a node from the fencing cycle.
+
+The administrator should be able to temporarily exclude nodes from being fenced.This will allow to track the machine manually without applying the machine fencing mechanism according to the system administrator considerations.
+
+User stories 
+-------------
+
+A machine is upgraded 
+A machine is set to maintenance for some reason.
+
+Goals 
+----------------
+
+Ability to temporarily exclude nodes from being fenced even if the fencing mechanism is active and the node is in a problematic status that makes it a candidate for fencing by MHC and enable administrators to handle the event manually.
+
+Proposal 
+---------------- 
+
+Set a annotation on a machine that describes if the node should be explicitly temporarily excluded from fencing mechanism handling.
+
+When MHC is looking for failed nodes, it will look for the healthchecking.openshift.io/disabled annotation on teh machine object and if it is set to "true" MHC will temporarily exclude this machine from the fencing flow.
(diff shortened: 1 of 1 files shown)


docs/proposals/disable-fencing.txt
```diff
@@ -0,0 +1,31 @@
...
+Summary
+----------------
+
+The Machine Healthcheck Controller (MHC) looks for failed nodes and deletes them after giving them grace time to recover (5 min by default). This feature provides a mechanism to prevent that from happening on a per-node basis.
+
+
+Motivation 
+----------------
+
+The administrator should be able to exclude nodes from being fenced.This will allow to track the machine manually without applying the machine fencing mechanism according to the system administrator considerations.
```
vikaschoudhary16: I am confused by the "fencing" term. In my understanding:
Fencing is the process of isolating a node to protect a cluster and its resources. Without fencing, a faulty node can cause data corruption in a cluster.

Your proposal looks like disabling machine healthchecking. is machine healthcking and fencing same thing?

docs/proposals/disable-fencing.txt
```diff
@@ -0,0 +1,31 @@
+Disable Machine Fencing
```
enxebre: why would you need this at all since mhc targets only machines of your choice via label selector?

cynepco3hahue: We have a maintenace controller, that should drain all VM's from the respective node, and also we do not want to fence it, because of it under the maintenance, but its temporary state and we do not want each time update node or MHC labels, because it easy to forgot to do it and disable fencing for the machine at all

enxebre: isn't this proposing exactly to use a label?

cynepco3hahue: this one who will set the label is node-maintenance controller, we just want respect this label under MHC and do not fence node
Added under the document

docs/proposals/disable-fencing.txt
markmc: I find the terminology confusing too.

We talk about health checking (noticing a failure) and remediation (currently just deletion).

I believe the point of both health checking and the remediation is (a) to allow safely moving workloads from unhealthy nodes and (b) attempting to bring the node back into a healthy state. And so we say this is fencing.

But if we're going to use that term at all, we should use it consistently.

e.g. README.md says:

> Machine healthcheck controller
> Reconciles desired state for MachineHealthChecks by ensuring that machines targeted by machineHealthCheck objects are healthy or remediated otherwise

Would we be happy to change that to:

> Machine healthcheck (fencing) controller
> Provides node level fencing by checking that machines targeted by MachineHealthCheck objects are healthy (as indicated by NodeConditions) and attempting remediation of unhealthy machines to bring them back to a healthy state

docs/proposals/disable-fencing.txt
vikaschoudhary16: I think this feature is useful. In some cases, it maybe desired to skip specific machines.
Another approach could be to support list of labels, that guide nodes to be skipped, as a configuration option on mhc

vikaschoudhary16:
PTAL https://github.com/openshift/cluster-api/pull/48

enxebre:
closing due to inactivity. Please create a PR against https://github.com/openshift/enhancements if still relevant

PR #333, opened 2019-06-24
Add support for node annotation to disable node fencing.

emesika:
Implementation for the requirement to disable fencing using a special label set on the machine object

### DIFF
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -28,10 +28,11 @@ import (
 )
 
 const (
-	machineAnnotationKey       = "machine.openshift.io/machine"
-	machineRebootAnnotationKey = "healthchecking.openshift.io/machine-remediation-reboot"
-	ownerControllerKind        = "MachineSet"
-	remediationStrategyReboot  = healthcheckingv1alpha1.RemediationStrategyType("reboot")
+	machineAnnotationKey           = "machine.openshift.io/machine"
+	machineRebootAnnotationKey     = "healthchecking.openshift.io/machine-remediation-reboot"
+	ownerControllerKind            = "MachineSet"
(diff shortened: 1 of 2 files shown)


emesika:
To do check list 
1) add unit tests 
2) add functional tests

vikaschoudhary16:
@emesika PTAL https://github.com/openshift/cluster-api/pull/48

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -31,6 +31,7 @@ import (
 const (
 	machineAnnotationKey = "machine.openshift.io/machine"
 	ownerControllerKind  = "MachineSet"
+	disableFencingLabel = "machine.openshift.io/healthcheck-disabled"
```
enxebre: if anything at all this should be an annotation

beekhof: @emesika ack. we'll also change to healthchecking.openshift.io/disabled

Do you have any other concerns about the concept in general?

enxebre: this will also need unit test. Can you please share a gdoc with all the rationale annotations you think you need, use cases and scenarious for the baremetal story?
If you want to opt out why don't just remove that machine from the MHC target or remove the MHC resource?

enxebre: Is there any reason why this can't be in the remediation controller for now?

beekhof: Generally its better that the work is not scheduled if we don't want it to take place. Otherwise it looks like there is work to do when there isn't, which can look messy and confuse admins.

enxebre:
Closing due to inactivity. Please create a proposal against https://github.com/openshift/enhancements if still relevant

PR #335, opened 2019-06-25
Machine metrics

vikaschoudhary16:
This PR is:
1. Adding basic prometheus metrics for reporting machine and machineset objects' state at cluster level
2. Adding logic to start the server which prometheus servers can use to scrape the metrics
3. Yamls for the objects required for integrating with openshift-monitoring

Sample output on querying the sever:
```
# HELP mapi_machine_info Information of the Machines currently managed by the machine controller.
# TYPE mapi_machine_info gauge
mapi_machine_info{api_version="machine.openshift.io/v1beta1",createdAt="1561525100",name="sdgfre-hqb4x-master-0",namespace="openshift-machine-api",node_ref="ip-10-0-131-137.us-east-2.compute.internal",spec_provider_id=""} 1
mapi_machine_info{api_version="machine.openshift.io/v1beta1",createdAt="1561525101",name="sdgfre-hqb4x-master-1",namespace="openshift-machine-api",node_ref="ip-10-0-153-127.us-east-2.compute.internal",spec_provider_id=""} 1
mapi_machine_info{api_version="machine.openshift.io/v1beta1",createdAt="1561525101",name="sdgfre-hqb4x-master-2",namespace="openshift-machine-api",node_ref="ip-10-0-165-12.us-east-2.compute.internal",spec_provider_id=""} 1
mapi_machine_info{api_version="machine.openshift.io/v1beta1",createdAt="1561525231",name="sdgfre-hqb4x-worker-us-east-2a-szfhb",namespace="openshift-machine-api",node_ref="ip-10-0-139-120.us-east-2.compute.internal",spec_provider_id="aws:///us-east-2a/i-01fd024fede672463"} 1
mapi_machine_info{api_version="machine.openshift.io/v1beta1",createdAt="1561525231",name="sdgfre-hqb4x-worker-us-east-2b-vfmkm",namespace="openshift-machine-api",node_ref="ip-10-0-146-195.us-east-2.compute.internal",spec_provider_id="aws:///us-east-2b/i-0bfcfe917b55d40ae"} 1
mapi_machine_info{api_version="machine.openshift.io/v1beta1",createdAt="1561525231",name="sdgfre-hqb4x-worker-us-east-2c-ms5gc",namespace="openshift-machine-api",node_ref="ip-10-0-166-251.us-east-2.compute.internal",spec_provider_id="aws:///us-east-2c/i-09f244e20e483f890"} 1
# HELP mapi_machine_items_total Count of machine objects currently at the apiserver
# TYPE mapi_machine_items_total gauge
mapi_machine_items_total 6
# HELP mapi_machine_set_info Information of the Machinesets currently managed by the machine controller
# TYPE mapi_machine_set_info gauge
mapi_machine_set_info{api_version="machine.openshift.io/v1beta1",createdAt="1561525102",name="sdgfre-hqb4x-worker-us-east-2a",namespace="openshift-machine-api"} 1
mapi_machine_set_info{api_version="machine.openshift.io/v1beta1",createdAt="1561525102",name="sdgfre-hqb4x-worker-us-east-2b",namespace="openshift-machine-api"} 1
mapi_machine_set_info{api_version="machine.openshift.io/v1beta1",createdAt="1561525103",name="sdgfre-hqb4x-worker-us-east-2c",namespace="openshift-machine-api"} 1
# HELP mapi_machine_set_info_spec_replicas Count of the Machinesets Spec Replicas.
# TYPE mapi_machine_set_info_spec_replicas gauge
mapi_machine_set_info_spec_replicas{name="sdgfre-hqb4x-worker-us-east-2a",namespace="openshift-machine-api"} 1
mapi_machine_set_info_spec_replicas{name="sdgfre-hqb4x-worker-us-east-2b",namespace="openshift-machine-api"} 1
mapi_machine_set_info_spec_replicas{name="sdgfre-hqb4x-worker-us-east-2c",namespace="openshift-machine-api"} 1
# HELP mapi_machine_set_status_available_replicas Information of the mapi managed Machinesets' status for available replicas.
# TYPE mapi_machine_set_status_available_replicas gauge
mapi_machine_set_status_available_replicas{name="sdgfre-hqb4x-worker-us-east-2a",namespace="openshift-machine-api"} 1
mapi_machine_set_status_available_replicas{name="sdgfre-hqb4x-worker-us-east-2b",namespace="openshift-machine-api"} 1
mapi_machine_set_status_available_replicas{name="sdgfre-hqb4x-worker-us-east-2c",namespace="openshift-machine-api"} 1
# HELP mapi_machine_set_status_ready_replicas Information of the mapi managed Machinesets' status for ready replicas.
# TYPE mapi_machine_set_status_ready_replicas gauge
mapi_machine_set_status_ready_replicas{name="sdgfre-hqb4x-worker-us-east-2a",namespace="openshift-machine-api"} 1
mapi_machine_set_status_ready_replicas{name="sdgfre-hqb4x-worker-us-east-2b",namespace="openshift-machine-api"} 1
mapi_machine_set_status_ready_replicas{name="sdgfre-hqb4x-worker-us-east-2c",namespace="openshift-machine-api"} 1
# HELP mapi_machine_set_status_replicas Information of the mapi managed Machinesets' status for replicas.
# TYPE mapi_machine_set_status_replicas gauge
mapi_machine_set_status_replicas{name="sdgfre-hqb4x-worker-us-east-2a",namespace="openshift-machine-api"} 1
mapi_machine_set_status_replicas{name="sdgfre-hqb4x-worker-us-east-2b",namespace="openshift-machine-api"} 1
mapi_machine_set_status_replicas{name="sdgfre-hqb4x-worker-us-east-2c",namespace="openshift-machine-api"} 1
# HELP mapi_machineset_items_total Count of machinesets at the apiserver.
# TYPE mapi_machineset_items_total gauge
mapi_machineset_items_total 3
```


mrunalp:
nit: Can we add some description in the commit? :)

derekwaynecarr:
i would anticipate that we integrate with https://github.com/openshift/openshift-state-metrics

@brancz can you weigh in here?

brancz:
As we just have a downstream here and we anticipate that this will be an upstream resource one day I originally advised the team to contain this within the controller. I think this particular API has a thin line as we’re actually registering it under an OpenShift umbrella. I think either ways would be appropriate. A plus I can see with it being in the controller could be that upstream ships it in upstream controllers as well. Kind of in the way that the upstream one would unlikely to be eligible for kube state metrics.

install/0000_30_machine-api-operator_09_rbac.yaml
```diff
@@ -212,6 +212,13 @@ rules:
       - list
       - patch
 
+  - apiGroups:
+      - machine.openshift.io
+    resources:
+      - '*'
+    verbs:
+      - '*'
```
bison: Does the operator really need full permissions here? Is read enough?

cmd/nodelink-controller/main.go
```diff
@@ -26,7 +26,7 @@ func main() {
 	printVersion()
 
 	watchNamespace := flag.String("namespace", "", "Namespace that the controller watches to reconcile machine-api objects. If unspecified, the controller watches for machine-api objects across all namespaces.")
-	klog.InitFlags(nil)
+	//klog.InitFlags()
```
bison: Why is this commented out?

install/0000_30_machine-api-operator_09_rbac.yaml
```diff
@@ -269,3 +276,37 @@ subjects:
   - kind: ServiceAccount
     name: machine-api-operator
     namespace: openshift-machine-api
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: prometheus-k8s-cluster-machine-api-operator
```
bison: This is a total nit, but should this just be `prometheus-k8s-cluster-machine-api-operator`? Not sure where the `-cluster-` comes from.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,239 @@
...
+	// MachineCountDesc is a metric about machine object count in the cluster
+	MachineCountDesc = prometheus.NewDesc("mapi_machine_items_total", "Count of machine objects currently at the apiserver", nil, nil)
+	// MachineSetCountDesc Count of machineset object count at the apiserver
+	MachineSetCountDesc = prometheus.NewDesc("mapi_machineset_items_total", "Count of machinesets at the apiserver.", nil, nil)
+
+	//MachineInfo Information of the Machines currently managed by the machine controller.
+	MachineInfo = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "mapi_machine_info",
+		Help: "Information of the Machines currently managed by the machine controller.",
+	}, []string{"name", "namespace", "createdAt",
```
bison: Are we sure we want things like timestamps and instance IDs as labels in these metrics? These things have a tendency in some cases to create a huge number of individual time series in Prometheus, and in the case of a `createdAt` timestamp it not super useful to query for.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	client, err := runtimeclient.New(config, runtimeclient.Options{})
+	if err != nil {
+		return nil, fmt.Errorf("error creating runtime client: %v", err.Error())
+	}
+
+	return client, nil
+}
+
+func NewMachineCollector(kubeconfig string) (*MachineCollector, error) {
```
bison: Don't we construct a client right after initializing this when the operator starts? Would it be easier to just pass a client instead of a kubeconfig?

vikaschoudhary16: mao creates two clients, one for concrete kubernetes apis and other for openshift specific apis. These clients will not work for CRDs. I think, in future we can try using controller-runtime abstractions/client instead of using client-go library directly for creating k8s client and os client. And then, may be able to work with a single controller-runtime provided client.

pkg/metrics/metrics.go
vikaschoudhary16: May be you are right. Another option is to let these metrics in at this point so that we have basic support working and then refine metrics in follow-up PRs. I am fine either way/
/cc @enxebre

install/0000_30_machine-api-operator_09_rbac.yaml
enxebre: `prometheus-k8s-machine-api-operator`?

pkg/metrics/metrics.go
enxebre: `These clients will not work for CRDs` not sure what you mean, we could analogous create another client for machineAPI or even better use listers() as we do for the deployments

vikaschoudhary16: `dynamic` and `kubernetes` are two packages of client-go. `kubernetes` package provides clientset and Kubernetes resource specific clientset methods for accessing standard Kubernetes APIs.  Deployments are standard kubernetes api type. Informers are then built on top of this client to use cache.
mao is using `kubernetes` pkg. [kubeclient](https://github.com/openshift/machine-api-operator/blob/master/cmd/machine-api-operator/controller_context.go#L30) returns [interface](https://github.com/kubernetes/client-go/blob/master/kubernetes/clientset.go#L63) which has only standard k8s apis. 
For CRDs, [dynamic](https://github.com/kubernetes/client-go/tree/master/dynamic) client is to be used. 

controller-runtime provides [ a single abstracted client over](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/client.go#L46) these two type of clients that client-go supports. This single client automatically handles both type of queries. I am using this controller-runtime provided client here to list machines. It already using informers underneath.

If we want to use a single client, we might have to make mao to use controller-runtime provided client and then same could be used by metrics pkg.

Please correct if something i misunderstood.

enxebre: you can do same we do for featuregates/deployments, see https://github.com/openshift/cluster-api/tree/openshift-4.2-cluster-api-0.1.0/pkg/client
"It already using informers underneath." that's true when instantiated via manager, not in this case
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L261
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L278

pkg/metrics/metrics.go
enxebre: I'd keep the `provider id`, also if we rename node_ref to node we could in a follow up have an alert for machines with no nodes/invalid nodes the likes of ```(machine_info unless on(node) kube_node_info) > 0```

cmd/machine-api-operator/client_builder.go
```diff
@@ -24,6 +25,11 @@ func (cb *ClientBuilder) OpenshiftClientOrDie(name string) osclientset.Interface
 	return osclientset.NewForConfigOrDie(rest.AddUserAgent(cb.config, name))
 }
 
+// MachineClientOrDie returns the machine api client interface for machine api objects.
+func (cb *ClientBuilder) MachineClientOrDie(name string) mapiclientset.Interface {
+	return mapiclientset.NewForConfigOrDie(rest.AddUserAgent(cb.config, name))
```
enxebre: I'm confused, why are you generating anything at all here?
Isn't all we need available here https://github.com/openshift/cluster-api/tree/openshift-4.2-cluster-api-0.1.0/pkg/client?
Can we just do exactly the same we do for the openshift config
mapiclientset "github.com/openshift/cluster-api/pkg/client/clientset_generated/clientset/typed/machine/v1beta1"

func (cb *ClientBuilder) KubeClientOrDie(name string) mapiclientset.MachinesGetter {
	c, _ := mapiclientset.NewForConfig(cb.config)
	return c
}

vikaschoudhary16: vendored cluster-api at mao is having only clientset, no lister/informer. https://github.com/openshift/machine-api-operator/tree/master/vendor/github.com/openshift/cluster-api/pkg

I will see how to vendor missing parts as well, thanks!

Gopkg.lock
```diff
@@ -384,7 +384,7 @@
```
enxebre: haven we have a seprate commit for go dep changes

pkg/generated/clientset/versioned/clientset.go
```diff
@@ -28,8 +28,6 @@ import (
 type Interface interface {
```
enxebre: not sure what's this commit, can we just have a commit for godep changes

cmd/machine-api-operator/client_builder.go
vikaschoudhary16: Resolved

Gopkg.lock
vikaschoudhary16: Resolved

pkg/generated/clientset/versioned/clientset.go
vikaschoudhary16: Resolved

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,189 @@
...
+		return
+	}
+	metric, err := prometheus.NewConstMetric(MachineSetCountDesc, prometheus.GaugeValue, float64(len(machineSetList)))
+	if err != nil {
+		ScrapeFailedCounter.With(prometheus.Labels{"kind": "Machineset-count", "reason": err.Error()}).Inc()
+		return
+	}
+	ch <- metric
+
+	for _, machineSet := range machineSetList {
```
frobware: I think it is easier to read without the temporaries: 

```go
	for _, machineSet := range machineSetList {
		MachineSetInfo.With(prometheus.Labels{
			"name":        machineSet.Name,
			"namespace":   machineSet.Namespace,
			"createdAt":   strconv.FormatInt(machineSet.GetCreationTimestamp().Time.Unix(), 10),
			"api_version": machineSet.APIVersion}).Set(float64(1))

		MachineSetInfoSpecReplicas.With(prometheus.Labels{
			"name":      machineSet.Name,
			"namespace": machineSet.Namespace}).Set(float64(pointer.Int32PtrDerefOr(machineSet.Spec.Replicas, 0)))

		MachineSetStatusAvailableReplicas.With(prometheus.Labels{
			"name":      machineSet.Name,
			"namespace": machineSet.Namespace,
		}).Set(float64(machineSet.Status.AvailableReplicas))

		MachineSetStatusReadyReplicas.With(prometheus.Labels{
			"name":      machineSet.Name,
			"namespace": machineSet.Namespace,
		}).Set(float64(machineSet.Status.ReadyReplicas))

		MachineSetStatusReplicas.With(prometheus.Labels{
			"name":      machineSet.Name,
			"namespace": machineSet.Namespace,
		}).Set(float64(machineSet.Status.ReadyReplicas))
	}
```

Also note the safe deref of `Replicas` in this ^^ diff: 

```go
		MachineSetInfoSpecReplicas.With(prometheus.Labels{
			"name":      machineSet.Name,
			"namespace": machineSet.Namespace}).Set(float64(pointer.Int32PtrDerefOr(machineSet.Spec.Replicas, 0)))
```

cmd/machine-api-operator/client_builder.go
```diff
@@ -24,23 +25,32 @@ func (cb *ClientBuilder) OpenshiftClientOrDie(name string) osclientset.Interface
...
 		config, err = rest.InClusterConfig()
 	}
-	if err != nil {
-		return nil, err
-	}
-
-	return &ClientBuilder{
-		config: config,
-	}, nil
+	return config, err
```
frobware: Not sure if this code needs to handle the `InClusterConfig()` case as it already handled by:

```go
// BuildConfigFromFlags is a helper function that builds configs from a master
// url or a kubeconfig filepath. These are passed in as command line flags for cluster
// components. Warnings should reflect this usage. If neither masterUrl or kubeconfigPath
// are passed in we fallback to inClusterConfig. If inClusterConfig fails, we fallback
// to the default config.
func BuildConfigFromFlags(masterUrl, kubeconfigPath string) (*restclient.Config, error) {
	if kubeconfigPath == "" && masterUrl == "" {
		klog.Warningf("Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.")
		kubeconfig, err := restclient.InClusterConfig()
		if err == nil {
			return kubeconfig, nil
		}
		klog.Warning("error creating inClusterConfig, falling back to default config: ", err)
	}
	return NewNonInteractiveDeferredLoadingClientConfig(
		&ClientConfigLoadingRules{ExplicitPath: kubeconfigPath},
		&ConfigOverrides{ClusterInfo: clientcmdapi.Cluster{Server: masterUrl}}).ClientConfig()
}
```

cmd/machine-api-operator/client_builder.go
frobware: If you vendored the missing parts does https://github.com/openshift/machine-api-operator/pull/335#discussion_r300611726 still hold true as suggested?

install/0000_30_machine-api-operator_09_rbac.yaml
```diff
@@ -212,6 +212,14 @@ rules:
       - list
       - patch
 
+  - apiGroups:
```
enxebre: why do we need this, the role already has `machine.openshift.io` for `openshift-machine-api`?

pkg/generated/clientset/versioned/clientset.go
enxebre: can you please explain why we need this commit?

cmd/machine-api-operator/client_builder.go
vikaschoudhary16: no, now informers and listers are being used as suggested.

cmd/machine-api-operator/client_builder.go
vikaschoudhary16: Since this part of logic is not being added in this PR, so i would like to leave it as it is.

install/0000_30_machine-api-operator_09_rbac.yaml
vikaschoudhary16: This is what we get if we dont add these roles:

```
E0710 14:33:47.569176       1 reflector.go:126] github.com/openshift/cluster-api/pkg/client/informers_generated/externalversions/factory.go:118: Failed to list *v1beta1.MachineSet: machinesets.machine.openshift.io is forbidden: User "system:serviceaccount:openshift-machine-api:machine-api-operator" cannot list resource "machinesets" in API group "machine.openshift.io" at the cluster scope
```

vikaschoudhary16: machine-api-operator cluster role does not have permission to list/watch resources that belong to machine.openshift.io group. Thats what my understanding is.

Gopkg.lock
```diff
@@ -384,11 +384,25 @@
 
 [[projects]]
   branch = "openshift-4.2-cluster-api-0.1.0"
-  digest = "1:efca7e07cbca01b615c1f9057da17b10ea6b7e43e263226a3ef04eebe16f34cf"
```
enxebre: Can we please have a proper commit message here: e.g "Revendor for fetching cluster-api informers" or so

install/0000_30_machine-api-operator_09_rbac.yaml
enxebre: that's because you are telling the informer to watch cluster wide. Can you please narrow the scope to the passed namespace? i.e same that we are doing for the kube informer `informers.WithNamespace(targetNamespace)`

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,185 @@
...
+	prometheus.MustRegister(MachineSetStatusReplicas)
+}
+
+type MachineCollector struct {
+	machineLister    machinelisters.MachineLister
+	machineSetLister machinelisters.MachineSetLister
+	namespace        string
+}
+
+func NewMachineCollector(machineInformer machineinformers.MachineInformer, machinesetInformer machineinformers.MachineSetInformer, namespace string) (*MachineCollector, error) {
```
enxebre: does this need to return error at all?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,185 @@
...
+func init() {
+	prometheus.MustRegister(ScrapeFailedCounter)
+	prometheus.MustRegister(MachineInfo)
+	prometheus.MustRegister(MachineSetInfo)
+	prometheus.MustRegister(MachineSetInfoSpecReplicas)
+	prometheus.MustRegister(MachineSetStatusAvailableReplicas)
+	prometheus.MustRegister(MachineSetStatusReadyReplicas)
+	prometheus.MustRegister(MachineSetStatusReplicas)
+}
+
```
enxebre: can you add comment with a reference to what interface is this implementing?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,185 @@
...
+			"name":      machineSet.Name,
+			"namespace": machineSet.Namespace,
+		}).Set(float64(machineSet.Status.AvailableReplicas))
+
+		MachineSetStatusReadyReplicas.With(prometheus.Labels{
+			"name":      machineSet.Name,
+			"namespace": machineSet.Namespace,
+		}).Set(float64(machineSet.Status.ReadyReplicas))
+
+		MachineSetStatusReplicas.With(prometheus.Labels{
```
enxebre: what is this, can we drop it? seems a copy of the above `MachineSetStatusReadyReplicas`

cmd/machine-api-operator/controller_context.go
```diff
@@ -29,14 +31,17 @@ type ControllerContext struct {
 func CreateControllerContext(cb *ClientBuilder, stop <-chan struct{}, targetNamespace string) *ControllerContext {
 	kubeClient := cb.KubeClientOrDie("kube-shared-informer")
 	configClient := cb.OpenshiftClientOrDie("config-shared-informer")
+	machineClient := cb.MachineClientOrDie("machine-shared-informer")
 
 	kubeNamespacedSharedInformer := informers.NewSharedInformerFactoryWithOptions(kubeClient, resyncPeriod()(), informers.WithNamespace(targetNamespace))
 	configSharedInformer := configinformersv1.NewSharedInformerFactoryWithOptions(configClient, resyncPeriod()())
+	machineSharedInformer := machineinformersv1beta1.NewSharedInformerFactoryWithOptions(machineClient, resyncPeriod()())
```
enxebre: can we please make this namespaced? also where is this informer waiting to be synced e.g `cache.WaitForCacheSync`, `*.Informer().HasSynced`

install/0000_30_machine-api-operator_09_rbac.yaml
vikaschoudhary16: clusterrole `machine-api-operator` has permissions only  for `config.openshift.io` api group. Even after namespaced informers, i am getting following error:
```
Failed to list *v1beta1.Machine: machines.machine.openshift.io is forbidden: User "system:serviceaccount:openshift-machine-api:machine-api-operator" cannot list resource "machines" in API group "machine.openshift.io" in the namespace "openshift-machine-api"
``` 
I think we will need to add permissions that i had initially.

pkg/metrics/metrics.go
vikaschoudhary16: This is for Status.Replicas reporting. Fixed the typo error

cmd/machine-api-operator/start.go
```diff
@@ -101,3 +115,43 @@ func startControllers(ctx *ControllerContext) {
...
+		componentNamespace)
+	go func() {
+		if !cache.WaitForCacheSync(ctx.Stop,
+			machineInformer.Informer().HasSynced,
+			machinesetInformer.Informer().HasSynced) {
+			glog.Error("Failed to sync caches")
+			return
+		}
+	}()
+	glog.Info("Synced up machine api informer caches")
```
enxebre: this always log this regardless the result of WaitForCacheSync? unless I'm reading it wrong

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,186 @@
...
+	// MachineSetInfoSpecReplicas Count of the Machinesets Spec Replicas.
+	MachineSetInfoSpecReplicas = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "mapi_machine_set_info_spec_replicas",
+		Help: "Count of the Machinesets Spec Replicas.",
+	}, []string{"name", "namespace"})
+
+	// MachineSetStatusAvailableReplicas Information of the Machinesets' status for available replicas.
+	MachineSetStatusAvailableReplicas = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "mapi_machine_set_status_available_replicas",
+		Help: "Information of the mapi managed Machinesets' status for available replicas.",
```
enxebre: Machinsets' <- typo?
Also here you say "mapi managed" whereus above " managed by the machine controller" can we use consistent message?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,186 @@
...
+	MachineInfo = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "mapi_machine_info",
+		Help: "Information of the Machines currently managed by the machine controller.",
+	}, []string{"name", "namespace", "createdAt",
+		"spec_provider_id", "node", "api_version"})
+
+	// MachineSetInfo Information of the Machinesets currently reported by the apiserver
+	MachineSetInfo = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "mapi_machine_set_info",
+		Help: "Information of the Machinesets currently managed by the machine controller",
```
enxebre: missing end period

cmd/machine-api-operator/start.go
```diff
@@ -101,3 +115,42 @@ func startControllers(ctx *ControllerContext) {
...
+		}
+		metricsPort = v
+	}
+	glog.V(4).Info("Starting server to serve prometheus metrics")
+	go startHTTPMetricServer(fmt.Sprintf(":%d", metricsPort))
+}
+
+func startHTTPMetricServer(metricsPort string) {
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", prometheus.Handler())
```
enxebre: this is deprecated?

cmd/machine-api-operator/start.go
```diff
@@ -101,3 +115,42 @@ func startControllers(ctx *ControllerContext) {
...
+		componentNamespace)
+	go func() {
+		if !cache.WaitForCacheSync(ctx.Stop,
+			machineInformer.Informer().HasSynced,
+			machinesetInformer.Informer().HasSynced) {
+			glog.Fatal("Failed to sync caches")
+		}
+		glog.Info("Synced up machine api informer caches")
+	}()
+	prometheus.MustRegister(machineMetricsCollector)
```
enxebre: should this be inside the go subroutine so the WaitForCacheSync blocks the rest of the execution?

cmd/machine-api-operator/start.go
```diff
@@ -101,3 +115,43 @@ func startControllers(ctx *ControllerContext) {
...
+
+func startMetricsCollectionAndServer(ctx *ControllerContext) {
+	machineInformer := ctx.MachineInformerFactory.Machine().V1beta1().Machines()
+	machinesetInformer := ctx.MachineInformerFactory.Machine().V1beta1().MachineSets()
+	machineMetricsCollector := metrics.NewMachineCollector(
+		machineInformer,
+		machinesetInformer,
+		componentNamespace)
+	go func() {
+		if !cache.WaitForCacheSync(ctx.Stop,
```
frobware: Why do we do this here?

This starts a server and this seems unrelated.

pkg/metrics/metrics.go
simonpasquier: From my Prometheus developer's POV, `createdAt` smells like something that shouldn't be a label. As @bison said, it won't useful because label values are opaque to Prometheus and you can't do any arithmetic on them. What would work is an additional metric reporting this value (eg `mapi_machine_created_at_seconds`).

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,186 @@
...
+		Name: "mapi_machine_info",
+		Help: "Information of the mapi managed Machine.",
+	}, []string{"name", "namespace", "createdAt",
+		"spec_provider_id", "node", "api_version"})
+
+	// MachineSetInfo Information of the Machinesets currently reported by the apiserver
+	MachineSetInfo = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "mapi_machine_set_info",
+		Help: "Information of the mapi managed Machineset.",
+	}, []string{"name", "namespace", "createdAt", "api_version"})
```
simonpasquier: ditto for `createdAt`

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,186 @@
...
+	machineinformers "github.com/openshift/cluster-api/pkg/client/informers_generated/externalversions/machine/v1beta1"
+	machinelisters "github.com/openshift/cluster-api/pkg/client/listers_generated/machine/v1beta1"
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/utils/pointer"
+)
+
+var (
+	// MachineCountDesc is a metric about machine object count in the cluster
+	MachineCountDesc = prometheus.NewDesc("mapi_machine_items_total", "Count of machine objects currently at the apiserver", nil, nil)
```
simonpasquier: It looks like a gauge so shouldn't have with a `_total` suffix which are reserved for counters by convention.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,186 @@
...
+	// MachineSetCountDesc Count of machineset object count at the apiserver
+	MachineSetCountDesc = prometheus.NewDesc("mapi_machineset_items_total", "Count of machinesets at the apiserver.", nil, nil)
```
simonpasquier: ditto

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,186 @@
...
+	MachineSetStatusReplicas = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "mapi_machine_set_status_replicas",
+		Help: "Information of the mapi managed Machineset's status for replicas.",
+	}, []string{"name", "namespace"})
+
+	// ScrapeFailedCounter is a Prometheus metric, which counts errors during metrics collection.
+	ScrapeFailedCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "mapi_scrape_failure_total",
+		Help: "Total count of scrape failures.",
+	}, []string{"kind", "reason"})
```
simonpasquier: I don't think you need labels here, especially with unpredictable value set as for `reason`.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,186 @@
...
+	for _, machine := range machineList {
+		mMeta := machine.ObjectMeta
+		typeMeta := machine.TypeMeta
+		mSpec := machine.Spec
+		providerid := ""
+		if mSpec.ProviderID != nil {
+			providerid = *mSpec.ProviderID
+		}
+
+		MachineInfo.With(prometheus.Labels{
```
simonpasquier: I'm not sure if it's relevant in this case but worth mentioning: once registered, the metric will continue to exist for the entire lifetime of the program even though the machine doesn't exist anymore.

install/0000_30_machine-api-operator_13_servicemonitor.yaml
```diff
@@ -0,0 +1,20 @@
...
+kind: ServiceMonitor
+metadata:
+  labels:
+    k8s-app: machine-api-operator
+  name: machine-api-operator
+  namespace: openshift-machine-api
+spec:
+  endpoints:
+    - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      interval: 30s
```
simonpasquier: It should be the default scrape interval for Prometheus so there's no need to specify it again.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+	MachineCountDesc = prometheus.NewDesc("mapi_machine_items_count", "Count of machine objects currently at the apiserver", nil, nil)
+	// MachineSetCountDesc Count of machineset object count at the apiserver
+	MachineSetCountDesc = prometheus.NewDesc("mapi_machineset_items_count", "Count of machinesets at the apiserver", nil, nil)
+	// MachineInfoDesc is a metric about machine object info in the cluster
+	MachineInfoDesc = prometheus.NewDesc("mapi_machine_created", "Timestamp of the mapi managed Machine creation time", []string{"name", "namespace", "spec_provider_id", "node", "api_version"}, nil)
+	// MachineSetInfoDesc is a metric about machine object info in the cluster
+	MachineSetInfoDesc = prometheus.NewDesc("mapi_machineset_created", "Timestamp of the mapi managed Machineset creation time", []string{"name", "namespace", "api_version"}, nil)
+
+	// MachineSetStatusAvailableReplicasDesc is the information of the Machineset's status for available replicas.
+	MachineSetStatusAvailableReplicasDesc = prometheus.NewDesc("mapi_machine_set_status_available_replicas", "Information of the mapi managed Machineset's status for available replicas", []string{"name", "namespace"}, nil)
```
simonpasquier: (nit) kube-state-metrics uses names like `<base name>_<status>` for similar metrics (eg `kube_deployment_status_replicas_available`, `kube_deployment_status_replicas_updated`, ...). It might be good to follow the same convention (IMO it is easier this way as the autocomplete would list the metrics together).

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,143 @@
...
+func (mc MachineCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- MachineCountDesc
+	ch <- MachineSetCountDesc
+}
+
+// Collect implements the prometheus.Collector interface.
+func (mc MachineCollector) collectMachineMetrics(ch chan<- prometheus.Metric) {
+	machineList, err := mc.listMachines()
+	if err != nil {
+		ScrapeFailedCounter.With(prometheus.Labels{"kind": "Machine-count", "reason": err.Error()}).Inc()
```
simonpasquier: `reason` isn't a label anymore.

vikaschoudhary16:
@simonpasquier Thanks a lot. Will address your comments in a follow up PR

PR #336, opened 2019-06-27
Add updateNodeRef for node link controller

enxebre:
Originally the logic for setting the nodeRef in a machine was split between the cluster-api node controller and opt-int external tooling that would set an annotation for the former controller to watch. There's no reason for this to keep leaving in two different places, furthermore the cluster-api node controller is deprecated. This PR unifies the current logic and behaviour in one single nodelink controller.
Counterpart PR https://github.com/openshift/cluster-api/pull/47

Note:
This PR is to unify and simplify current code in one single nodelink controller while maintaining current behaviour to simplify farther development and maintainability.
Once we promote MHC, we could decide we don't care about the machine controller taking care of recreating a machine if something out of band deletes the cloud instance, and remove the newNodeReady != currentNodeReady check, and just watch machines instead of nodes and exclusively rely on MHC

### DIFF
--- a/pkg/controller/nodelink/nodelink_controller.go
+++ b/pkg/controller/nodelink/nodelink_controller.go
@@ -8,6 +8,7 @@ import (
 	mapiv1beta1 "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -38,6 +39,7 @@ type ReconcileNodeLink struct {
 	// and emulate Client.List.MatchingField behaviour
 	listNodesByFieldFunc    func(key, value string) ([]corev1.Node, error)
 	listMachinesByFieldFunc func(key, value string) ([]mapiv1beta1.Machine, error)
+	nodeReadinessCache      map[string]bool
 }
 
 // Add creates a new Nodelink Controller and adds it to the Manager. The Manager will set fields on the Controller
@@ -146,6 +148,7 @@ func newReconciler(mgr manager.Manager) (*ReconcileNodeLink, error) {
 	r := ReconcileNodeLink{
 		client: mgr.GetClient(),
 	}
+	r.nodeReadinessCache = make(map[string]bool)
 
 	// This is useful for unit testing so we can mock cache IndexField
 	// and emulate Client.List.MatchingField behaviour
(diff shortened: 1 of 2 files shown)


pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -234,10 +246,43 @@ func (r *ReconcileNodeLink) Reconcile(request reconcile.Request) (reconcile.Resu
 		}
 	}
 
-	// TODO(alberto): bring nodeRef logic here and drop cluster-api node controller
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcileNodeLink) updateNodeRef(machine *mapiv1beta1.Machine, node *corev1.Node) error {
```
bison: Could we add a comment here, or at the call site, about the reason for "touching" the machine resource on node status change. It's not immediately clear.

enxebre: done https://github.com/openshift/machine-api-operator/pull/336/files#diff-69885f1e6e1f6b4cb889ce80021c95a1R273

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -474,3 +519,16 @@ func (r *ReconcileNodeLink) listMachinesByField(key, value string) ([]mapiv1beta
 	}
 	return machineList.Items, nil
 }
+
+// isNodeReady returns true if a node is ready; false otherwise.
+func isNodeReady(node *corev1.Node) bool {
+	if node == nil || &node.Status == nil {
```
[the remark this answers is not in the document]
enxebre: removed, don't really need the check

pkg/controller/nodelink/nodelink_controller.go
```diff
@@ -210,6 +213,15 @@ func (r *ReconcileNodeLink) Reconcile(request reconcile.Request) (reconcile.Resu
 		return reconcile.Result{}, nil
 	}
 
+	if err := r.updateNodeRef(machine, node); err != nil {
+		return reconcile.Result{}, fmt.Errorf("error updating nodeRef for machine %q and node %q: %v", machine.GetName(), node.GetName(), err)
+	}
+
+	if !node.DeletionTimestamp.IsZero() {
```
vikaschoudhary16: Why updating NodeRef on the machine but skipping to update machine annotation on node?

vikaschoudhary16: nevermind. Now i see handling of deletion timsestamp case in `updateNodeRef`.

pkg/controller/nodelink/nodelink_controller_test.go
```diff
@@ -705,3 +706,67 @@ func TestIndexMachineByInternalIP(t *testing.T) {
 		}
 	}
 }
+
+func TestUpdateNodeRef(t *testing.T) {
+	testCases := []struct {
+		machine            *mapiv1beta1.Machine
+		node               *corev1.Node
+		nodeRef            *corev1.ObjectReference
+		nodeReadinessCache map[string]bool
```
vikaschoudhary16: where is this being used?

enxebre: https://github.com/openshift/machine-api-operator/pull/336/files#diff-69885f1e6e1f6b4cb889ce80021c95a1R273

PR #338, opened 2019-07-03
Update actuator-pkg dependency

cynepco3hahue:
Bring https://github.com/openshift/cluster-api-actuator-pkg/pull/70 into mao so the mhc tests can be run through a different make target (out of `make test-e2e`).

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -404,7 +404,7 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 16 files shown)


enxebre:
once this settles down, please do as we do for cao https://github.com/openshift/cluster-autoscaler-operator/blob/master/hack/e2e.sh

ingvagabund:
> once this settles down, please do as we do for cao https://github.com/openshift/cluster-autoscaler-operator/blob/master/hack/e2e.sh

Done through https://github.com/openshift/machine-api-operator/pull/345

PR #340, opened 2019-07-03
Monitoring for ironic-prometheus-exporter

iurygregory:
This PR depends on #302 and adds the changes to enable monitoring
in the machine-api-operator that will allow Prometheus
to collect data from the ironic-prometheus-exporter[1] that
runs in the ironic-image [2]. 

- Added monitoring label to namespace yaml
- Added monitoring information to the rbac yaml
- Added Service for the ironic-prometheus-exporter
- Added the ServiceMonitor for the ironic-prometheus-exporter
- Added PrometheusRule with alert for baremetal_temp_celsius
metric.

[1] https://github.com/metal3-io/ironic-prometheus-exporter
[2] https://github.com/metal3-io/ironic-image

### DIFF
--- a/install/0000_30_machine-api-operator_10_service-ironic-prometheus-exporter.yaml
+++ b/install/0000_30_machine-api-operator_10_service-ironic-prometheus-exporter.yaml
@@ -1,0 +1,22 @@
+{{if .usingBareMetal}}
+apiVersion: monitoring.coreos.com/v1
+kind: Service
+metadata:
+  name: metal3-baremetalhost-controller
+  namespace: openshift-machine-api
+  labels:
+    app: ironic-exporter
+spec:
+  ports:
+    - name: http
+      protocol: TCP
+      port: 9608
+      targetPort: 9608
+  selector:
+    app: ironic-exporter
+  clusterIP: None
+  type: ClusterIP
+  sessionAffinity: None
+status:
+  loadBalancer: {}
+{{- end}}
--- a/install/0000_90_machine-api-operator_04_servicemonitor-ironic-prometheus-exporter.yaml
+++ b/install/0000_90_machine-api-operator_04_servicemonitor-ironic-prometheus-exporter.yaml
@@ -1,0 +1,22 @@
+{{if .usingBareMetal}}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app: ironic-exporter
+  name: metal3-baremetalhost-controller
+  namespace: openshift-machine-api
+spec:
+  endpoints:
+  - port: "9608-tcp"
+    scheme: http
+    path: /metrics
+    targetPort: 9608
+  jobLabel: app
+  namespaceSelector:
+    matchNames:
+    - metal3-baremetalhost-controller
+  selector:
+    matchLabels:
+      app: ironic-exporter
+{{- end}}
--- a/install/0000_90_machine-api-operator_05_prometheusrules-ironic-prometheus-exporter.yaml
+++ b/install/0000_90_machine-api-operator_05_prometheusrules-ironic-prometheus-exporter.yaml
@@ -1,0 +1,27 @@
+{{if .usingBareMetal}}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: metal3-baremetalhost-controller
+  namespace: openshift-machine-api
+spec:
+  groups:
+  - name: metal3-baremetalhost-controller
+    rules:
+    - alert: HighCPUTemperature
+      annotations:
+        summary: "The baremetal node {{ $labels.node_name }} CPU {{ $labels.entity_id }} is too high"
+        description: "The baremetal node {{ $labels.node_name }} CPU {{ $labels.entity_id }} was too high in the past 5 minutes. Last measurement {{ $value }}"
+      expr: baremetal_temp_celsius > 96
+      for: 5m
+      labels:
+        severity: warning
+    - alert: LowCPUTemperature
+      annotations:
+        summary: "The baremetal node {{ $labels.node_name }} CPU {{ $labels.entity_id }} is too low"
+        description: "The baremetal node {{ $labels.node_name }} CPU {{ $labels.entity_id }} was too low in the past 5 minutes. Last measurement {{ $value }}"
+      expr: baremetal_temp_celsius < 3
+      for: 5m
+      labels:
+        severity: warning
+{{- end}}
--- a/pkg/operator/baremetal_pod.go
+++ b/pkg/operator/baremetal_pod.go
@@ -290,6 +290,7 @@ func newMetal3Containers(config *OperatorConfig, baremetalProvisioningConfig Bar
 	containers = append(containers, createContainerMetal3IronicApi(config, baremetalProvisioningConfig))
 	containers = append(containers, createContainerMetal3IronicInspector(config, baremetalProvisioningConfig))
 	containers = append(containers, createContainerMetal3StaticIpManager(config, baremetalProvisioningConfig))
+	containers = append(containers, createContainerMetal3IronicExporter(config))
 	return containers
 }
 
@@ -421,6 +422,21 @@ func createContainerMetal3StaticIpManager(config *OperatorConfig, baremetalProvi
 			buildEnvVar("PROVISIONING_IP", "provisioning_ip", baremetalProvisioningConfig),
 			buildEnvVar("PROVISIONING_INTERFACE", "provisioning_interface", baremetalProvisioningConfig),
 		},
+	}
+	return container
+}
+
+func createContainerMetal3IronicExporter(config *OperatorConfig) corev1.Container {
+
+	container := corev1.Container{
+		Name:            "metal3-ironic-exporter",
+		Image:           config.BaremetalControllers.Ironic,
+		ImagePullPolicy: "Always",
+		SecurityContext: &corev1.SecurityContext{
+			Privileged: pointer.BoolPtr(true),
+		},
+		Command:      []string{"/bin/runironic-exporter"},
+		VolumeMounts: volumeMounts,
 	}
 	return container
 }


iurygregory:
@openshift/sig-monitoring

iurygregory:
This will be used by #302

brancz:
I don’t actually see the service or daemonset for this exporter so it’s hard to say if this is correct. Generally speaking it looks ok.

russellb:
This will need to be reworked so that it's only applied when the platform is of type "baremetal".  I think it would make sense to include this as part of #302, since that's the PR that would deploy the ironic promethus exporter as well.

brancz:
Yes it would be easier to verify this if it included the deployment of the exporter.

hardys:
Agreed, this should either be rolled into #302 or this PR can be rebased onto that one when ready.

iurygregory:
I would like some reviews to see if this is ok even if it will depends on PR #302 =)

install/0000_30_machine-api-operator_09_rbac.yaml
```diff
@@ -269,3 +269,21 @@ subjects:
   - kind: ServiceAccount
     name: machine-api-operator
     namespace: openshift-machine-api
+
+---
```
simonpasquier: IIUC it should happen later in the deployment because the `openshift-monitoring` namespace won't exist yet (the cluster monitoring operator runs at the 50 run-level) with the other [OLM](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/operators.md#how-do-i-get-added-as-a-special-run-level) stuff. See the [cluster-kube-apiserver-operator](https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/manifests/0000_90_kube-apiserver-operator_01_prometheusrole.yaml) for instance.

You also need to [bind](https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/manifests/0000_90_kube-apiserver-operator_02_prometheusrolebinding.yaml) this role to the `prometheus-k8s` service account.

install/0000_30_machine-api-operator_15_ironic-exporter-prometheusrules.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: metal3-baremetalhost-controller
+  namespace: openshift-machine-api
+spec:
+  groups:
+  - name: metal3-baremetalhost-controller
+    rules:
+    - alert: HighCPUTemperature
```
simonpasquier: Are those real alerts?

install/0000_30_machine-api-operator_09_rbac.yaml
iurygregory: Thanks Simon, going to look

install/0000_30_machine-api-operator_15_ironic-exporter-prometheusrules.yaml
iurygregory: I've created these alerts based on the values from the Hardware specification that we are going to use. I will send you the doc

install/0000_30_machine-api-operator_09_rbac.yaml
iurygregory: I've followed the example from the monitoring guide [1], would it be ok to move this rbac to a specific file so I can change the level according to OLM etc? Also, if I move to other level I would have to move the `Service`, `ServiceMonitor` and `PrometheusRule` right?


[1] https://github.com/openshift/cluster-monitoring-operator/blob/23201e012586d4864ca23593621f843179c47412/assets/prometheus-k8s/role-specific-namespaces.yaml#L35-L50

s-urbaniak: @iurygregory afaik it's fine to separate the rbac definition in a separate file as long as it stays in the same runlevel. your statement regarding the service, servicemonitor, and prometheus rules are correct :+1:

iurygregory: @s-urbaniak Thanks for your review, the rbac would need to change to `kind: RoleBinding` ?

s-urbaniak: @iurygregory you actually need both:

1. a role that refers to the services/endpoints/pods resources in your namespace and the allowed verbs get/list/watch (random example https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/manifests/0000_90_kube-apiserver-operator_01_prometheusrole.yaml)
2. a role binding which associates the above role with the `prometheus-k8s` service account (random example https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/manifests/0000_90_kube-apiserver-operator_02_prometheusrolebinding.yaml)

install/0000_30_machine-api-operator_09_rbac.yaml
```diff
@@ -269,3 +269,21 @@ subjects:
...
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: Role
+  metadata:
+    name: prometheus-k8s
+    namespace: openshift-monitoring
```
s-urbaniak: as mentioned in https://github.com/openshift/machine-api-operator/pull/340/files/e25d72e68810159a0fcfa547ad02c37cc1130981#r304315286, this should be your namespace, i.e. `openshift-machine-api`.

install/0000_30_machine-api-operator_09_rbac.yaml
iurygregory: @s-urbaniak the `RoleBinding` level to run would be 90? or can be 30 also?

install/0000_30_machine-api-operator_09_rbac.yaml
iurygregory: ack

install/0000_30_machine-api-operator_09_rbac.yaml
s-urbaniak: @iurygregory as discussed out of band, I suggest runlevel 90, there is no monitoring stack yet in runlevel 30.

install/0000_30_machine-api-operator_15_ironic-exporter-prometheusrules.yaml
simonpasquier: Ack

install/0000_30_machine-api-operator_15_ironic-exporter-prometheusrules.yaml
```diff
@@ -0,0 +1,25 @@
...
+      annotations:
+        summary: "The baremetal node {{ $labels.node_name }} CPU {{ $labels.entity_id }} is too high"
+        description: "The baremetal node {{ $labels.node_name }} CPU {{ $labels.entity_id }} is too high in the past minute. Last measurement {{ $value }}"
```
simonpasquier: It says `in the past minute` while the for condition is `5m`.

iurygregory: Changed

install/0000_90_machine-api-operator_01_prometheusrole.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
elfosardo: this line needs to go away :)

install/0000_90_machine-api-operator_03_ironic-exporter-service.yaml
```diff
@@ -0,0 +1,21 @@
+---
```
elfosardo: ditto

install/0000_90_machine-api-operator_04_ironic-exporter-servicemonitor.yaml
```diff
@@ -0,0 +1,21 @@
+---
```
elfosardo: ditto

install/0000_90_machine-api-operator_01_prometheusrole.yaml
```diff
@@ -0,0 +1,17 @@
...
+- apiVersion: rbac.authorization.k8s.io/v1
```
simonpasquier: This is already defined in `0000_30_machine-api-operator_09_rbac.yaml`

https://github.com/openshift/machine-api-operator/blob/b10f7c0c52f8bd98c534ec0be7e94238b6f84cb3/install/0000_30_machine-api-operator_09_rbac.yaml#L296-L312

And the same is true for the other Prometheus operator CRs.

@vikaschoudhary16 said offline that it would be tackled in another PR so you probably to coordinate there.

hardys:
Since we had to switch to an older version of ironic that doesn't contain the exporter, ref https://github.com/openshift/machine-api-operator/pull/380 this should probably be abandoned until we can consume a newer ironic version?

sadasu:
@hardys Is the version of Ironic we are using now in deploying "metal3" appropriate to restart this work?

iurygregory:
@sadasu I'm re-working this PR, the version on metal3 is fine but for openshift we will first need  https://github.com/openshift/ironic-image/pull/27 and also an update on BMO to be aware of the `ironic-prometheus-exporter` container

hardys:
@sadasu can you take another look at this please - in particular I'd like to confirm how we can ensure the Service/ServiceMonitor etc only get created for the platform: baremetal case where we expect the metal3 pod part to be enabled?

install/0000_30_machine-api-operator_10_service-ironic-prometheus-exporter.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: monitoring.coreos.com/v1
+kind: Service
+metadata:
+  name: metal3-baremetalhost-controller
```
sadasu: I am not sure what the "name" here should be. But, "metal3-baremetalhost-controller" is the name of the baremetal_operator controller. See https://github.com/metal3-io/baremetal-operator/blob/master/pkg/controller/baremetalhost/baremetalhost_controller.go#L105

install/0000_90_machine-api-operator_04_servicemonitor-ironic-prometheus-exporter.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app: ironic-exporter
+  name: metal3-baremetalhost-controller
```
sadasu: See https://github.com/openshift/machine-api-operator/pull/340/files#r382116862.

install/0000_90_machine-api-operator_05_prometheusrules-ironic-prometheus-exporter.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: metal3-baremetalhost-controller
```
sadasu: See https://github.com/openshift/machine-api-operator/pull/340/files#r382116862.

sadasu:
The .yaml files added in the /install dir will be installed by the CVO for all platforms. What would the behavior be when the ironic-prometheus-exporter container is not running for not "BareMetal" platform types?

sadasu:
You also need to add the ironic-prometheus-exporter image to https://github.com/openshift/machine-api-operator/blob/master/install/image-references and https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/fixtures/images.json.

install/0000_30_machine-api-operator_10_service-ironic-prometheus-exporter.yaml
sadasu: The name we are using for the deployment is "metal3". See https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/baremetal_pod.go#L119.

iurygregory:
> You also need to add the ironic-prometheus-exporter image to https://github.com/openshift/machine-api-operator/blob/master/install/image-references and https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/fixtures/images.json.

The exporter doesn't need a new image AFAIK, we will be using the ironic image (https://github.com/openshift/ironic-image) , we only need to add the container for the exporter.

simonpasquier:
AFAICT creating the service, service monitor and prometheus rule with CVO shouldn't be an issue for non-baremetal platforms as the service monitor would match 0 endpoints and the Prometheus rule would generate 0 series. That being said, a better option would be to have the machine API operator manage those resources as it knows whether or not the Ironic exporter is deployed.

install/0000_90_machine-api-operator_05_prometheusrules-ironic-prometheus-exporter.yaml
```diff
@@ -0,0 +1,25 @@
...
+  namespace: openshift-machine-api
+spec:
+  groups:
+  - name: metal3-baremetalhost-controller
+    rules:
+    - alert: HighCPUTemperature
+      annotations:
+        summary: "The baremetal node {{ $labels.node_name }} CPU {{ $labels.entity_id }} is too high"
+        description: "The baremetal node {{ $labels.node_name }} CPU {{ $labels.entity_id }} was too high in the past 5 minutes. Last measurement {{ $value }}"
```
enxebre: wouldn't this trigger for non baremetal envs?

enxebre: Should porbably rename `syncBaremetalControllers` to something more generic e.g `syncBaremetal`
Then `syncBaremetal`  calls `syncBaremetalControllers`  and `syncBaremetalMonitoring`

iurygregory: @enxebre I don't think it would trigger in non baremetal envs, check @simonpasquier comment. 
The sync part you are talking would be if we make the yamls go code right?

iurygregory:
Does anyone know people from the CVO that can take a look at the PR to see if this approach (using the if in the yamls is valid?)

enxebre:
>Does anyone know people from the CVO that can take a look at the PR to see if this approach (using the if in the yamls is valid?)

What's that coming from? It's breaking all tests. Anything under install/ is to be managed by the CVO and the only way to discriminate are profiles which are not a good case for this scenario. https://github.com/openshift/enhancements/pull/200

Can we please move any baremetal specific under https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/sync.go#L97? Any monitoring related resource could be managed there, you can rename `syncBaremetalControllers` and let it manage the monitoring resources that you can define as yaml or golang code whatever you feel more comfortable with.

iurygregory:
> > Does anyone know people from the CVO that can take a look at the PR to see if this approach (using the if in the yamls is valid?)
> 
> What's that coming from? It's breaking all tests. Anything under install/ is to be managed by the CVO and the only way to discriminate are profiles which are not a good case for this scenario. [openshift/enhancements#200](https://github.com/openshift/enhancements/pull/200)
> 
> Can we please move any baremetal specific under https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/sync.go#L97? Any monitoring related resource could be managed there, you can rename `syncBaremetalControllers` and let it manage the monitoring resources that you can define as yaml or golang code whatever you feel more comfortable with.
Hi @enxebre , I was talking with some people from kuryr and they pointed me to https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/kuryr/008-admission.secret.yaml#L1 (not sure if MAO has support for that, but the conditions would ensure that the yamls would only apply when it's true).

Is there any other directory where I can place the yamls and make a call in syncBaremetalControllers to apply?

enxebre:
>Hi @enxebre , I was talking with some people from kuryr and they pointed me to https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/kuryr/008-admission.secret.yaml#L1 (not sure if MAO has support for that, but the conditions would ensure that the yamls would only apply when it's true).
Is there any other directory where I can place the yamls and make a call in syncBaremetalControllers to apply?

That's using bindata to compile and embed the yamls into static golang code that you can template at runtime as you wish. You could follow the same approach here. e.g
https://github.com/openshift/cluster-kube-apiserver-operator/tree/master/bindata/v4.1.0

https://github.com/openshift/cluster-kube-apiserver-operator/tree/master/pkg/operator/v410_00_assets

https://github.com/openshift/cluster-kube-apiserver-operator/blob/3161546a248f20eb67231017d0ae43c245bfa4cb/pkg/recovery/apiserver.go#L134-L177

Or alternatively you can define your resources as API golang types e.g https://github.com/openshift/cluster-autoscaler-operator/blob/8e6f95038c9eee84ef7e305e2e1f4960c918b30d/pkg/controller/clusterautoscaler/monitoring.go

PR #341, opened 2019-07-04
Introduce gosec for security checks

JAORMX:
gosec does static code analysis and checks for common security issues in golang codebases.

This PR introduces the tool, and sets it up as a test target in the Makefile; it also fixes a couple of warnings that were found (which were probably not very problematic).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,11 @@ build-integration: ## Build integration test binary
 test-e2e: ## Run openshift specific e2e test
 	$(MAKE) -C ./vendor/github.com/openshift/cluster-api-actuator-pkg $@
 
+.PHONY: test-sec
+test-sec:
+	@which gosec 2> /dev/null >&1 || { echo "gosec must be installed to lint code";  exit 1; }
+	gosec -severity medium --confidence medium -quiet ./...
+
 .PHONY: deploy-kubemark
 deploy-kubemark:
 	kustomize build config | kubectl apply -f -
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -67,6 +67,11 @@ build-integration: ## Build integration test binary
 test-e2e: ## Run openshift specific e2e test
 	$(MAKE) -C ./vendor/github.com/openshift/cluster-api-actuator-pkg $@
 
+.PHONY: test-sec
+test-sec:
+	@which gosec 2> /dev/null >&1 || { echo "gosec must be installed to lint code";  exit 1; }
```
vikaschoudhary16: should we run this in containerized env like  `gen-crd` , `update-codegen` etc?

JAORMX: I don't think so, since this doesn't really generate code. It merely prints out a report if security issues are found.

enxebre:
/approve
thanks! @frobware PTAL

Makefile
frobware: I think the premise is: if this has value and finds problems then we should gate CI on the result of running the tool. It's not about generating code, more that we have a rule that runs this tool during CI.

JAORMX: @frobware right! The first step is to add this so folks can run it through make; the next step is to add it to CI. We already added this for the machine-config-operator: https://github.com/openshift/release/pull/4273

frobware:
/lgtm

Thanks!

PR #343, opened 2019-07-06
Fix flaking unit test TestOperatorSync_NoOp

vikaschoudhary16:
Problem was that the informer cache was not being ensured to be synced before running the tests.

```
E0705 08:14:52.443801    6362 sync.go:37] Error syncing cluster api controller: timed out waiting for the condition
--- FAIL: TestOperatorSync_NoOp (320.10s)
    --- FAIL: TestOperatorSync_NoOp/AWS (300.07s)
    	operator_test.go:165: 
    			Error Trace:	operator_test.go:165
    			Error:      	Received unexpected error:
    			            	timed out waiting for the condition
    			Test:       	TestOperatorSync_NoOp/AWS
    			Messages:   	unexpected sync failure
    	operator_test.go:166: 
```

### DIFF
--- a/pkg/operator/operator_test.go
+++ b/pkg/operator/operator_test.go
@@ -1,11 +1,7 @@
 package operator
 
 import (
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
-	"os"
-	"path/filepath"
 	"testing"
 	"time"
 
@@ -127,25 +123,6 @@ func TestOperatorSync_NoOp(t *testing.T) {
 			expectedNoop: true,
 		},
 	}
-
-	tempDir, err := ioutil.TempDir("", "TestOperatorSync")
-	if err != nil {
-		t.Fatalf("could not create the temp dir: %v", err)
-	}
-	defer os.RemoveAll(tempDir)
-
-	images := Images{
-		MachineAPIOperator: "test-mao",
-	}
-	imagesAsJSON, err := json.Marshal(images)
-	if err != nil {
-		t.Fatalf("failed to marshal images: %v", err)
-	}
-	imagesFilePath := filepath.Join(tempDir, "test-images.json")
-	if err := ioutil.WriteFile(imagesFilePath, imagesAsJSON, 0666); err != nil {
-		t.Fatalf("could not write the images file: %v", err)
-	}
-
 	for _, tc := range cases {
 		t.Run(string(tc.platform), func(t *testing.T) {
 			infra := &v1.Infrastructure{
@@ -158,15 +135,10 @@ func TestOperatorSync_NoOp(t *testing.T) {
 			}
 
 			stopCh := make(<-chan struct{})
-			optr := newFakeOperator(nil, []runtime.Object{infra}, stopCh)
-			optr.imagesFile = imagesFilePath
+			optr := newFakeOperator(nil, []runtime.Object{newFeatureGate(v1.TechPreviewNoUpgrade), infra}, stopCh)
+			go optr.Run(2, stopCh)
 
-			err = optr.sync("test-key")
-			if !assert.NoError(t, err, "unexpected sync failure") {
-				t.Fatal()
-			}
-
-			err = wait.PollImmediate(1*time.Second, 5*time.Second, func() (bool, error) {
+			err := wait.PollImmediate(1*time.Second, 5*time.Second, func() (bool, error) {
 				_, err := optr.deployLister.Deployments(targetNamespace).Get(deploymentName)
 				if err != nil {
 					t.Logf("Failed to get %q deployment: %v", deploymentName, err)


vikaschoudhary16:
```
--- FAIL: TestReconcileWithoutUnhealthyConditionsConfigMap (0.02s)
	machinehealthcheck_controller_test.go:262: Test case: recentlyUnhealthy. Expected RequeueAfter between: 5m0s and 5m1s, got: {true 4m59.992730885s}
```
This seems another flake.

cynepco3hahue:
lgtm, I think we can also run `go optr.Run(2, stopCh)` instead of `sync` operation and it will fix the problem, but the current solution also works for me

ingvagabund:
@vikaschoudhary16 thanks for the easy fix. 

> lgtm, I think we can also run go optr.Run(2, stopCh) instead of sync operation and it will fix the problem, but the current solution also works for me

As @cynepco3hahue, `go optr.Run(2, stopCh)` is the right way how to fix the case since it already waits for the changes and sets additional bits.

ingvagabund:
/hold

please gimme me some time to review the PR

enxebre:
the name of the test `TestOperatorSync_NoOp` seems to be outdated since the test is doing more and not even calling the sync function explicitely

pkg/operator/operator_test.go
```diff
@@ -158,15 +135,10 @@ func TestOperatorSync_NoOp(t *testing.T) {
...
 			stopCh := make(<-chan struct{})
-			optr := newFakeOperator(nil, []runtime.Object{infra}, stopCh)
-			optr.imagesFile = imagesFilePath
-
-			err = optr.sync("test-key")
-			if !assert.NoError(t, err, "unexpected sync failure") {
-				t.Fatal()
-			}
+			optr := newFakeOperator(nil, []runtime.Object{newFeatureGate(v1.TechPreviewNoUpgrade), infra}, stopCh)
+			go optr.Run(2, stopCh)
```
frobware: The problem I have with this is the design of `Run()` itself.

That looks like:

```go
// Run runs the machine config operator.
func (optr *Operator) Run(workers int, stopCh <-chan struct{}) {
	defer utilruntime.HandleCrash()
	defer optr.queue.ShutDown()

	glog.Info("Starting Machine API Operator")
	defer glog.Info("Shutting down Machine API Operator")

	if !cache.WaitForCacheSync(stopCh,
		optr.deployListerSynced,
		optr.featureGateCacheSynced) {
		glog.Error("Failed to sync caches")
		return
	}
	glog.Info("Synced up caches")
	for i := 0; i < workers; i++ {
		go wait.Until(optr.worker, time.Second, stopCh)
	}

	<-stopCh
}
```

- The call to `WaitForCacheSync` can fail, yet we ignore the error.
- Callers of `operator.Run()` can silently fail
- Perhaps the cache in some runs is not sync'd and therefore flakes?

frobware:
> Problem was that the informer cache was not being ensured to be synced before running the tests.

See: https://github.com/openshift/machine-api-operator/pull/343#discussion_r301477428

pkg/operator/operator_test.go
vikaschoudhary16: Agree that optr.Run() should be returning the error ideally. Though we have not seen "Failed to sync caches" error log in any of those flakes.

This PR is for the flake mentioned in the description. Lets create another PR for returning the error.

frobware: > This PR is for the flake mentioned in the description. Lets create another PR for returning the error.

Please can we hold for the moment.

frobware:
I see how it fixes the flake

/lgtm

pkg/operator/operator_test.go
```diff
@@ -158,15 +135,10 @@ func TestOperatorSync_NoOp(t *testing.T) {
...
 
 			stopCh := make(<-chan struct{})
-			optr := newFakeOperator(nil, []runtime.Object{infra}, stopCh)
-			optr.imagesFile = imagesFilePath
-
-			err = optr.sync("test-key")
-			if !assert.NoError(t, err, "unexpected sync failure") {
-				t.Fatal()
-			}
+			optr := newFakeOperator(nil, []runtime.Object{newFeatureGate(v1.TechPreviewNoUpgrade), infra}, stopCh)
```
ingvagabund: Normally, I would be against passing `newFeatureGate(v1.TechPreviewNoUpgrade)` object to the fake informer as it is no needed for the test. Though, the operator processes an item only when a deployment or a feature gate object is created/updated/deleted. Without passing either of those, the operator workers will not process anything. So, passing the feature gate object is actually a trigger to render new deployment. In the real cluster there has to be at least one deployment available in order to trigger the processing loop.

PR #346, opened 2019-07-08
Add reboot annotation

cynepco3hahue:
Instead of deleting the unhealthy machine under the bare metal provider
we want only to reboot it, the one who will responsible for the reboot will
be the machine-remediation controller and MHC will just set reboot annotation on
the node.

### DIFF
--- a/install/0000_30_machine-api-operator_09_rbac.yaml
+++ b/install/0000_30_machine-api-operator_09_rbac.yaml
@@ -133,6 +133,14 @@ rules:
       - list
       - watch
 
+  - apiGroups:
+      - config.openshift.io
+    resources:
+      - infrastructures
+      - infrastructures/status
+    verbs:
+      - get
+
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/golang/glog"
+	configv1 "github.com/openshift/api/config/v1"
 	mapiv1 "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	healthcheckingv1alpha1 "github.com/openshift/machine-api-operator/pkg/apis/healthchecking/v1alpha1"
 	"github.com/openshift/machine-api-operator/pkg/util"
@@ -29,8 +30,9 @@ import (
 )
 
 const (
-	machineAnnotationKey = "machine.openshift.io/machine"
-	ownerControllerKind  = "MachineSet"
+	machineAnnotationKey       = "machine.openshift.io/machine"
+	machineRebootAnnotationKey = "healthchecking.openshift.io/machine-remediation-reboot"
+	ownerControllerKind        = "MachineSet"
 )
 
 // Add creates a new MachineHealthCheck Controller and adds it to the Manager. The Manager will set fields on the Controller
@@ -45,18 +47,23 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
-	r := &ReconcileMachineHealthCheck{
-		client: mgr.GetClient(),
-		scheme: mgr.GetScheme(),
-	}
-
+	c := mgr.GetClient()
 	ns, err := util.GetNamespace(util.ServiceAccountNamespaceFile)
 	if err != nil {
-		return r, err
+		return nil, err
 	}
 
-	r.namespace = ns
-	return r, nil
+	infra, err := getClusterInfrastructure(c)
+	if err != nil {
+		return nil, err
+	}
+
+	return &ReconcileMachineHealthCheck{
+		client:    mgr.GetClient(),
+		scheme:    mgr.GetScheme(),
+		namespace: ns,
+		platform:  infra.Status.Platform,
+	}, nil
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
@@ -78,6 +85,7 @@ type ReconcileMachineHealthCheck struct {
 	client    client.Client
 	scheme    *runtime.Scheme
 	namespace string
+	platform  configv1.PlatformType
 }
 
 // Reconcile reads that state of the cluster for MachineHealthCheck, machine and nodes objects and makes changes based on the state read
@@ -167,7 +175,8 @@ func getMachineHealthCheckListOptions() *client.ListOptions {
 
 func remediate(r *ReconcileMachineHealthCheck, machine *mapiv1.Machine) (reconcile.Result, error) {
 	glog.Infof("Initialising remediation logic for machine %s", machine.Name)
-	if isMaster(*machine, r.client) {
+
+	if isMaster(*machine, r.client) && !isBareMetalProvider(r.platform) {
 		glog.Infof("The machine %s is a master node, skipping remediation", machine.Name)
 		return reconcile.Result{}, nil
 	}
@@ -213,6 +222,24 @@ func remediate(r *ReconcileMachineHealthCheck, machine *mapiv1.Machine) (reconci
 
 		// apply remediation logic, if at least one condition last more than specified timeout
 		if unhealthyForTooLong(nodeCondition, conditionTimeout) {
+			if isBareMetalProvider(r.platform) {
+				// we already have reboot annotation on the node, stop reconcile
+				if _, ok := node.Annotations[machineRebootAnnotationKey]; ok {
+					return reconcile.Result{}, nil
+				}
+
+				if node.Annotations == nil {
+					node.Annotations = map[string]string{}
+				}
+
+				glog.Infof("Baremetal machine %s has been unhealthy for too long, adding reboot annotation", machine.Name)
+				node.Annotations[machineRebootAnnotationKey] = ""
+				if err := r.client.Update(context.TODO(), node); err != nil {
+					return reconcile.Result{}, err
+				}
+				return reconcile.Result{}, nil
+			}
+
 			glog.Infof("machine %s has been unhealthy for too long, deleting", machine.Name)
 			if err := r.client.Delete(context.TODO(), machine); err != nil {
 				glog.Errorf("failed to delete machine %s, requeuing referenced node", machine.Name)
@@ -351,4 +378,20 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
 		}
 	}
 	return false
+}
+
+func getClusterInfrastructure(c client.Client) (*configv1.Infrastructure, error) {
+	infra := &configv1.Infrastructure{}
+	key := types.NamespacedName{
+		Namespace: metav1.NamespaceNone,
+		Name:      "cluster",
+	}
+	if err := c.Get(context.TODO(), key, infra); err != nil {
+		return nil, err
+	}
+	return infra, nil
+}
+
+func isBareMetalProvider(platform configv1.PlatformType) bool {
+	return platform == configv1.BareMetalPlatformType
 }
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
@@ -1,11 +1,13 @@
 package machinehealthcheck
 
 import (
+	"context"
 	"fmt"
 	"reflect"
 	"testing"
 	"time"
 
+	configv1 "github.com/openshift/api/config/v1"
 	mapiv1alpha1 "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
 	healthcheckingv1alpha1 "github.com/openshift/machine-api-operator/pkg/apis/healthchecking/v1alpha1"
 	"github.com/openshift/machine-api-operator/pkg/util/conditions"
@@ -32,6 +34,7 @@ func init() {
 	// Add types to scheme
 	mapiv1alpha1.AddToScheme(scheme.Scheme)
 	healthcheckingv1alpha1.AddToScheme(scheme.Scheme)
+	configv1.AddToScheme(scheme.Scheme)
 }
 
 func TestHasMatchingLabels(t *testing.T) {
@@ -170,6 +173,7 @@ func testReconcile(t *testing.T, remediationWaitTime time.Duration, initObjects 
 	machineWithoutNodeRef.Status.NodeRef = nil
 
 	machineHealthCheck := maotesting.NewMachineHealthCheck("machineHealthCheck")
+	infra := maotesting.NewInfrastructure(configv1.AWSPlatformType)
 
 	testsCases := []struct {
 		machine  *mapiv1alpha1.Machine
@@ -233,6 +237,7 @@ func testReconcile(t *testing.T, remediationWaitTime time.Duration, initObjects 
 		objects := []runtime.Object{}
 		objects = append(objects, initObjects...)
 		objects = append(objects, machineHealthCheck)
+		objects = append(objects, infra)
 		if tc.machine != nil {
 			objects = append(objects, tc.machine)
 		}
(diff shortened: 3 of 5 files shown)


pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -213,6 +221,24 @@ func remediate(r *ReconcileMachineHealthCheck, machine *mapiv1.Machine) (reconci
 
 		// apply remediation logic, if at least one condition last more than specified timeout
 		if unhealthyForTooLong(nodeCondition, conditionTimeout) {
+			if isBareMetalProvider(infra) {
+				// we already have reboot annotation on the node, stop reconcile
+				if _, ok := node.Annotations[annotationReboot]; ok {
+					return reconcile.Result{}, nil
```
beekhof: Does this mean it is up to the entity processing the annotation to power off the node delete the `Node` if reboot fails?

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -327,3 +331,32 @@ func TestUnhealthyForTooLong(t *testing.T) {
 		}
 	}
 }
+
+func TestRemediationOnBareMetalProvider(t *testing.T) {
```
beekhof: Seems like we're missing a positive BM test to see that the annotation is created.

vikaschoudhary16: +1, just create unhealthy node and verify if after reconcile annotation got added or not.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -213,6 +221,24 @@ func remediate(r *ReconcileMachineHealthCheck, machine *mapiv1.Machine) (reconci
...
+				}
+
+				if node.Annotations == nil {
+					node.Annotations = map[string]string{}
+				}
+
+				glog.Infof("bare metal machine %s has been unhealthy for too long, rebooting", machine.Name)
```
vikaschoudhary16: s/bare metal/Baremetal

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -30,6 +31,7 @@ import (
 
 const (
 	machineAnnotationKey = "machine.openshift.io/machine"
+	annotationReboot     = "metal3.io/BareMetalHostReboot"
```
vikaschoudhary16: should we have a generic annotation key and not specific to metal3?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
vikaschoudhary16: s/adding reboot annotation

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -213,6 +221,24 @@ func remediate(r *ReconcileMachineHealthCheck, machine *mapiv1.Machine) (reconci
 
 		// apply remediation logic, if at least one condition last more than specified timeout
 		if unhealthyForTooLong(nodeCondition, conditionTimeout) {
+			if isBareMetalProvider(infra) {
+				// we already have reboot annotation on the node, stop reconcile
+				if _, ok := node.Annotations[annotationReboot]; ok {
```
vikaschoudhary16: Are we sure we want to annotate node and not the machine?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
cynepco3hahue: exactly, I created two controllers under https://github.com/cynepco3hahue/machine-remediation-operator repository
- the first one is responsible to monitor nodes and create the `MachineRemediation` object in case if it recognizes the reboot annotation on the node(we can remove this one, ones the MHC will create the `MachineRemediation` object instead of add annotation on the node)
- the second one, monitors `MachineRemediation` and applies relevant remediation logic, it includes process rebooting, removing reboot annotation from the node once the reboot succeeded and remove the node once the reboot failed

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
cynepco3hahue: @vikaschoudhary16  Do you know how can I do it? I tried to do it, and at least with the controller-runtime fake client it does not update the node, I can see it enters the update method but the node object that I send to the fake client does not update

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
cynepco3hahue: @vikaschoudhary16 Does `machineremediation.openshift.io/reboot` suit well?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
cynepco3hahue: @vikaschoudhary16 To be honest, I do not have a strong opinion, but in node case, if reboot failed and we delete the node I do not need in addition to remove the annotation, so currently I prefer this way

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
vikaschoudhary16: @cynepco3hahue i can try. will let you know once done.

cynepco3hahue: No matter already got it:)

cynepco3hahue:
@vikaschoudhary16 @beekhof Updated PR, can you please take a look?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
vikaschoudhary16: @cynepco3hahue  Lets say 'node' is k8s node object and 'instance' is the corresponding machine. Assuming you meant, if an instance reboot fails and we delete the instance, there will be no need to remove annotation from the k8s node object. Did i get you right?

If yes, i am not sure if external controller should delete the instance on reboot failure. Because instance deletion will not remediate the machine actually. New instance will not be able to become k8s node because of failing CSR. Instead, on reboot failing, if machine object is deleted, that would remediate the situation.

Another thing which i am not very confident is that if controller handling "reboot" annotation, should do "delete". Will not it be a bit inconsistent?

cynepco3hahue: @vikaschoudhary16 Nope I meant the k8s node, under the bare metal platform we do not want to delete the machine object at all(it the reason why we have all this W/A and additional controllers), but we still want to release workload in case when reboot failed, so we will delete the node(you can do it because you already know that you rebooted the host via power management and it does not run any workloads)

vikaschoudhary16: @cynepco3hahue copying from the slack for the record:
ok, so if node is annotated and to handle this first, node is deleted and then instance is rebooted. In failure case, machine object will hang there without NodeRef. And an alert will generated after some configured time ~10 min.
Now if we discuss, alternative way, in which machine is being annotated..when alert for nil NodeRef, would this annotation of node reboot be helpful for the operator to understand why NodeRef is nil? As a clue on machine object

vikaschoudhary16: /cc @enxebre

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -30,6 +31,7 @@ import (
 
 const (
 	machineAnnotationKey = "machine.openshift.io/machine"
+	annotationReboot     = "machineremediation.openshift.io/reboot"
```
enxebre: `machineremediation.healthchecking.openshift.io`?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -167,7 +169,13 @@ func getMachineHealthCheckListOptions() *client.ListOptions {
 
 func remediate(r *ReconcileMachineHealthCheck, machine *mapiv1.Machine) (reconcile.Result, error) {
 	glog.Infof("Initialising remediation logic for machine %s", machine.Name)
-	if isMaster(*machine, r.client) {
+
+	infra, err := getClusterInfrastructure(r.client)
```
enxebre: can this be fetch only once at instantiation time?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
ingvagabund: `s/annotationReboot/machineRebootAnnotationKey`

ingvagabund: Given `healthchecking.openshift.io` is the CRD group name I would prefer to still keep `healthchecking.openshift.io` prefix without subgrouping it into `machineremediation`. If any, you might go with `healthchecking.openshift.io/machine-remediation-reboot` or similar.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +378,19 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
 	return false
 }
+
+func getClusterInfrastructure(c client.Client) (*configv1.Infrastructure, error) {
+	infra := &configv1.Infrastructure{}
+	key := types.NamespacedName{
+		Namespace: metav1.NamespaceNone,
+		Name:      "cluster",
+	}
+	if err := c.Get(context.TODO(), key, infra); err != nil {
```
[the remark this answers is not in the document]
ingvagabund: I would prefer to use informer here so you don't have to query the apiserver everytime you request for infrastructure object. Plus, as Alberto says, you can check the infrastructure platform only once since it's unlikely we will support hybrid installations in 4.2. So calling it only once will do as well.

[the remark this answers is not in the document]
cynepco3hahue: I think if I will fetch it only once it will be good enough also without informer

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go

cgwalters:
I was just made aware of this by @ericavonb - we are working on rebooting in the MCO as well https://github.com/openshift/machine-config-operator/issues/662
Will try to comment more intelligently in a bit.

cgwalters:
Still haven't looked in detail but two questions:

 - How is this (or is it) handling reboot loops?  What if the machine comes back up unhealthy, do we just keep rebooting it?
 - What's the actual mechanism used to reboot?  From the IaaS side or by a privileged pod on the node?

cgwalters:
On the reboot loop side, Linux has a concept of a unique "boot ID" which is exposed by kubelet on the node object, FWIW.

OK, and I think I see, at least *currently* this code is only about bare metal machines, is that correct?

cynepco3hahue:
@

> On the reboot loop side, Linux has a concept of a unique "boot ID" which is exposed by kubelet on the node object, FWIW.
> 
> OK, and I think I see, at least _currently_ this code is only about bare metal machines, is that correct?

Yes, it is correct, to avoid introducing a lot of new code under the machine-api-operator, the MHC will only set the reboot annotation, please see - https://github.com/openshift/machine-api-operator/pull/346#discussion_r301430109

cynepco3hahue:
for the reference - https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_machine-api-operator/346/pull-ci-openshift-machine-api-operator-master-e2e-aws-operator/1125

ingvagabund:
@cgwalters as @cynepco3hahue said this is only to allow better integration of the MHC with berametal without introducing a lot of changes in machine-api-operator before we come up with a though-through proposal. In parallel to baremetal activity, we can discuss generic bits that can be used among many cloud providers. We still need to figure out better mechanism how to capture state of a remedy operation (e.g. CRDs), exercise machine disruption budget, identify weak points, etc. Also, prepare a proposal for upstream community. So even the code in this PR will most likely get re-implemented.

Given our previous discussion, this is ready to go:
- mhc's only responsibility will be to annotate a node (and only for baremetal)
- this change will be transparent for any other cloud provider
- the annotation is considered to be experimental and can change in the future or get completely removed
- it's up to baremetal team to take responsibility for processing nodes with the annotation (including removing the annotation once annotated node is ready again)

/approve

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -327,3 +332,44 @@ func TestUnhealthyForTooLong(t *testing.T) {
 		}
 	}
 }
+
```
enxebre: This should be called `TestRemediation` and baremetal should be test case. That would also force you to have the baremetal domain logic well scoped in its own function even it that produces some duplication

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -352,3 +379,19 @@ func isMaster(machine mapiv1.Machine, client client.Client) bool {
...
+		return nil, err
+	}
+	return infra, nil
+}
+
+func isBareMetalProvider(platform configv1.PlatformType) bool {
```
enxebre: can we have a unit for this?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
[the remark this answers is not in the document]
enxebre: the controller runtime client already use an informer cache underneath, also see here for an example of how to use indexers https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/nodelink/nodelink_controller.go#L141

install/0000_30_machine-api-operator_09_rbac.yaml
```diff
@@ -19,6 +19,14 @@ metadata:
   namespace: openshift-machine-api
 rules:
 
+  - apiGroups:
```
enxebre: does this work for `role` given the infra CRD is cluster scoped? it should be in the machine controllers `ClusterRole`

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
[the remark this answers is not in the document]
enxebre: see https://github.com/openshift/machine-api-operator/pull/346/files#r302447055

enxebre:
@cynepco3hahue @beekhof I dropped a few comments, only blocker to me is https://github.com/openshift/machine-api-operator/pull/346/files#r302447055 which I think might be breaking.
I'm ok merging this under the feature gate tech preview so breaking changes will come up.
Can you please share a doc where people can see the rationale behind this and the expectations for the different scenarios

install/0000_30_machine-api-operator_09_rbac.yaml
cynepco3hahue: I ran tests, and it passed so probable it does not have any problems

enxebre: not sure which tests you mean, but even if it works can you please fix it and put clusterScope resources under the clusterRole? https://kubernetes.io/docs/reference/access-authn-authz/rbac/ thanks

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
cynepco3hahue: I think that BareMetal is the only provider that has different logic on how to remediate machines, so for me, it makes sense to have it under the different test case, also I hope in 4.3 we will have the same logic for all providers, so it is only short term solution

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
cynepco3hahue: it tested as part of all reconcile and remediation tests, so I believe it will be duplication, but if you insist I can create small unittest for it

enxebre:
Eventually this should be platform agnostic. So instead of checking the platform a remediation choice should be configurable so when in baremetal you configure the controller to use `reboot` remediation which just sets the annotation so external opinionated remediation controllers can watch

enxebre:
closing in favour of https://github.com/openshift/machine-api-operator/pull/349 which enables the annotation via reboot strategy

PR #349, opened 2019-07-11
Remediaton strategies

enxebre:
Add `Remediation` field to spec
Defaults to simplest deletion behaviour
Support `remediationReboot` which sets an annotation that consumer remediation systems can react upon
Supersedes https://github.com/openshift/machine-api-operator/pull/346 This allows separation at the right level so consumers can architect their solution by choosing and implementing its strategy, no need for getting infrastructure specifics in the MHC

### DIFF
--- a/pkg/apis/healthchecking/v1alpha1/machinehealthcheck_types.go
+++ b/pkg/apis/healthchecking/v1alpha1/machinehealthcheck_types.go
@@ -7,6 +7,8 @@ import (
 // ConfigMapNodeUnhealthyConditions contains the name of the unhealthy conditions config map
 const ConfigMapNodeUnhealthyConditions = "node-unhealthy-conditions"
 
+type RemediationStrategyType string
+
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
@@ -36,6 +38,9 @@ type MachineHealthCheckList struct {
 
 // MachineHealthCheckSpec defines the desired state of MachineHealthCheck
 type MachineHealthCheckSpec struct {
+	// RemediationStrategy to use in case of problem detection
+	// default is machine deletion
+	RemediationStrategy RemediationStrategyType `json:"remediationStrategy"`
 	// Label selector to match machines whose health will be exercised
 	Selector metav1.LabelSelector `json:"selector"`
 }
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -29,8 +29,10 @@ import (
 )
 
 const (
-	machineAnnotationKey = "machine.openshift.io/machine"
-	ownerControllerKind  = "MachineSet"
+	machineAnnotationKey       = "machine.openshift.io/machine"
+	machineRebootAnnotationKey = "healthchecking.openshift.io/machine-remediation-reboot"
+	ownerControllerKind        = "MachineSet"
+	remediationStrategyReboot  = healthcheckingv1alpha1.RemediationStrategyType("reboot")
 )
 
 // Add creates a new MachineHealthCheck Controller and adds it to the Manager. The Manager will set fields on the Controller
@@ -144,7 +146,7 @@ func (r *ReconcileMachineHealthCheck) Reconcile(request reconcile.Request) (reco
 	for _, hc := range allMachineHealthChecks.Items {
 		if hasMatchingLabels(&hc, machine) {
 			glog.V(4).Infof("Machine %s has a matching machineHealthCheck: %s", machineKey, hc.Name)
-			return remediate(r, machine)
+			return remediate(r, hc.Spec.RemediationStrategy, machine)
 		}
 	}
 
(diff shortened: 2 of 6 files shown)


cynepco3hahue:
I like it in general, but have some concerns:
- it possible that we want to have some complex remediation strategies, like try the first reboot and in case if it fails to delete the machine, in this case, this field can mislead a user(not critical for me)
- we need to reboot master as well in case of bare metal hosts
- lack unittests

enxebre:
> it possible that we want to have some complex remediation strategies, like try the first reboot and in case if it fails to delete the machine, in this case, this field can mislead a user(not critical for me)
That should be handle at the remediation watcher level. We discuss find better naming later on.

> we need to reboot master as well in case of bare metal hosts
As far as I can see this already accounts for it

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -327,3 +327,32 @@ func TestUnhealthyForTooLong(t *testing.T) {
 		}
 	}
 }
+
+func TestApplyRemediationReboot(t *testing.T) {
```
cynepco3hahue: can you just set the type of remediation on the MHC object and run reconcile?

enxebre: that'd be a different unit test, we can add the case to testReconcile though

pkg/apis/healthchecking/v1alpha1/machinehealthcheck_types.go
```diff
@@ -36,6 +36,9 @@ type MachineHealthCheckList struct {
 
 // MachineHealthCheckSpec defines the desired state of MachineHealthCheck
 type MachineHealthCheckSpec struct {
+	// Remediation strategy to use in case of problem detection
+	// default is machine deletion
+	Remediation string `json:"remediation"`
```
bison: Nit: Can we call this `remediatonStrategy`? It fits with things like `updateStrategy` in `Deployment` objects.

bison: Also, maybe this should be a `type RemediationStrategyType string`? That will let us pass strategy type around with some guarantees beyond it being a string.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -165,12 +167,8 @@ func getMachineHealthCheckListOptions() *client.ListOptions {
 	}
 }
 
-func remediate(r *ReconcileMachineHealthCheck, machine *mapiv1.Machine) (reconcile.Result, error) {
+func remediate(r *ReconcileMachineHealthCheck, hc *healthcheckingv1alpha1.MachineHealthCheck, machine *mapiv1.Machine) (reconcile.Result, error) {
```
bison: Do we need the whole `MachineHealthCheck` here? If all we need is the remediation strategy, can we just pass that?

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
enxebre: done

pkg/apis/healthchecking/v1alpha1/machinehealthcheck_types.go
```diff
@@ -36,6 +38,9 @@ type MachineHealthCheckList struct {
 
 // MachineHealthCheckSpec defines the desired state of MachineHealthCheck
 type MachineHealthCheckSpec struct {
+	// RemediationStrategy to use in case of problem detection
+	// default is machine deletion
+	RemediationStrategy RemediationStrategyType `json:"remediationStrategy"`
```
bison: If this is optional and no value means use the default behavior, should this be a pointer with `omitempty`? Maybe it doesn't matter.

Also, if we have the ability, it might be nice to annotate this for validation. Something like:
```
// +kubebuilder:validation:Enum=reboot
```

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -265,6 +283,16 @@ func testReconcile(t *testing.T, remediationWaitTime time.Duration, initObjects
...
 			}
 		}
+		if tc.remediationStrategy == remediationStrategyReboot {
+			node := &corev1.Node{}
+			if err := r.client.Get(context.TODO(), request.NamespacedName, node); err != nil {
+				t.Errorf("Expected: no error, got: %v", err)
+			}
+
+			if _, ok := node.Annotations[machineRebootAnnotationKey]; !ok {
+				t.Errorf("Expected: node to have reboot annotion %s, got: %v", machineRebootAnnotationKey, node.Annotations)
```
vikaschoudhary16: s/annotion/annotation

pkg/apis/healthchecking/v1alpha1/machinehealthcheck_types.go
enxebre: I'll explore the validation, no need to be a pointer since no need to difference between nil, "" and "value". We just have 2 cases ""/"value"

PR #353, opened 2019-07-14
Make remediation strategy optional parameter

cynepco3hahue:
Otherwise, we should pass some dummy value under e2e tests.


enxebre:
can we also add +optional tag? @bison PTAL

cynepco3hahue:
sure will add it

bison:
This looks good, but I think optional values should be pointers: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

I'm not sure how consistently that rule is followed, but in any case I think this should be:
```golang
// +optional
RemediationStrategy *RemediationStrategyType `json:"remediationStrategy,omitempty"`
```

cynepco3hahue:
@bison @enxebre As far as I saw it does not require to be a pointer, but I do not have problems to change it to be a pointer, just give me know what do you prefer?

bison:
I'd prefer it be a pointer. Nil pointer when not provided makes more sense than empty string for this since it's basically an enum.

This is the part in the API conventions doc I was referencing:
> Optional fields have the following properties:
> They have the +optional comment tag in Go.
They are a pointer type in the Go definition (e.g. AwesomeFlag *SomeFlag) or have a built-in nil value (e.g. maps and slices).

pkg/apis/healthchecking/v1alpha1/machinehealthcheck_types.go
```diff
@@ -7,6 +7,7 @@ import (
 // ConfigMapNodeUnhealthyConditions contains the name of the unhealthy conditions config map
 const ConfigMapNodeUnhealthyConditions = "node-unhealthy-conditions"
 
+// RemediationStrategyType contains remediatio strategy type
```
bison: Typo: remediation

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -200,7 +200,7 @@ func remediate(r *ReconcileMachineHealthCheck, remediationStrategy healthcheckin
 		// apply remediation logic, if at least one condition last more than specified timeout
 		// specific remediation logic goes here
 		if unhealthyForTooLong(nodeCondition, conditionTimeout) {
-			switch remediationStrategy {
+			switch *remediationStrategy {
```
bison: This method now needs to check for nil. It should probably do that early on and set a variable that isn't this pointer to a default value in the case of nil, then switch on that.

pkg/apis/healthchecking/v1alpha1/machinehealthcheck_types.go
cynepco3hahue: done

cynepco3hahue:
@bison Thanks for the review

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -200,10 +200,9 @@ func remediate(r *ReconcileMachineHealthCheck, remediationStrategy healthcheckin
 		// apply remediation logic, if at least one condition last more than specified timeout
 		// specific remediation logic goes here
 		if unhealthyForTooLong(nodeCondition, conditionTimeout) {
-			switch remediationStrategy {
-			case remediationStrategyReboot:
+			if remediationStrategy != nil && *remediationStrategy == remediationStrategyReboot {
 				return r.remediationStrategyReboot(machine, node)
-			default:
+			} else {
```
ingvagabund: Nit: no need for the `else` part given `return` in the `if` branch. Though, given the list of strategies might grow over time, it's just a nit.

bison: I agree. I'd actually prefer this remained a `swtich`, but it's not a big deal. We can change this as the list of remediation strategies grows.

PR #354, opened 2019-07-14
Add 'apps' apiGroup to daemonsets RBAC

michaelgugino:
Newer version of k8s libraries are referring to
DaemonSets in apps instead of extensions.

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#deprecations

```
DaemonSet, Deployment, and ReplicaSet resources will
no longer be served from extensions/v1beta1, apps/v1beta1,
or apps/v1beta2 in v1.16. Migrate to the apps/v1 API,
available since v1.9. Existing persisted data can be
retrieved via the apps/v1 API.
```

### DIFF
--- a/config/machine-api-operator-patch.yaml
+++ b/config/machine-api-operator-patch.yaml
@@ -157,6 +157,7 @@ rules:
 
   - apiGroups:
       - extensions
+      - apps
     resources:
       - daemonsets
     verbs:


frobware:
/lgtm

Please could you update the PR description to indicate which version of k8s libraries forced our hand. TIA.

PR #356, opened 2019-07-18
Increase run-level  of  monitoring objects, better metric naming and new alert rule

vikaschoudhary16:
To make sure that service monitor and another monitoring resources gets created after openshift monitoring(prometheus) has got started, run-level of all monitoring resources is being increased to 90

reference: https://github.com/openshift/cluster-kube-controller-manager-operator/tree/master/manifests


pkg/metrics/metrics.go
```diff
@@ -70,7 +70,7 @@ func (mc MachineCollector) Describe(ch chan<- *prometheus.Desc) {
 func (mc MachineCollector) collectMachineMetrics(ch chan<- prometheus.Metric) {
 	machineList, err := mc.listMachines()
 	if err != nil {
-		ScrapeFailedCounter.With(prometheus.Labels{"kind": "Machine-count", "reason": err.Error()}).Inc()
+		ScrapeFailedCounter.With(prometheus.Labels{"kind": "Machine-count"}).Inc()
```
simonpasquier: On second thought, this metric should be a [gauge](https://github.com/openshift/machine-api-operator/pull/335#discussion_r304289521) rather than a counter. See https://prometheus.io/docs/instrumenting/writing_exporters/#metrics-about-the-scrape-itself.

install/0000_30_machine-api-operator_09_rbac.yaml
```diff
@@ -276,37 +276,3 @@ subjects:
   - kind: ServiceAccount
     name: machine-api-operator
     namespace: openshift-machine-api
-
----
```
simonpasquier: pinging @sur to acknowledge that this is the correct move.

pkg/metrics/metrics.go
vikaschoudhary16: @simonpasquier 
> These should be exposed as gauges as they’re about an event

Looks like guage is suggested for the metric which are for specific scrape and since this metrics is not ablut any specific scrape rather reporting the overall status till any moment, do you think Guwague makes sense? i am not sure.

simonpasquier: It is still relevant IMO. This code runs in some background loop, it is only executed when Prometheus (or something else) scrapes the metrics endpoint.

pkg/metrics/metrics.go
```diff
@@ -11,19 +11,19 @@ import (
```
enxebre: can you please reword this commit to the specific corrections is doing and follow imperative style https://chris.beams.io/posts/git-commit/?

pkg/metrics/metrics.go
vikaschoudhary16: agree that it runs when prometheus (or something else) scrapes the metrics endpoint but the metric value is not "about this particular scrape(event)", rather it is cumulative info about all such events that happened till any moment.

simonpasquier: Think about alerting on this metric: you're interesting to know if several scrapes fail in a row (as in general your other alerts should be resilient to occasional scrape failures).
With a gauge, the alert is `mapi_collector_up == 0` combined with a `for` condition equal to a few times the scrape interval.
With a counter you would have to expose `mapi_scrape_failures_total` and `mapi_scrapes_total` then query `rate(mapi_scrape_failures_total[5m]) / rate(mapi_scrapes_total[5m] > x` with `x` being harder to determine. And this alert may not resolved immediately when the scrapes are successful again.

install/0000_30_machine-api-operator_09_rbac.yaml
enxebre: I don't think there's a need to move this roles

pkg/metrics/metrics.go
vikaschoudhary16: @simonpasquier thanks! make sense!

install/0000_90_machine-api-operator_04_alertrules.yaml
```diff
@@ -18,3 +18,25 @@ spec:
             severity: critical
           annotations:
             message: "machine {{ $labels.name }} does not have valid node reference"
+
```
enxebre: can you move this to different commit/PR

pkg/metrics/metrics.go
```diff
@@ -11,32 +11,32 @@ import (
 
 var (
 	// MachineCountDesc is a metric about machine object count in the cluster
-	MachineCountDesc = prometheus.NewDesc("mapi_machine_items_count", "Count of machine objects currently at the apiserver", nil, nil)
+	MachineCountDesc = prometheus.NewDesc("mapi_machine_items", "Count of machine objects currently at the apiserver", nil, nil)
```
enxebre: This does nothing to do with PR title

PR #358, opened 2019-07-22
Decrement MD allowed count only when machine has not be previously accounted

ingvagabund:
SSIA

TODO:
- [x] unit test call to `RetryDecrementMachineDisruptionsAllowed`

### DIFF
--- a/pkg/controller/disruption/disruption.go
+++ b/pkg/controller/disruption/disruption.go
@@ -482,10 +482,16 @@ func isMachineDisruptionAllowed(mdb *healthcheckingv1alpha1.MachineDisruptionBud
 }
 
 func decrementMachineDisruptionsAllowed(c client.Client, machineName string, mdb *healthcheckingv1alpha1.MachineDisruptionBudget) error {
-	mdb.Status.MachineDisruptionsAllowed--
 	if mdb.Status.DisruptedMachines == nil {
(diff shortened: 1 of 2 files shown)


bison:
This looks good, though I think we also need to talk about how `buildDisruptedMachineMap()` works. Doesn't seem to really fit now that there are remediation strategies other than `delete`.

cynepco3hahue:
> This looks good, though I think we also need to talk about how `buildDisruptedMachineMap()` works. Doesn't seem to really fit now that there are remediation strategies other than `delete`.

One who adds machines to the `DisruptedMachines` is MHC controller once the remediation was initiated, so the logic will be the same, but we need to rename all delete word to remediate.

ingvagabund:
> One who adds machines to the DisruptedMachines is MHC controller once the remediation was initiated, so the logic will be the same, but we need to rename all delete word to remediate.

Renaming is not sufficient. `buildDisruptedMachineMap` hardcodes `DeletionTimeout` const (set to 2 minutes) and assumes all re-mediated machines are to be deleted eventually.

cynepco3hahue:
> Renaming is not sufficient. `buildDisruptedMachineMap` hardcodes `DeletionTimeout` const (set to 2 minutes) and assumes all re-mediated machines are to be deleted eventually.

True :worried: If we will move to work with `MachineRemediation` CR, we can poll the state of the remediation process.

PR #359, opened 2019-07-22
WIP: Remove configmap in favor of crd

ingvagabund:
In practice, each machine pool will have its own list of conditions under which a node is considered unhealthy. Thus, making general config map based configuration obsolete. Also, CRD allows to have
- versioned configuration
- server side validation
- each machine pool configuration self-contained


cynepco3hahue:
@ingvagabund MDB should use the same conditions as MHC to decide if the node is healthy or not, so I do not sure that I want to have it under MHC CR, another point, I still want to have the default behavior when a user did not specify conditions at all.

ingvagabund:
> @ingvagabund MDB should use the same conditions as MHC to decide if the node is healthy or not, so I do not sure that I want to have it under MHC CR,

I disagree. MDB has no business of knowing how MHC is working. MHC machine pools are independent of MDB and visa verse. We need to figure out different way how to decide when a machine is no longer disrupted.

> another point, I still want to have the default behavior when a user did not specify conditions at all.

I have nothing against that. We can still fallback to the default conditions. We just need to make sure all the conditions are reasonable as default.

enxebre:
Closing as this is already in master

PR #360, opened 2019-07-23
Get values for CRD's additionalPrinterColumns from annotations and labels instead of machine provider config

ingvagabund:
Given we already support AWS, Azure and GCP instalations of OpenShift,
reading values from machine provider config is no longer an option.
Instead, let's read all needed values from machine annotations and labels.
Each actuator will be then responsible for making sure every machine
has all the annotations and labels that are needed to properly display
additional machine columns set.

- [x] Azure: https://github.com/openshift/cluster-api-provider-azure/pull/59
- [x] AWS: https://github.com/openshift/cluster-api-provider-aws/pull/242
- [ ] GCP: TBD
- [ ] OpenStack: TBD

### DIFF
--- a/install/0000_30_machine-api-operator_02_machine.crd.yaml
+++ b/install/0000_30_machine-api-operator_02_machine.crd.yaml
@@ -7,23 +7,19 @@ metadata:
   name: machines.machine.openshift.io
 spec:
   additionalPrinterColumns:
-  - JSONPath: .status.providerStatus.instanceId
-    description: Instance ID of machine created in AWS
-    name: Instance
-    type: string
-  - JSONPath: .status.providerStatus.instanceState
-    description: State of the AWS instance
+  - JSONPath: .metadata.annotations['machine\.openshift\.io/instance-state']
+    description: State of instance
     name: State
     type: string
-  - JSONPath: .spec.providerSpec.value.instanceType
+  - JSONPath: .metadata.labels['machine\.openshift\.io/instance-type']
     description: Type of instance
     name: Type
     type: string
-  - JSONPath: .spec.providerSpec.value.placement.region
+  - JSONPath: .metadata.labels['machine\.openshift\.io/region']
     description: Region associated with machine
     name: Region
     type: string
-  - JSONPath: .spec.providerSpec.value.placement.availabilityZone
+  - JSONPath: .metadata.labels['machine\.openshift\.io/zone']
     description: Zone associated with machine
     name: Zone
     type: string
@@ -31,6 +27,16 @@ spec:
     description: Machine age
     name: Age
     type: date
+  - JSONPath: .status.nodeRef.name
+    description: Node associated with machine
+    name: Node
+    priority: 1
+    type: string
+  - JSONPath: .spec.providerID
+    description: Provider ID of machine created in cloud provider
+    name: ProviderID
+    priority: 1
+    type: string
   group: machine.openshift.io
   names:
     kind: Machine
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -384,7 +384,7 @@
 
 [[projects]]
   branch = "openshift-4.2-cluster-api-0.1.0"
-  digest = "1:1fd0478928373af6be747a9191b2f60179ded98f612d33200fbd7521aea5f4df"
+  digest = "1:2ccd67b04d2483877ff0c4bdf59a3e7124de72dde98a7391bdfea314f420316f"
   name = "github.com/openshift/cluster-api"
   packages = [
     "pkg/apis/cluster/common",
@@ -405,7 +405,7 @@
     "pkg/client/listers_generated/machine/v1beta1",
   ]
   pruneopts = "NUT"
-  revision = "e911af77e9a33cfb45f82ba9778876d715cdb4ea"
+  revision = "f8de78af80fcecf1fd69e35005b591dc9e1df61a"
 
 [[projects]]
   branch = "master"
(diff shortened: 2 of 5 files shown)


install/0000_30_machine-api-operator_02_machine.crd.yaml
```diff
@@ -7,23 +7,23 @@ metadata:
   name: machines.machine.openshift.io
 spec:
   additionalPrinterColumns:
-  - JSONPath: .status.providerStatus.instanceId
+  - JSONPath: .metadata.annotations['machine\.openshift\.io/instance-id']
     description: Instance ID of machine created in AWS
```
Fedosin: I think we also need to remove all references to AWS

install/0000_30_machine-api-operator_02_machine.crd.yaml
```diff
@@ -7,23 +7,23 @@ metadata:
...
     name: Instance
     type: string
-  - JSONPath: .status.providerStatus.instanceState
+  - JSONPath: .metadata.annotations['machine\.openshift\.io/instance-state']
     description: State of the AWS instance
```
Fedosin: ditto

Fedosin:
@ingvagabund could you also add OpenStack to the list of platforms in the PR description?

install/0000_30_machine-api-operator_02_machine.crd.yaml
ingvagabund: Thanks, fixed.

ingvagabund:
@Fedosin thanks for the review. Once Azure PR is merged (where we are deciding which annotations/labels we want to use), we can distribute the same change to other actuators.

ingvagabund:
Updating machine and machineset types in openshift/cluster-api first: https://github.com/openshift/cluster-api/pull/55

install/0000_30_machine-api-operator_02_machine.crd.yaml
```diff
@@ -7,23 +7,23 @@ metadata:
   name: machines.machine.openshift.io
 spec:
   additionalPrinterColumns:
-  - JSONPath: .status.providerStatus.instanceId
-    description: Instance ID of machine created in AWS
+  - JSONPath: .metadata.annotations['machine\.openshift\.io/instance-id']
```
enxebre: can we just get this from providerID?

ingvagabund: `ProviderID` is cloud provider prefixed, e.g. `aws://...`. My usual use case was to just copy paste the ID and use it to search instance in AWS portal. With `aws://` I would always have to remove the prefix. Also, for Azure it has no value as I don't think you can search Azure vm based on its ID. I tried and I failed. What would be the benefit of reading ProviderID?

enxebre: `providerID` is the machine API source of truth. It's build on top of the `instance-id` that you can just pick from it, hence no need nor value for an specific annotation

ingvagabund: I don't mind reading ProviderID instead. Though, the question is what's the user expectation of `InstanceID` field. Though, it does not make sense to name the column `Instance` anymore.

Also, as I asked in https://github.com/openshift/machine-api-operator/pull/360/files#r308229498, what would be the `providerID` value used for?

ingvagabund: We can put the providerID under `-o wide` [1] so it's not printed by default. Right now, it has value only for AWS.

[1] https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#priority

enxebre: FYI providerID is also in azure and must be in gcp.
ok cool I'm good with `instance-id'` please make sure to add that in azure https://github.com/openshift/cluster-api-provider-azure/commit/747894809218b1826aa4a9f454c25708c92e5052#diff-7c34c0cb525c96c7b732376b0c8994e1R56

ingvagabund:
/hold

Just validated OpenShift in both AWS and Azure properly display all relevant columns when running `oc get machines -n openshift-machine-api`.

install/0000_30_machine-api-operator_02_machine.crd.yaml
```diff
@@ -7,30 +7,36 @@ metadata:
...
   - JSONPath: .metadata.creationTimestamp
     description: Machine age
     name: Age
     type: date
+  - JSONPath: .status.nodeRef.name
+    description: Node associated with machine
+    name: Node
+    priority: 1
+    type: string
+  - JSONPath: .spec.providerID
```
abhinavdahiya: Should this be from the status??

ingvagabund: ProviderID is spec field and there was a long discussion which of spec or status is the right place to put it. Spec won.

Fedosin:
@ingvagabund hi! add the OpenStack PR please https://github.com/openshift/cluster-api-provider-openstack/pull/55

ingvagabund:
@Fedosin there is no need to update machine CRD anymore. The CRD expects certain annotations and labels exists. If they don't, corresponding column is just empty. So, only step needed to is update an actuator to set the requires annotations/labels.

Fedosin:
@ingvagabund correct. that's what I did in openshift/cluster-api-provider-openstack#55

PR #362, opened 2019-07-23
credentials request: fix gcp roles

csrwng:
The cloud credentials operator requires that roles be specified with a
prefix indicating their scope (roles/* for global roles,
roles/project/[project]/* for project-scoped roles, etc)


csrwng:
@enxebre ptal

PR #363, opened 2019-07-26
Allow reading Infrastructures from CAP*

tomassedovic:
This lets the cluster-api-provider-* services read the Infrastructures
object under the `config.openshift.io` API group.

cluster-api-provider-openstack uses it to get the VIP addresses it needs
to configure on the OpenStack ports it creates.

### DIFF
--- a/install/0000_30_machine-api-operator_09_rbac.yaml
+++ b/install/0000_30_machine-api-operator_09_rbac.yaml
@@ -133,6 +133,15 @@ rules:
       - list
       - watch
 
+  - apiGroups:
+    - config.openshift.io
+    resources:
+    - infrastructures
+    verbs:
+    - get
+    - list
+    - watch
+
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role


tomassedovic:
I'm not sure whether this is the right way of doing this. We've got some values in Infrastructure.PlatformStatus.OpenStack that we need to pass to cluster-api-provider-openstack because the cloud resources (the OpenStack Port specifically) needs them:

https://github.com/openshift/cluster-api-provider-openstack/pull/52

There are other ways we could achieve this, but this is the simplest and cleanest from our point of view. However it does mean that CAPO needs to be able to read this object.

celebdor:
> I'm not sure whether this is the right way of doing this. We've got some values in Infrastructure.PlatformStatus.OpenStack that we need to pass to cluster-api-provider-openstack because the cloud resources (the OpenStack Port specifically) needs them:
> 
> [openshift/cluster-api-provider-openstack#52](https://github.com/openshift/cluster-api-provider-openstack/pull/52)
> 
> There are other ways we could achieve this, but this is the simplest and cleanest from our point of view. However it does mean that CAPO needs to be able to read this object.

I think it is perfectly legitimate for a machine controller to modify its behavior in accordance to the reported PlatformStatus that it is enabling.

ingvagabund:
MAO is still owner of RBAC rules for all actuators. I don't think there is a better place for this. In the future we might break the RBAC rules per an actuator (for cost of adding some duplicity).

/approve

tomassedovic:
Thanks!

PR #365, opened 2019-07-29
Add node to machine additionalPrinterColumns

enxebre:
Add node to machine additionalPrinterColumns

### DIFF
--- a/install/0000_30_machine-api-operator_02_machine.crd.yaml
+++ b/install/0000_30_machine-api-operator_02_machine.crd.yaml
@@ -27,6 +27,10 @@ spec:
     description: Zone associated with machine
     name: Zone
     type: string
+  - JSONPath: .status.nodeRef.name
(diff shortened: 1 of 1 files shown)


install/0000_30_machine-api-operator_02_machine.crd.yaml
```diff
@@ -27,6 +27,10 @@ spec:
     description: Zone associated with machine
     name: Zone
     type: string
+  - JSONPath: .status.nodeRef.name
```
ingvagabund: It needs to be set in `machine_types.go` first and the CRD updated by running `make gen-crd`. Otherwise, your  change will get removed after the CRD gets regen again. Let's collect all the bits in https://github.com/openshift/cluster-api/pull/55 first.

ingvagabund:
Part of https://github.com/openshift/machine-api-operator/pull/360

PR #368, opened 2019-08-03
Expose metric endpoint on https

vikaschoudhary16:
This PR is adding changes to expose metrics endpoint on https and also changes at other objects for integration with openshift-monitoring at https.
This PR is also restricting http port access only on loopback ip.

## Approach for https support:
[kube-rbac-proxy](https://github.com/openshift/kube-rbac-proxy/tree/master/examples/rewrites#rewriting-subjectaccessreviews-example) container is being run alongside operator container which runs metrics server in the same pod. kube-rbac -proxy validates authz and authn on behalf of the metrics server and if success, proxies the incoming https request to the metrics server(which is listening on 127.0.0.1 and http port 8080).
### How authn and authz is being validated by kube-rbac-proxy
It uses `TokenReviews()` and `SubjectAccessReviews()` apis internally for validation of token authentication and authorization respectively. These apis are used against a stub resource, `namespace/metrics` which is passed as config to the kube-rbac-proxy
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kube-rbac-proxy
data:
  config-file.yaml: |+
    authorization:
      resourceAttributes:
        apiVersion: v1
        resource: namespace
        subresource: metrics
 ```
[RBAC permissions](https://github.com/openshift/machine-api-operator/pull/368/commits/cfacd9dc1d5666dd1958dcd3fa6f32c78fd7d7e1#diff-2043d8125934ffff48ca0cb4d630543bR329-R334) for this rule are defined same as that we expect for metrics endpoint.
So when a https request is received for metrics endpoint, token and rbac of the request user is validated against this stub resource. If it passes, proxied to the metrics endpoint.

 ### How I tested it locally
- launch a cluster using installer
- make cluster-version deployment replica to 0
- change mao image with the one with the PR changes
- change/create all the api objects(like clusterrole, rbac-config etc) that are part of this PR
- I had to make a curl command on the metrics service with service account, `prometheus-k8s`. I got this token using following command:
> kubectl -n openshift-monitoring get secret $(kubectl -n openshift-monitoring get sa prometheus-k8s -o 'jsonpath={.secrets[0].name}') -o 'jsonpath={.data.token}' | base64 -d

- Next, created a pod in the openshift-monitoring namespace to make curl call at service
- Then copy token obtained above in the curl container and make the curl call
> curl -v -s -k -H "Authorization: Bearer `cat ./token2`" https://machine-api-operator.openshift-machine-api.svc:8443/metrics

This should return metrics

### DIFF
--- a/config/machine-api-operator-deployment.yaml
+++ b/config/machine-api-operator-deployment.yaml
@@ -1,3 +1,4 @@
+---
 apiVersion: apps/v1beta2
 kind: Deployment
 metadata:
@@ -23,8 +24,6 @@ spec:
         args:
         - "--secure-listen-address=0.0.0.0:8443"
         - "--upstream=http://localhost:8080/"
-        - "--tls-cert-file=/etc/tls/private/tls.crt"
-        - "--tls-private-key-file=/etc/tls/private/tls.key"
         - "--config-file=/etc/kube-rbac-proxy/config-file.yaml"
         - "--logtostderr=true"
         - "--v=10"
@@ -34,8 +33,6 @@ spec:
         volumeMounts:
         - name: config
           mountPath: /etc/kube-rbac-proxy
-        - mountPath: /etc/tls/private
-          name: machine-api-operator-tls
       - name: machine-api-operator
         image: docker.io/openshift/origin-machine-api-operator:v4.0.0
         command:
@@ -85,7 +82,4 @@ spec:
           name: kube-rbac-proxy
       - name: images
         configMap:
-                name: machine-api-operator-images
-      - name: machine-api-operator-tls
-        secret:
-          secretName: machine-api-operator-tls
+          name: machine-api-operator-images
--- a/cmd/machine-api-operator/start.go
+++ b/cmd/machine-api-operator/start.go
@@ -145,7 +145,7 @@ func startMetricsCollectionAndServer(ctx *ControllerContext) {
 		metricsPort = v
 	}
 	glog.V(4).Info("Starting server to serve prometheus metrics")
-	go startHTTPMetricServer(fmt.Sprintf(":%d", metricsPort))
+	go startHTTPMetricServer(fmt.Sprintf("localhost:%d", metricsPort))
 }
 
 func startHTTPMetricServer(metricsPort string) {
--- a/install/0000_30_machine-api-operator_09_rbac.yaml
+++ b/install/0000_30_machine-api-operator_09_rbac.yaml
@@ -200,7 +200,14 @@ kind: ClusterRole
 metadata:
   name: machine-api-operator
 rules:
-
+  - apiGroups: ["authentication.k8s.io"]
+    resources:
+      - tokenreviews
+    verbs: ["create"]
+  - apiGroups: ["authorization.k8s.io"]
+    resources:
+      - subjectaccessreviews
+    verbs: ["create"]
   - apiGroups:
       - config.openshift.io
     resources:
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

File: config/machine-api-operator-deployment.yaml [hunk 1]
```diff
@@ -0,0 +1,85 @@
...
+        - "--logtostderr=true"
+        - "--v=10"
+        ports:
+        - containerPort: 8443
+          name: https
+        volumeMounts:
+        - name: config
+          mountPath: /etc/kube-rbac-proxy
+      - name: machine-api-operator
+        image: docker.io/openshift/origin-machine-api-operator:v4.0.0
```
frobware: This image refers to v4.0 whereas the kube-rbac-proxy refers to 4.2.0. Is this intentional?

vikaschoudhary16: mao image version is same as it is today in https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_11_deployment.yaml#L22

kube-rbac-proxy image 4.2.0 is taken just to be latest. Not sure if these two should be same. I see both image streams 4.0.0 and 4.2.0 in the install/image-references, so chose to follow the same pattern and use 4.2.0

frobware:
Do we care that we listen on IPv4 only? Why not explicitly use `localhost` which would cover IPv6 too?

config/machine-api-operator-deployment.yaml
```diff
@@ -0,0 +1,85 @@
+---
+apiVersion: apps/v1beta2
+kind: Deployment
+metadata:
+  name: machine-api-operator
+  namespace: openshift-machine-api
```
enxebre: can you elaborate why we need this file?

vikaschoudhary16: This is required for the k8s-e2e ci job which runs test against the k8s. In the original deployment, we are mounting a secret into the pod. Based on the annotation in the mao metrics service object, this [`secret`](https://github.com/openshift/machine-api-operator/pull/368/files/5d2b38bc7a0e838f2032a29c6efb7c966c256912#diff-273b08d16d3abfbf3bc5df0438e0c450R91) will get created automatically in the openshift cluster.  With the minikube cluster, this secret will not be there and thus deployment will fail. To workaround this, in the k8s-e2e tests this deployment will be used which is not mounting secrets.

enxebre:
/approve
@openshift/openshift-team-monitoring could anyone have a quick look? cc @simonpasquier @s-urbaniak @paulfantom

config/machine-api-operator-deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+  template:
+    metadata:
+      labels:
+        k8s-app: machine-api-operator
+    spec:
+      priorityClassName: system-node-critical
+      serviceAccountName: machine-api-operator
+      containers:
+      - name: kube-rbac-proxy
+        image: quay.io/openshift/origin-kube-rbac-proxy:4.2.0
```
brancz: this should be an image stream no?

brancz: nevermind, it is

config/machine-api-operator-deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+        args:
+        - "--secure-listen-address=0.0.0.0:8443"
```
brancz: this will use an at startup generated self-signed cert, is that intended?

vikaschoudhary16: yes, and this is re;ated, https://github.com/openshift/machine-api-operator/pull/368/files/5d2b38bc7a0e838f2032a29c6efb7c966c256912#diff-56c9657f199776b5f04119428ca26c17R8

PR #369, opened 2019-08-05
Bug 1727021: Set Upgradeable condition on ClusterOperator status

enxebre:
This is to be compliant with https://github.com/openshift/cluster-version-operator/blob/b57ee63baf65f7cb6e95a8b2b304d88629cfe3c0/docs/dev/clusteroperator.md#what-should-an-operator-report-with-clusteroperator-custom-resource
When known hazardous states for upgrades are determined, specific "Upgradeable=False" status can be added with messages for how admins can resolve it.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1727021

### DIFF
--- a/pkg/operator/status_test.go
+++ b/pkg/operator/status_test.go
@@ -7,6 +7,7 @@ import (
 
 	osconfigv1 "github.com/openshift/api/config/v1"
 	fakeconfigclientset "github.com/openshift/client-go/config/clientset/versioned/fake"
+	cvoresourcemerge "github.com/openshift/cluster-version-operator/lib/resourcemerge"
 	"github.com/stretchr/testify/assert"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/record"
@@ -36,7 +37,7 @@ func TestOperatorStatusProgressing(t *testing.T) {
 	type tCase struct {
 		currentVersion        []osconfigv1.OperandVersion
 		desiredVersion        []osconfigv1.OperandVersion
-		expectedStatus        osconfigv1.ConditionStatus
+		expectedConditions    []osconfigv1.ClusterOperatorStatusCondition
 		transitionTimeUpdated bool
 	}
 	tCases := []tCase{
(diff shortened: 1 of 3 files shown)


pkg/operator/status_test.go
```diff
@@ -81,27 +92,44 @@ func TestOperatorStatusProgressing(t *testing.T) {
 		optr.osClient = fakeconfigclientset.NewSimpleClientset(co)
 		startTime := metav1.Now()
 		optr.statusProgressing()
-		o, _ := optr.osClient.ConfigV1().ClusterOperators().Get(clusterOperatorName, metav1.GetOptions{})
+		got, _ := optr.osClient.ConfigV1().ClusterOperators().Get(clusterOperatorName, metav1.GetOptions{})
```
ingvagabund: Why is it `got` now?

ingvagabund:
Is it safe to assume `Upgradeable=True` by default as we did in CAO?

pkg/operator/status_test.go
vikaschoudhary16: `got` does not seem very intuitive. may be `co`?

enxebre: I usually use `got` vs `expected`, I don't mind, if you find this a blocker for the PR please Iet me know so I'll put it back or whatever the majority prefer

enxebre:
>Is it safe to assume Upgradeable=True by default as we did in CAO?

That's the assumption here, remain current functionality and make the field explicit.
I'd anticipate that as the system evolve we'll determine known hazardous states for upgrades where specific "Upgradeable=False" status can be added with messages for how admins
can resolve it.

pkg/operator/status_test.go
```diff
@@ -81,27 +92,44 @@ func TestOperatorStatusProgressing(t *testing.T) {
...
 			}
 		}
-		assert.Equal(t, tc.expectedStatus, condition.Status, "test-case %v expected condition %v to be %v, but got %v", i, condition.Type, tc.expectedStatus, condition.Status)
+
+		for _, expectedCondition := range tc.expectedConditions {
+			ok := cvoresourcemerge.IsOperatorStatusConditionPresentAndEqual(
+				got.Status.Conditions, expectedCondition.Type, expectedCondition.Status,
+			)
+			if !ok {
+				t.Errorf("wrong status for condition: %s", expectedCondition.Type)
```
frobware: Can we express this with both expected and got?

pkg/operator/status_test.go
vikaschoudhary16: in that case may be `gotCo`, just like "Condition" in `expectedCondition`.
not a blocker though.

PR #370, opened 2019-08-05
Change image reference name for ironic

juliakreger:


install/image-references
```diff
@@ -34,11 +34,11 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-baremetal-operator:v4.2.0
-  - name: ironic-image
+  - name: ironic
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-ironic-image:v4.2.0
```
juliakreger: This and the similar line below likely needs to change.

PR #371, opened 2019-08-07
Bug 1704587: Reconcile nodes on MHC changes

enxebre:
Watch MachineHealthChecks and enqueue reconcile.Request for the backed nodes.
This is useful to trigger remediation when e.g a machineHealCheck is created against
a node which is already unhealthy and is not able to receive status updates which would otherwise be ignored by MHC
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1704587

### DIFF
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -3,6 +3,7 @@ package machinehealthcheck
 import (
 	"context"
 	golangerrors "errors"
+	"fmt"
 	"time"
 
 	"github.com/golang/glog"
@@ -37,11 +38,12 @@ const (
 // Add creates a new MachineHealthCheck Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and start it when the Manager is started.
 func Add(mgr manager.Manager, opts manager.Options) error {
-	return add(mgr, newReconciler(mgr, opts))
+	r := newReconciler(mgr, opts)
+	return add(mgr, r, r.nodeRequestsFromMachineHealthCheck)
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager, opts manager.Options) reconcile.Reconciler {
+func newReconciler(mgr manager.Manager, opts manager.Options) *ReconcileMachineHealthCheck {
 	return &ReconcileMachineHealthCheck{
 		client:    mgr.GetClient(),
 		scheme:    mgr.GetScheme(),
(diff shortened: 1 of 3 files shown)


pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -142,6 +153,78 @@ func (r *ReconcileMachineHealthCheck) Reconcile(request reconcile.Request) (reco
...
+		return requests
+	}
+	return []reconcile.Request{}
+}
+
+func (r *ReconcileMachineHealthCheck) getNodeNamesForMHC(mhc healthcheckingv1alpha1.MachineHealthCheck) ([]types.NodeName, error) {
+	machineList := &mapiv1.MachineList{}
+	selector, err := metav1.LabelSelectorAsSelector(&mhc.Spec.Selector)
+	if err != nil {
+		fmt.Errorf("failed to build selector")
```
frobware: `return err`?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -142,6 +153,78 @@ func (r *ReconcileMachineHealthCheck) Reconcile(request reconcile.Request) (reco
...
+		machineList,
+		client.UseListOptions(options.InNamespace(mhc.GetNamespace()))); err != nil {
+		return nil, fmt.Errorf("failed to list machines: %v", err)
+	}
+
+	if len(machineList.Items) < 1 {
+		return nil, nil
+	}
+
+	nodeNames := []types.NodeName{}
```
frobware: I think this cold be simplified to:

```go
var nodeNames []types.NodeName
for loop {
	// as-is
}
return nodeNames, nil
```

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -142,6 +153,78 @@ func (r *ReconcileMachineHealthCheck) Reconcile(request reconcile.Request) (reco
...
+	if err := r.client.Get(
+		context.Background(),
+		client.ObjectKey{
+			Namespace: o.Meta.GetNamespace(),
+			Name:      o.Meta.GetName(),
+		},
+		mhc,
+	); err != nil {
+		glog.Errorf("No-op: Unable to retrieve mhc %s/%s from store: %v", o.Meta.GetNamespace(), o.Meta.GetName(), err)
+		return []reconcile.Request{}
```
frobware: This wants `[]reconcile.Request{}` as opposed to just `nil`?

enxebre: yeh that's intentional as it's the expectation by controller runtime internals

PR #374, opened 2019-08-09
baremetal: Fix service account name.

russellb:
The "metal3-controller" service account does not exist.  Instead, use
the "machine-api-controllers" service account.


eparis:
when do I get CI so this doesn't show up to start with?

russellb:
> when do I get CI so this doesn't show up to start with?

once we can use openshift release images instead of building custom ones, and then we can build some proper CI integration with OpenShift.

getting an installer build into the release is the last thing

PR #382, opened 2019-08-15
baremetal: Adjust container order for baremetal pod

hardys:
Aligning with https://github.com/openshift-metal3/dev-scripts/pull/737
we need to ensure the static-ip-manager container is started first,
and the baremetal-operator pod last due to depenencies between the
containers.

### DIFF
--- a/pkg/operator/baremetal_pod.go
+++ b/pkg/operator/baremetal_pod.go
@@ -186,54 +186,53 @@ func createInitContainerStaticIpSet(config *OperatorConfig) corev1.Container {
 }
 
 func newMetal3Containers(config *OperatorConfig) []corev1.Container {
-	//Starting off with the metal3-baremetal-operator container
-	containers := []corev1.Container{
-		{
-			Name:  "metal3-baremetal-operator",
-			Image: config.BaremetalControllers.BaremetalOperator,
-			Ports: []corev1.ContainerPort{
-				{
-					Name:          "metrics",
-					ContainerPort: 60000,
-				},
-			},
-			Command:         []string{"/baremetal-operator"},
-			ImagePullPolicy: "Always",
-			Env: []corev1.EnvVar{
-				{
-					Name: "WATCH_NAMESPACE",
-					ValueFrom: &corev1.EnvVarSource{
-						FieldRef: &corev1.ObjectFieldSelector{
-							FieldPath: "metadata.namespace",
-						},
-					},
-				},
-				{
-					Name: "POD_NAME",
-					ValueFrom: &corev1.EnvVarSource{
-						FieldRef: &corev1.ObjectFieldSelector{
-							FieldPath: "metadata.name",
-						},
-					},
-				},
-				{
-					Name:  "OPERATOR_NAME",
-					Value: "baremetal-operator",
-				},
-				setEnvVar("DEPLOY_KERNEL_URL", "deploy_kernel_url"),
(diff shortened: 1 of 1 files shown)


hardys:
cc @derekhiggins @russellb

russellb:
@derekhiggins @hardys Are you sure this has an effect?  From what I'm reading, initContainers are ordered, but the application containers are not.  I found this feature request to add support for container ordering (among other ideas) here: https://github.com/kubernetes/kubernetes/issues/65502

The easiest alternative seems to be building your own dependency checking into startup scripts in each container, as needed.

hardys:
> @derekhiggins @hardys Are you sure this has an effect? From what I'm reading, initContainers are ordered, but the application containers are not. I found this feature request to add support for container ordering (among other ideas) here: [kubernetes/kubernetes#65502](https://github.com/kubernetes/kubernetes/issues/65502)

Hmm, in the yaml testing done my @derekhiggins I thought we saw changed behaviour with reordering, but I'm not certain and am happy to consider a different approach.

> The easiest alternative seems to be building your own dependency checking into startup scripts in each container, as needed.

Yeah perhaps it's better if we make each container with a dependency on the provisioning interface just spin waiting if it's not up, then the ordering doesn't matter for static-ip-refresh.

The issue with the BMO was different IIRC, which was that it tried to register/adopt hosts with ironic whilst it was still coming up - @derekhiggins were those errors transient, e.g can we ignore them and stick with the current order if we fix the static-ip-refresh dependency that causes some of the other containers to exit?

derekhiggins:
> > @derekhiggins @hardys Are you sure this has an effect? From what I'm reading, initContainers are ordered, but the application containers are not. I found this feature request to add support for container ordering (among other ideas) here: [kubernetes/kubernetes#65502](https://github.com/kubernetes/kubernetes/issues/65502)
> 
> Hmm, in the yaml testing done my @derekhiggins I thought we saw changed behaviour with reordering, but I'm not certain and am happy to consider a different approach.

It may not be explicitly documented but the container start order definitely matched the order we had in the yaml in both cases (before and after I reordered it). 

The kubernetes issue you linked is concerned with holding off on starting containers until it dependencies are ready (which could be delayed after the container startup), so it goes further then what we are trying to achieve. 

> 
> > The easiest alternative seems to be building your own dependency checking into startup scripts in each container, as needed.
> 
> Yeah perhaps it's better if we make each container with a dependency on the provisioning interface just spin waiting if it's not up, then the ordering doesn't matter for static-ip-refresh.

Yes this would be a good check to put into them but we should also put static-ip-refresh first 

> 
> The issue with the BMO was different IIRC, which was that it tried to register/adopt hosts with ironic whilst it was still coming up - @derekhiggins were those errors transient, e.g can we ignore them and stick with the current order if we fix the static-ip-refresh dependency that causes some of the other containers to exit?

I thought they were the cause of the failure I was investigating but it now looks like there were other problems so I don't think we need to change the order, but we should to avoid confusing errors in logs.

hardys:
Ok so it sounds like we should probably go ahead with this reordering, but I've also pushed https://github.com/metal3-io/ironic-image/pull/89 which adds the check/wait for the provisioining IP in the container entrypoint scripts

hardys:
After discussion it sounds like the preferred approach is to just update the containers to wait ref:

metal3-io/ironic-image#88
metal3-io/ironic-image#89
openshift/ironic-image#15 (cherry-pick of above two PRs)

So I'll close this PR

PR #383, opened 2019-08-19
Fix status update when operator does not have new version

mfojtik:
The available message is cleaned up and and "running resync" progressing
reason is set multiple times per second even if there is nothing to sync
and there is no progress to be made because current and desired versions
are matching.

/cc @smarterclayton 
/cc @enxebre

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -57,17 +57,16 @@ func (optr *Operator) statusProgressing() error {
 		return err
 	}
 
-	var message string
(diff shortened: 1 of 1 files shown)


enxebre:
re-syncing should happen when the informer watches an event on the openshift-machine-api/deployments (to reconcile the mao owned deployment if this is modified out of band), the feature gate or the resync period goes through. Instead let's set the resync period to a higher explicit value https://github.com/openshift/machine-api-operator/pull/392

enxebre:
covered by https://github.com/openshift/machine-api-operator/pull/695

PR #384, opened 2019-08-19
rbac: Allow update and patch of secrets.

russellb:
When using the baremetal infrastructure platform, the
baremetal-operator component within the metal3 deployment needs to be
able to set the owner reference on Secrets.

Secrets are used to hold credentials used to talk to the out-of-band
management controllers (BMCs) on bare metal servers.  These Secrets
are associated with a BareMetalHost resource.  The BareMetalHost
controller will set the owner on the Secret.

### DIFF
--- a/install/0000_30_machine-api-operator_09_rbac.yaml
+++ b/install/0000_30_machine-api-operator_09_rbac.yaml
@@ -55,6 +55,8 @@ rules:
       - list
       - watch
       - create
+      - update
+      - patch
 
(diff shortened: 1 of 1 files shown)


ingvagabund:
We are trying to avoid granting other than read permissions unless it's really necessary. It's a precaution against accidentally modifying/deleting resources controller are not allowed to modify/delete. It's unfortunate for the baremetal provider. Though, if you need it, you need it.

/approve

PR #385, opened 2019-08-20
RBAC: Add additional metal3.io resource access

rdoxenham:
When utilising the baremetal infrastructure platform the baremetal
operator component needs to be able to update additional resources
for baremetal hosts, namely the status and the finalizer. This patch
ensures that it can do so, and will now allow hosts to be registered
with OpenShift as `BareMetalHosts` and as a result inside of Ironic.

### DIFF
--- a/install/0000_30_machine-api-operator_09_rbac.yaml
+++ b/install/0000_30_machine-api-operator_09_rbac.yaml
@@ -83,6 +83,8 @@ rules:
       - metal3.io
     resources:
       - baremetalhosts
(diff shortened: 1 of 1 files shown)


rdoxenham:
This is a follow on from #384 and has been tested locally.

hardys:
cc @dhellmann @russellb but lgtm, thanks for digging into it!

russellb:
/lgtm

Didn’t realize those had to be listed separately.  TIL

PR #389, opened 2019-08-28
Create maraidb_password within the MAO for use by the baremetal IPI deployment

sadasu:
Several containers with the Metal3 deployment need to have access to the mariadb_password. Generate a random password within the MAO and make it available within the openshift-machine-api namespace only if it doesn't already exist.

Mariadb-password is currently being created by the installer at : https://github.com/openshift-metal3/dev-scripts/blob/master/utils.sh#L193. Once the capability to generate the maridb password from within the MAO has been added, it will be removed from the installer.

### DIFF
--- a/install/0000_30_machine-api-operator_09_rbac.yaml
+++ b/install/0000_30_machine-api-operator_09_rbac.yaml
@@ -193,6 +193,13 @@ rules:
       - list
       - watch
 
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - get
+      - create
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
--- a/pkg/operator/baremetal_pod.go
+++ b/pkg/operator/baremetal_pod.go
@@ -1,18 +1,28 @@
 package operator
 
 import (
+	"math/rand"
+	"time"
+
+	"github.com/golang/glog"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/utils/pointer"
 )
 
-var baremetalConfigmap = "metal3-config"
-var sharedVolume = "metal3-shared"
+const (
+	baremetalConfigmap    = "metal3-config"
+	baremetalSharedVolume = "metal3-shared"
+	baremetalSecretName   = "metal3-mariadb-password"
+	baremetalSecretKey    = "password"
+)
 
 var volumes = []corev1.Volume{
 	{
-		Name: sharedVolume,
+		Name: baremetalSharedVolume,
 		VolumeSource: corev1.VolumeSource{
 			EmptyDir: &corev1.EmptyDirVolumeSource{},
 		},
@@ -21,7 +31,7 @@ var volumes = []corev1.Volume{
 
 var volumeMounts = []corev1.VolumeMount{
 	{
-		Name:      sharedVolume,
+		Name:      baremetalSharedVolume,
 		MountPath: "/shared",
 	},
 }
@@ -46,12 +56,46 @@ func setMariadbPassword() corev1.EnvVar {
 		ValueFrom: &corev1.EnvVarSource{
 			SecretKeyRef: &corev1.SecretKeySelector{
 				LocalObjectReference: corev1.LocalObjectReference{
-					Name: "mariadb-password",
+					Name: baremetalSecretName,
 				},
-				Key: "password",
+				Key: baremetalSecretKey,
 			},
 		},
 	}
+}
+
+func generateRandomPassword() string {
+	rand.Seed(time.Now().UnixNano())
+	chars := []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ" +
+		"abcdefghijklmnopqrstuvwxyz" +
+		"0123456789")
+	length := 16
+	buf := make([]rune, length)
+	for i := range buf {
+		buf[i] = chars[rand.Intn(len(chars))]
+	}
+	return (string(buf))
+}
+
+func createMariadbPasswordSecret(client coreclientv1.SecretsGetter, config *OperatorConfig) error {
+	glog.V(3).Info("Checking if the Maridb password secret already exists")
+	_, err := client.Secrets(config.TargetNamespace).Get(baremetalSecretName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		// Secret does not already exist. So, create one.
+		_, err := client.Secrets(config.TargetNamespace).Create(
+			&corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      baremetalSecretName,
+					Namespace: config.TargetNamespace,
+				},
+				StringData: map[string]string{
+					baremetalSecretKey: generateRandomPassword(),
+				},
+			},
+		)
+		return err
+	}
+	return err
 }
 
 func newMetal3Deployment(config *OperatorConfig) *appsv1.Deployment {
--- a/pkg/operator/baremetal_test.go
+++ b/pkg/operator/baremetal_test.go
@@ -1,0 +1,73 @@
+package operator
+
+import (
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	fakekube "k8s.io/client-go/kubernetes/fake"
+)
+
+func TestGenerateRandomPassword(t *testing.T) {
+	pwd := generateRandomPassword()
+	if pwd == "" {
+		t.Errorf("Expected a valid string but got null")
+	}
+}
+
+func newOperatorWithBaremetalConfig() *OperatorConfig {
+	return &OperatorConfig{
+		targetNamespace,
+		Controllers{
+			"docker.io/openshift/origin-aws-machine-controllers:v4.0.0",
+			"docker.io/openshift/origin-machine-api-operator:v4.0.0",
+			"docker.io/openshift/origin-machine-api-operator:v4.0.0",
+		},
+		BaremetalControllers{
+			"quay.io/openshift/origin-baremetal-operator:v4.2.0",
+			"quay.io/openshift/origin-ironic:v4.2.0",
+			"quay.io/openshift/origin-ironic-inspector:v4.2.0",
+			"quay.io/openshift/origin-ironic-ipa-downloader:v4.2.0",
+			"quay.io/openshift/origin-ironic-rhcos-downloader:v4.2.0",
+			"quay.io/openshift/origin-ironic-static-ip-manager:v4.2.0",
+		},
+	}
+}
+
+//Testing the case where the password does already exist
+func TestCreateMariadbPasswordSecret(t *testing.T) {
+	kubeClient := fakekube.NewSimpleClientset(nil...)
+	operatorConfig := newOperatorWithBaremetalConfig()
+	client := kubeClient.CoreV1()
+
+	// First create a mariadb password secret
+	if err := createMariadbPasswordSecret(kubeClient.CoreV1(), operatorConfig); err != nil {
+		t.Fatalf("Failed to create first Mariadb password. %s ", err)
+	}
+	// Read and get Mariadb password from Secret just created.
+	oldMaridbPassword, err := client.Secrets(operatorConfig.TargetNamespace).Get(baremetalSecretName, metav1.GetOptions{})
+	if err != nil {
+		t.Fatal("Failure getting the first Mariadb password that just got created.")
+	}
+	oldPassword, ok := oldMaridbPassword.StringData[baremetalSecretKey]
+	if !ok || oldPassword == "" {
+		t.Fatal("Failure reading first Mariadb password from Secret.")
+	}
+
+	// The pasword definitely exists. Try creating again.
+	if err := createMariadbPasswordSecret(kubeClient.CoreV1(), operatorConfig); err != nil {
+		t.Fatal("Failure creating second Mariadb password.")
+	}
+	newMaridbPassword, err := client.Secrets(operatorConfig.TargetNamespace).Get(baremetalSecretName, metav1.GetOptions{})
+	if err != nil {
+		t.Fatal("Failure getting the second Mariadb password.")
+	}
+	newPassword, ok := newMaridbPassword.StringData[baremetalSecretKey]
+	if !ok || newPassword == "" {
+		t.Fatal("Failure reading second Mariadb password from Secret.")
+	}
+	if oldPassword != newPassword {
+		t.Fatalf("Both passwords do not match.")
+	} else {
+		t.Logf("First Mariadb password is being preserved over re-creation as expected.")
+	}
+}
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -94,6 +94,12 @@ func (optr *Operator) syncClusterAPIController(config *OperatorConfig) error {
 }
 
 func (optr *Operator) syncBaremetalControllers(config *OperatorConfig) error {
+	// First create a Secret needed for the Metal3 deployment
+	if err := createMariadbPasswordSecret(optr.kubeClient.CoreV1(), config); err != nil {
+		glog.Error("Not proceeding with Metal3 deployment. Failed to create Mariadb password.")
+		return err
+	}
+
 	metal3Deployment := newMetal3Deployment(config)
 	_, updated, err := resourceapply.ApplyDeployment(optr.kubeClient.AppsV1(), metal3Deployment)
 	if err != nil {


pkg/operator/baremetal_pod.go
```diff
@@ -41,6 +46,7 @@ func setEnvVar(name string, key string) corev1.EnvVar {
 }
 
 func setMariadbPassword() corev1.EnvVar {
+	glog.Info("Inside setMariadbPassword")
```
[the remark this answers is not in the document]
sadasu: Done. Sorry I had put up a really early version to be able to talk about my current changes.

sadasu:
@enxebre and @michaelgugino would appreciate your comments on this.

pkg/operator/baremetal_pod.go
```diff
@@ -5,10 +5,23 @@ import (
...
+	"encoding/json"
+	"github.com/golang/glog"
+	"io/ioutil"
+	"math/rand"
+	"time"
+)
+
+const (
+	baremetalConfigmap = "metal3-config"
+	sharedVolume       = "metal3-shared"
```
dhellmann: These names will be global to the operator package, so they should all be clearly identified as being related to the baremetal platform.

pkg/operator/baremetal_pod.go
```diff
@@ -5,10 +5,23 @@ import (
...
+	sharedSecret       = "metal3-password"
+	secretName         = "mariadb-password"
+	secretKey          = "password"
 )
 
-var baremetalConfigmap = "metal3-config"
-var sharedVolume = "metal3-shared"
+var mariadbPassword = generateRandomPassword()
```
dhellmann: The timing for generating this value is an interesting problem. Doing it here means that every time the machine-api-operator restarts, we will end up with a new value. Is that what we want? Will that cause an update to the Secret, or Deployment? Can the baremetal code handle an update to the Secret?

Can we create a new password only if the Secret doesn't exist or doesn't have a valid password key? That would mean checking for an existing Secret before creating a new one and not updating it if it does exist, which complicates the logic but I think will make this more robust.

pkg/operator/baremetal_pod.go
```diff
@@ -46,14 +78,49 @@ func setMariadbPassword() corev1.EnvVar {
...
+			secretKey: mariadbPassword,
+		},
+	}
+	//Now convert it to a json file and store locally
+	secretJson, err := json.Marshal(secret)
+	if err != nil {
+		glog.Info("Converting Secret to Json returned error")
+		return err
+	}
+	//Write this to a local file
```
dhellmann: What is this for?

pkg/operator/baremetal_pod.go
```diff
@@ -17,13 +30,32 @@ var volumes = []corev1.Volume{
...
+	},
 }
 
 var volumeMounts = []corev1.VolumeMount{
 	{
 		Name:      sharedVolume,
 		MountPath: "/shared",
 	},
+	{
+		Name:      sharedSecret,
```
hardys: I don't think this additional volume should be needed, we just need to pass the value as before via an environment variable.

pkg/operator/baremetal_pod.go
hardys: Yeah, seems like we could just leave the name as something predictable e.g mariadb-password, and only create it when it doesn't exist?

pkg/operator/baremetal_pod.go
sadasu: Changed the names so that is clear that they are related to the baremetal platform.

pkg/operator/baremetal_pod.go
sadasu: Updated my diffs to create the password only when it doesn't already exist in the namespace.

pkg/operator/baremetal_pod.go
sadasu: The goal for this portion of the code was to add the Secret to the openshift-machine-api namespace. I now have a better way of doing this.

pkg/operator/baremetal_pod.go
sadasu: The maridb-password is already being passed into the relevant containers via environment variables. This is still part of the logic to add the Secret to the openshift-machine-api namespace so that it can later be used to update the env vars. I have a better way of doing it now.

sadasu:
Removing the WIP tag because I was able to test the case where the mraidb-password already existed and didn't have to be created along with the case where it needed to be created by MAO.

pkg/operator/baremetal_pod.go
```diff
@@ -46,14 +56,56 @@ func setMariadbPassword() corev1.EnvVar {
...
 		},
 	}
 }
 
+func generateRandomPassword() string {
+	rand.Seed(time.Now().UnixNano())
+	chars := []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ" +
+		"abcdefghijklmnopqrstuvwxyz" +
+		"0123456789")
+	length := 8
```
michaelgugino: We should make this a bit longer?  8 characters is pretty low given the character set.

pkg/operator/baremetal_pod.go
```diff
@@ -46,14 +56,56 @@ func setMariadbPassword() corev1.EnvVar {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      baremetalSecretName,
+				Namespace: config.TargetNamespace,
+			},
+			StringData: map[string]string{
+				baremetalSecretKey: mariadbPassword,
+			},
+		}
+
+		glog.V(3).Info("Maridb password does not exist so creating one.")
```
michaelgugino: v1 verbosity.

pkg/operator/baremetal_pod.go
```diff
@@ -1,18 +1,28 @@
...
+	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/utils/pointer"
 )
 
-var baremetalConfigmap = "metal3-config"
-var sharedVolume = "metal3-shared"
+const (
+	baremetalConfigmap    = "metal3-config"
+	baremetalSharedVolume = "metal3-shared"
+	baremetalSecretName   = "mariadb-password"
```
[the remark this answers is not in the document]
sadasu: Done.

pkg/operator/baremetal_pod.go
sadasu: Made it 16 bytes long. @imain was suggesting I use the uuid lib.

pkg/operator/baremetal_pod.go
sadasu: Done.

pkg/operator/baremetal_pod.go
```diff
@@ -46,14 +56,56 @@ func setMariadbPassword() corev1.EnvVar {
...
+					Name: baremetalSecretName,
 				},
-				Key: "password",
+				Key: baremetalSecretKey,
 			},
 		},
 	}
 }
 
+func generateRandomPassword() string {
```
imain: I wonder if there is a library we could use for this?  Generally I've always been taught to use trusted libs for crypto/random generators etc.

sadasu: I am using math/rand library to generate the random string using the unicode characters in "chars". rand is seeded with the current time so every time we try to generate a random password, the seed is different.

russellb:
Thanks for the review, Michael!  Before this goes in, I'd like to wait until Sandya's work to demonstrate this in the openshift-metal3 CI is working.  This will help with future MAO PRs while we're still working to get a job working in OpenShift CI.

https://github.com/openshift-metal3/dev-scripts/pull/799

/hold

russellb:
openshift-metal3 passed with this change - https://github.com/openshift-metal3/dev-scripts/pull/807

/hold cancel

dhellmann:
It looks like we need the local prow job to re-run.

install/0000_30_machine-api-operator_09_rbac.yaml
```diff
@@ -193,6 +193,19 @@ rules:
       - list
       - watch
 
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+    verbs:
```
enxebre: why do you need all this verbs?

pkg/operator/baremetal_pod.go
```diff
@@ -46,14 +56,56 @@ func setMariadbPassword() corev1.EnvVar {
...
+	for i := range buf {
+		buf[i] = chars[rand.Intn(len(chars))]
+	}
+	return (string(buf))
+}
+
+func createMariadbPasswordSecret(client coreclientv1.SecretsGetter, config *OperatorConfig) error {
+	glog.V(3).Info("Checking if the Maridb password secret already exists")
+	_, err := client.Secrets(config.TargetNamespace).Get(baremetalSecretName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
```
enxebre: this could be inlined:
```
	if _, err := client.Secrets(config.TargetNamespace).Get(baremetalSecretName, metav1.GetOptions{}); err != nil {
			// Secret does not already exist. So, create one.
			mariadbPassword := generateRandomPassword()
			secret := &corev1.Secret{
				ObjectMeta: metav1.ObjectMeta{
					Name:      baremetalSecretName,
					Namespace: config.TargetNamespace,
				},
				StringData: map[string]string{
					baremetalSecretKey: mariadbPassword,
				},
			}

			glog.V(1).Infof("Maridb password does not exist so creating one.")
			_, err := client.Secrets(config.TargetNamespace).Create(secret)
			return err
		}
		glog.Errorf("Error while looking for existing Mariadb password : %v", err)
		return err
	}
```

pkg/operator/sync.go
```diff
@@ -94,6 +94,13 @@ func (optr *Operator) syncClusterAPIController(config *OperatorConfig) error {
 }
 
 func (optr *Operator) syncBaremetalControllers(config *OperatorConfig) error {
+	// First create a Secret needed for the Metal3 deployment
+	err := createMariadbPasswordSecret(optr.kubeClient.CoreV1(), config)
```
enxebre: this could be inlined

install/0000_30_machine-api-operator_09_rbac.yaml
sadasu: machine-api-operator needs to read if the metal3-mariadb-password already exists and then creates it if doesn't. So, we would need create, update, delete and get. And, I looked at the verbs used for other "secret" resources in the file for guidance.

pkg/operator/sync.go
enxebre: It's a bit awkward  `syncBaremetalControllers` is now creating a secret. Why is this needed and it wasn't before?

sadasu: Our baremetal installer (https://github.com/openshift-metal3/dev-scripts) was creating the password before. The best place to create it would be right before starting the Metal3 deployment. Once this capability is added to the MAO, then it will be removed from our installer.

install/0000_30_machine-api-operator_09_rbac.yaml
enxebre: mm I'm confused, based on your comment and looking at the PR it seems you'd only need `get/create`

pkg/operator/sync.go
enxebre: I see, could you please elaborate and point to where this was created originally in the PR description?

pkg/operator/baremetal_pod.go
sadasu: Done, I think.

install/0000_30_machine-api-operator_09_rbac.yaml
sadasu: Update verbs associated with secret in Role to just get and create.

pkg/operator/sync.go
hardys: Note that dev-scripts is not intended for end-users, so this is a gap which we need to fix to enable non-dev users to interact directly with openshift-install instead.

pkg/operator/baremetal_pod.go
```diff
@@ -46,14 +56,48 @@ func setMariadbPassword() corev1.EnvVar {
...
+		"0123456789")
+	length := 16
+	buf := make([]rune, length)
+	for i := range buf {
+		buf[i] = chars[rand.Intn(len(chars))]
+	}
+	return (string(buf))
+}
+
+func createMariadbPasswordSecret(client coreclientv1.SecretsGetter, config *OperatorConfig) error {
```
enxebre: @sadasu can we have a unit test for this func?

sadasu: @enxebre please look at baremetal_test.go for the unit tests.

pkg/operator/baremetal_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	password, _ := maridbPassword.StringData["password"]
+	if password == "" {
+		t.Errorf("Did not find newly created Mariadb password.")
+		return
+	}
+}
+
+//Testing the case where the password does already exist
+func TestCreateDuplicateMariadbPasswordSecret(t *testing.T) {
```
enxebre: Thanks for adding unit tests @sadasu . This is covering the case above as well, can we just have `TestCreateMariadbPasswordSecret` which validate all cases the way this function is doing or having testCases which pre-creates or not a secret?

pkg/operator/baremetal_test.go
```diff
@@ -0,0 +1,98 @@
...
+	client := kubeClient.CoreV1()
+
+	// First create a mariadb password secret
+	if err := createMariadbPasswordSecret(kubeClient.CoreV1(), operatorConfig); err != nil {
+		t.Errorf("Failed to create new Mariadb password. %s ", err)
+		return
+	}
+	// Make sure mariadb password exists.
+	oldMaridbPassword, _ := client.Secrets(operatorConfig.TargetNamespace).Get(baremetalSecretName, metav1.GetOptions{})
+	oldPassword, _ := oldMaridbPassword.StringData["password"]
```
enxebre: this should fatal if `!ok`?

pkg/operator/baremetal_test.go
```diff
@@ -0,0 +1,98 @@
...
+	if oldPassword == "" {
+		t.Errorf("Did not find Mariadb password.")
+		return
```
enxebre: why return here?

pkg/operator/baremetal_test.go
```diff
@@ -0,0 +1,98 @@
...
+	} else {
+		t.Logf("Old Mariadb password : %s", oldPassword)
+	}
+
+	// The pasword definitely exists. Try creating again.
+	if err := createMariadbPasswordSecret(kubeClient.CoreV1(), operatorConfig); err != nil {
+		t.Errorf("Error creating duplicate Mariadb password.")
```
enxebre: shouldn't this  be fatal?

pkg/operator/baremetal_test.go
```diff
@@ -0,0 +1,98 @@
...
+		return
+	}
+	newMaridbPassword, _ := client.Secrets(operatorConfig.TargetNamespace).Get(baremetalSecretName, metav1.GetOptions{})
+	newPassword, _ := newMaridbPassword.StringData["password"]
```
enxebre: this should fatal if `!ok`?

pkg/operator/baremetal_test.go
enxebre: `newPassword, ok := newMaridbPassword.StringData["password"]
if !ok {
  then fatal
}
`

pkg/operator/baremetal_test.go
sadasu: Sure. Will have one test case that does what is currently being done in TestCreateDuplicateMariadbPasswordSecret.

Ideally, I would like to 1st run TestCreateMariadbPasswordSecret() and then run TestCreateDuplicateMariadbPasswordSecret() only when the 1st was successful. Not sure how to do that though.

pkg/operator/baremetal_test.go
```diff
@@ -0,0 +1,78 @@
...
+		t.Errorf("Error reading first Mariadb password from Secret.")
+	} else {
+		t.Logf("First Mariadb password : %s", oldPassword)
+	}
+
+	// The pasword definitely exists. Try creating again.
+	if err := createMariadbPasswordSecret(kubeClient.CoreV1(), operatorConfig); err != nil {
+		t.Fatal("Failure creating second Mariadb password.")
+	}
+	newMaridbPassword, _ := client.Secrets(operatorConfig.TargetNamespace).Get(baremetalSecretName, metav1.GetOptions{})
```
enxebre: as above: why are we ignoring error here?

pkg/operator/baremetal_test.go
```diff
@@ -0,0 +1,78 @@
...
+		return
+	}
+	// Make sure mariadb password exists.
+	oldMaridbPassword, _ := client.Secrets(operatorConfig.TargetNamespace).Get(baremetalSecretName, metav1.GetOptions{})
+	oldPassword, ok := oldMaridbPassword.StringData["password"]
+	if !ok {
+		t.Fatal("Failure extracting first Mariadb password from Secret.")
+	}
+	if oldPassword == "" {
+		t.Errorf("Error reading first Mariadb password from Secret.")
```
enxebre: does it make sense to continue if we reach this? If not this should be fatal

pkg/operator/baremetal_test.go
```diff
@@ -0,0 +1,78 @@
...
+	client := kubeClient.CoreV1()
+
+	// First create a mariadb password secret
+	if err := createMariadbPasswordSecret(kubeClient.CoreV1(), operatorConfig); err != nil {
+		t.Errorf("Failed to create new Mariadb password. %s ", err)
+		return
+	}
+	// Make sure mariadb password exists.
+	oldMaridbPassword, _ := client.Secrets(operatorConfig.TargetNamespace).Get(baremetalSecretName, metav1.GetOptions{})
+	oldPassword, ok := oldMaridbPassword.StringData["password"]
```
enxebre: can we use `baremetalSecretKey` constant here?

pkg/operator/baremetal_test.go
```diff
@@ -0,0 +1,78 @@
...
+	operatorConfig := newOperatorWithBaremetalConfig()
+	client := kubeClient.CoreV1()
+
+	// First create a mariadb password secret
+	if err := createMariadbPasswordSecret(kubeClient.CoreV1(), operatorConfig); err != nil {
+		t.Errorf("Failed to create new Mariadb password. %s ", err)
+		return
+	}
+	// Make sure mariadb password exists.
+	oldMaridbPassword, _ := client.Secrets(operatorConfig.TargetNamespace).Get(baremetalSecretName, metav1.GetOptions{})
```
enxebre: why are we ignoring the error here?

pkg/operator/baremetal_pod.go
```diff
@@ -46,14 +56,48 @@ func setMariadbPassword() corev1.EnvVar {
...
+	glog.V(3).Info("Checking if the Maridb password secret already exists")
+	_, err := client.Secrets(config.TargetNamespace).Get(baremetalSecretName, metav1.GetOptions{})
```
enxebre: I'd rather return early here, if err != nil return

sadasu: "return err" in line # 98 takes care of returning the error when the error is something other than IsNotFound() and also the case when there is no error at all.

sadasu:
Thank you @enxebre  and @michaelgugino for your reviews!

PR #390, opened 2019-08-30
bug 1747246: nodelink_controller: fix SEGV in findNodeFromMachine()

frobware:
If an error occurs looking up the machine we went on to log the
machine name using `machine.GetName()` as part of the error message.
That triggered a nil pointer dereference given that machine is nil
when lookup fails.

Add explicit unit test so that we don't regress.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1747246

### DIFF
--- a/pkg/controller/nodelink/nodelink_controller_test.go
+++ b/pkg/controller/nodelink/nodelink_controller_test.go
@@ -2,8 +2,10 @@ package nodelink
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/controller/nodelink/nodelink_controller_test.go
```diff
@@ -770,3 +772,36 @@ func TestUpdateNodeRef(t *testing.T) {
 		}
 	}
 }
+
+func TestFindMachineFromNodeDoesNotPanicBZ1747246(t *testing.T) {
+	testMachine := machine("matchingInternalIP", "test", []corev1.NodeAddress{
+		{
```
enxebre: nice

PR #392, opened 2019-09-03
Resync

enxebre:
Increase resync period and remove unused field
Related to https://github.com/openshift/machine-api-operator/pull/383

### DIFF
--- a/cmd/machine-api-operator/controller_context.go
+++ b/cmd/machine-api-operator/controller_context.go
@@ -1,11 +1,8 @@
 package main
 
 import (
-	"time"
-
 	configinformersv1 "github.com/openshift/client-go/config/informers/externalversions"
 	machineinformersv1beta1 "github.com/openshift/cluster-api/pkg/client/informers_generated/externalversions"
-
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/informers"
 )
@@ -23,8 +20,6 @@ type ControllerContext struct {
 	Stop <-chan struct{}
 
 	InformersStarted chan struct{}
-
-	ResyncPeriod func() time.Duration
 }
 
 // CreateControllerContext creates the ControllerContext with the ClientBuilder.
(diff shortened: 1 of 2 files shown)


enxebre:
cc @vikaschoudhary16

cmd/machine-api-operator/start.go
```diff
@@ -159,3 +171,30 @@ func startHTTPMetricServer(metricsPort string) {
...
+
+	// add a uniquifier so that two processes on the same host don't accidentally both become active
+	id = id + "_" + string(uuid.NewUUID())
+
+	return &resourcelock.ConfigMapLock{
+		ConfigMapMeta: metav1.ObjectMeta{
+			Namespace: componentNamespace,
+			Name:      componentName,
+		},
+		Client: cb.KubeClientOrDie("leader-election").CoreV1(),
```
frobware: Return an error? It makes unit testing easier.

enxebre: The second commit is just moving code as there's no need to have a helpers file for this function. Let's keep this PR for changing the default resync and I'll put any farther improvement in a different one

enxebre:
```2019/09/03 12:51:03 Pod e2e-aws-operator succeeded after 52m49s
2019/09/03 12:51:03 error: unable to signal to artifacts container to terminate in pod e2e-aws-operator, triggering deletion: could not run remote command: container artifacts is not valid for pod e2e-aws-operator
2019/09/03 12:51:03 error: unable to retrieve artifacts from pod e2e-aws-operator: could not read gzipped artifacts: container artifacts is not valid for pod e2e-aws-operator```
/retest

ingvagabund:
> There's no need for randomness on the resync period. Let's set ResyncPeriod explicitly to a higher value, e.g 20 min.

Any reason why it's not needed anymore? The point of randomizing the resync period is to avoid cases where to many controllers request to list objects simultaneously. Thus, hitting the apiserver with a lot of requests in a short period of time.

frobware:
> Any reason why it's not needed anymore? The point of randomizing the resync period is to avoid cases where to many controllers request to list objects simultaneously. Thus, hitting the apiserver with a lot of requests in a short period of time.

Why don't we use 0?

PR #391, opened 2019-08-30
Add machine resource to related objects

michaelgugino:
This commit ensures all machines and machinesets
are included in related-resources.

Implements: https://jira.coreos.com/browse/CLOUD-563

PR #395, opened 2019-09-05
Add baremetalhosts.metal3.io to related objects.

russellb:
This patch is similar to PR #391, which added machines and
machinesets to the related objects list for the cluster operator.
This will ensure that BareMetalHosts are collected by must-gather, as
well.

This is only applicable when the "baremetal" infrastructure platform
is in use, but is harmless otherwise as the CRD is always defined.


enxebre:
close in favour of https://github.com/openshift/machine-api-operator/pull/630

PR #399, opened 2019-09-19
Go modules

enxebre:
Drop dep in favour of go modules

`go mod init`
`go mod tidy`
`go mod vendor`

Use replace for forks, prometheus.client and code-generator.
Enforce `GO111MODULE=on` and `GOFLAGS += -mod=vendor` via Make file and DOCKER_CMD


enxebre:
/hold
until we open master for 4.3

Issue #407, opened 2019-10-09
openshift-tests: Pods found with invalid container image pull policy not equal to IfNotPresent


derekhiggins:
Running openshift-tests against a baremetal ipi deployed cluster fails with 

fail [github.com/openshift/origin/test/extended/operators/images.go:142]: Oct  7 09:22:27.146: Pods found with invalid container image pull policy not equal to IfNotPresent: 

```
openshift-machine-api/metal3-788d885944-vn99t/metal3-dnsmasq imagePullPolicy=Always
openshift-machine-api/metal3-788d885944-vn99t/metal3-httpd imagePullPolicy=Always
openshift-machine-api/metal3-788d885944-vn99t/metal3-ipa-downloader imagePullPolicy=Always
openshift-machine-api/metal3-788d885944-vn99t/metal3-ironic-api imagePullPolicy=Always
openshift-machine-api/metal3-788d885944-vn99t/metal3-ironic-conductor imagePullPolicy=Always
openshift-machine-api/metal3-788d885944-vn99t/metal3-ironic-inspector imagePullPolicy=Always
openshift-machine-api/metal3-788d885944-vn99t/metal3-mariadb imagePullPolicy=Always
openshift-machine-api/metal3-788d885944-vn99t/metal3-rhcos-downloader imagePullPolicy=Always
openshift-machine-api/metal3-788d885944-vn99t/metal3-static-ip-manager imagePullPolicy=Always
openshift-machine-api/metal3-788d885944-vn99t/metal3-static-ip-set imagePullPolicy=Always

```
test here
openshift-machine-api/
https://github.com/openshift/origin/blob/4f3d9bd4502a841922b35914d0fd3036d54c5e64/test/extended/operators/images.go#L136

PR #408, opened 2019-10-09
Switch imagePullPolicy in baremetal pod to IfNotPresent

derekhiggins:
openshift-tests says "Always" is a invalid container image pull policy.

Fixes: #407


enxebre:
/approve
cc @russellb

Issue #409, opened 2019-10-09
openshift-tests: Managed cluster should ensure control plane pods do not run in best-effort QoS


derekhiggins:
openshift-tests complains about the Qos of the baremetal pod

failed: (1.3s) 2019-10-07T13:21:43 "[Feature:Platform][Smoke] Managed cluster should ensure control plane pods do not run in best-effort QoS [Suite:openshift/conformance/parallel]"

openshift-machine-api/metal3-788d885944-vn99t is running in best-effort QoS

https://github.com/openshift/origin/blob/32e88dec78fa081dafa31c9ea1d79897761a0cc3/test/extended/operators/qos.go#L45

PR #410, opened 2019-10-09
Make qosClass of the bm pod Burstable

derekhiggins:
Set a Memory requirement on the most memory hungry
part of the baremetal pod.

Not having any resource requirements results in a
qosClass == best-effort. openshift-tests requires the pod
to be either Burstable or Guaranteed. Setting a single resource
will result in it being Burstable.

Fixes: #409

**Untested patch - will remove message once tested**


derekhiggins:
No longer relevant

PR #411, opened 2019-10-10
Move unhealthy conditions list to MachineHealthCheck CRD

bison:
This moves the list of unhealthy conditions from the global ConfigMap to the MachineHealthCheck CRD, giving each health-check its own set of conditions.


cynepco3hahue:
@bison Under the MDB controller, we also should fetch conditions from the MHC CR.

bison:
@cynepco3hahue right now, it looks like the MDB controller only looks at the Node's `Ready` condition to determine if it's healthy or not. There's not a one-to-one mapping between MachineDisruptionBudget and MachineHealthCheck resources, so I'm not sure how it could look at these conditions. Am I missing something?

cynepco3hahue:
Sorry for the noise, I added this functionality under our fork and was sure that I backported it under the MAO.
But in general MDB and MHC should have the same unhealthy conditions for the machine, otherwise, it will be very hard to predict when MDB will prevent machine remediation.

enxebre:
> Sorry for the noise, I added this functionality under our fork and was sure that I backported it under the MAO.
> But in general MDB and MHC should have the same unhealthy conditions for the machine, otherwise, it will be very hard to predict when MDB will prevent machine remediation.

fwiw we plan to keep mdb under feature gate https://github.com/openshift/enhancements/blob/master/enhancements/machine-health-checking.md

cynepco3hahue:
@enxebre Oh I wasn't aware of it, so will you currently move MDB functionality under the MHC or we will not have MDB functionality at all under 4.3 out of feature gate.

enxebre:
> @enxebre Oh I wasn't aware of it, so will you currently move MDB functionality under the MHC or we will not have MDB functionality at all under 4.3 out of feature gate.

we plan to go start with the simplest possible approach
https://github.com/openshift/enhancements/blob/master/enhancements/machine-health-checking.md#machinehealthcheck-crd

cynepco3hahue:
Will it include `minAvailable` parameter or only `maxUnavailable`?

beekhof:
> fwiw we plan to keep mdb under feature gate

this is a little concerning as MDB is the mechanism for preventing more that a single master from being fenced.  understood that its not a concern for cloud since masters cannot be reprovisioned yet, but the ability to fence masters is part of why we're using reboot.

pkg/apis/healthchecking/v1alpha1/machinehealthcheck_types.go
```diff
@@ -43,8 +44,39 @@ type MachineHealthCheckSpec struct {
...
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:MinLength=1
+	Status corev1.ConditionStatus `json:"status"`
+
+	// It would be preferable for timeout to be a metav1.Duration, but there's
+	// no good way to validate the format here.  Invalid input would cause
+	// problems with marshaling, so it's better to just make it a string and
+	// handle the conversion in the controller.
+	//
+	// Intentional blank line to keep this out of the OpenAPI description...
```
enxebre: can we add a comment with format example

enxebre:
> this is a little concerning as MDB is the mechanism for preventing more that a single master from being fenced. understood that its not a concern for cloud since masters cannot be reprovisioned yet, but the ability to fence masters is part of why we're using reboot.
> 
> Is there any option to leave it disabled by default, with a non-promoted v1alphaN api, but not behind the feature gate?

We plan to have a short circuit mechanism, but start with something simpler. Please any concern discuss it on the proposal https://github.com/openshift/machine-api-operator/pull/411#issuecomment-540617592

cynepco3hahue:
@bison Can you please create the relevant documentation PR?

PR #413, opened 2019-10-16
Marking a couple of fields in the BaremetalHost CRD optional

sadasu:
A couple of fields in the BaremetalHost CRD are being marked as
optional as part of https://github.com/metal3-io/baremetal-operator/pull/322.
Keeping the BaremetalHost CRD in the MAO current with those changes.


hardys:
Note for reviewers that https://github.com/openshift/baremetal-operator/pull/44 which was a dependency for this landed.

@sadasu I think you also need to remove the WIP from the commit message?

PR #414, opened 2019-10-17
Add short-circuiting support and drop MDB

enxebre:
Introduce support for short-circuiting based on: https://github.com/openshift/enhancements/blob/master/enhancements/machine-health-checking.md#goals

- Introduce `spec.MaxUnhealthy` (int/string with percentage)
- Introduce `status.ExpectedMachines`
- Introduce `status.CurrentHealthy`
- Every new function introduced is unit tested

The MHC does a best effort to wait for the cluster to stabilise and to ensure that no farther remediation takes place for a targeted pool of machines once its `MaxUnhealthy` threshold is reached. This is similar to what the node life cycle controller does for reducing the eviction rate as nodes goes unhealthy in a given zone.

Initial
```
I1017 11:52:39.074373   45304 machinehealthcheck_controller.go:97] Reconciling openshift-machine-api/example
I1017 11:52:39.074444   45304 machinehealthcheck_controller.go:110] Reconciling openshift-machine-api/example: finding targets
I1017 11:52:39.074697   45304 machinehealthcheck_controller.go:199] Reconciling openshift-machine-api/example/agl1610192-prs4g-worker-us-east-1a-sd962/ip-10-0-140-128.ec2.internal: health checking
I1017 11:52:40.328357   45304 machinehealthcheck_controller.go:134] Reconciling openshift-machine-api/example: monitoring mhc: total targets: 1,  maxUnavailable: 40%, unhealthy: 0. Remediations are allowed
I1017 11:52:40.328442   45304 machinehealthcheck_controller.go:162] Reconciling openshift-machine-api/example: no targets meet unhealthy criteria
```

During a scale out
```
I1017 11:55:00.546021   45359 machinehealthcheck_controller.go:97] Reconciling openshift-machine-api/example
I1017 11:55:00.546107   45359 machinehealthcheck_controller.go:110] Reconciling openshift-machine-api/example: finding targets
I1017 11:55:00.546643   45359 machinehealthcheck_controller.go:199] Reconciling openshift-machine-api/example/agl1610192-prs4g-worker-us-east-1a-sd962/ip-10-0-140-128.ec2.internal: health checking
I1017 11:55:00.546678   45359 machinehealthcheck_controller.go:199] Reconciling openshift-machine-api/example/agl1610192-prs4g-worker-us-east-1a-f4jl6/: health checking
[...]

PR #415, opened 2019-10-17
Drop strategy field

enxebre:
There's only two potential use case strategies today: delete and reboot.
Reboot is not yet supported at the machine API level but is rather a baremetal specific process.
Delete is not a good fit on baremetal.
As there's no specific strong use case or clear path forward yet for strategies reusable across multiple environments we don't want to bake this in the API just yet. Instead we are adding an annotation for the strategy on the MHC that can be leveraged for plugin strategies.
- Introduce remediationStrategyAnnotation = "healthchecking.openshift.io/strategy"
- Drop spec.RemediationStrategy
Needs https://github.com/openshift/machine-api-operator/pull/414


enxebre:
This is a follow up for https://github.com/openshift/machine-api-operator/pull/414
I'm fast merging this to enable farther development and put up the PR for:
Remove feature gate #416

/approve

PR #416, opened 2019-10-17
Remove feature gate from MHC

enxebre:
Remove feature gate from MHC


enxebre:
This is a follow up for #415
I'm fast merging this to enable farther QA testing and development

/approve

PR #422, opened 2019-10-21
reduce reconciling events

enxebre:
MAO is going through reconciles about every three minutes consistently. This is because the mao deployment receives update events which I can only think are coming from the CVO.
e.g
- https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-api-operator/417/pull-ci-openshift-machine-api-operator-master-e2e-aws-operator/1411/artifacts/e2e-aws-operator/pods/openshift-machine-api_machine-api-operator-6b74c4f5d6-zcd26_machine-api-operator.log
- https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-api-operator/422/pull-ci-openshift-machine-api-operator-master-e2e-aws/1534/artifacts/e2e-aws/pods/openshift-machine-api_machine-api-operator-7bf945646d-vc2pk_machine-api-operator.log

When CVO calls ensureDeployment it might wrongly return `modified=true` and do an api server update extra call when comparing the raw spec (without defaulted values) against the version fetch from the API server. See also https://github.com/openshift/cluster-version-operator/pull/256
This is one of a series of PRs to simplify mao reconciling loop and avoid unnecessary reconciling loops.
Follow up:
- Stop mao logic from watching mao/cao deployment and only watch "owned" deployments

### DIFF
--- a/cmd/machine-api-operator/start.go
+++ b/cmd/machine-api-operator/start.go
@@ -125,7 +125,7 @@ func startControllers(ctx *ControllerContext) {
 		ctx.ClientBuilder.KubeClientOrDie(componentName),
 		ctx.ClientBuilder.OpenshiftClientOrDie(componentName),
 		recorder,
-	).Run(2, ctx.Stop)
+	).Run(1, ctx.Stop)
 }
 
 func startMetricsCollectionAndServer(ctx *ControllerContext) {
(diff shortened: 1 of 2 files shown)


enxebre:
cc @wking

Issue #418, opened 2019-10-18
Update vendored openshift/api code 


sadasu:
The openshift/api code that is vendored with the machine-api-operator is currently outdated. I would like to make enhancements to the MAO that include accessing the "BareMetalPlatformStatus" structure in openshift/api/config/v1/types_infrastructure.go

sadasu:
This is no longer required.

PR #423, opened 2019-10-21
WIP: Updating vendor code for openshift/api

sadasu:
Attempting to provide a fix for: https://github.com/openshift/machine-api-operator/issues/418


sadasu:
We are not adding the new provisioning CRD to openshift/api. So, this PR is not necessary anymore.

PR #424, opened 2019-10-22
Only reconcile owned deployments

enxebre:
Only reconcile owned deployments for update/delete events.
This would prevent other deployments in the `openshift-machine-api` e.g autoscaler, mao deployment (owned by cvo) to trigger the reconciliation loop here.
Opted by annotation rather than ownerRefs pointing to mao deployment as this would increase complexity and skew to contemplate reconciliation when the mao is running remotely


bison:
Can you elaborate on why we wouldn't use an owner reference here? I'm actually surprised we don't already set one.

enxebre:
if we do so we'd need to account for the case where the mao is running remotely/not as a deployment, so there's no owner to set, therefore needing different business logic for both cases and having skew. MAO is a special case in the sense that the controller itself is the owner, usually the controller creates a CRD resource instance which own things

PR #426, opened 2019-10-23
Revendor controller runtime

enxebre:
We were pointing to a dead commit of a fork with kube 1.16. We can now point back to canonical repo since this is merged https://github.com/kubernetes-sigs/controller-runtime/pull/618


go.mod
```diff
@@ -58,5 +57,4 @@ replace k8s.io/client-go => github.com/openshift/kubernetes-client-go v0.0.0-201
 
 replace k8s.io/kube-aggregator => github.com/openshift/kubernetes-kube-aggregator v0.0.0-20190918161219-8c8f079fddc3
 
-// https://github.com/kubernetes-sigs/controller-runtime/pull/618
-replace sigs.k8s.io/controller-runtime => github.com/enxebre/controller-runtime v0.2.0-beta.1.0.20190930160522-58015f7fc885
+replace sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.3.1-0.20191016212439-2df793d02076
```
michaelgugino: Do we still require the replace directive here since we're not really replacing the source?  I think this can probably go in the required section instead?

enxebre: closed by accident, reopened here https://github.com/openshift/machine-api-operator/pull/430

PR #436, opened 2019-11-12
bug 1765427: Add index for getting machines from node.

enxebre:
This allows to find a backing machine for a node not found, therefore letting remediation to take place when reconciliation is triggered by a node deletion.

### DIFF
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -18,7 +18,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	apimachineryutilerrors "k8s.io/apimachinery/pkg/util/errors"
-	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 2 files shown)


pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -271,38 +295,32 @@ func (r *ReconcileMachineHealthCheck) getMachinesFromMHC(mhc mhcv1beta1.MachineH
...
+	machineList := &mapiv1.MachineList{}
+	if err := r.client.List(
 		context.TODO(),
-		types.NamespacedName{
-			Namespace: namespace,
-			Name:      machineName,
-		},
-		machine,
+		machineList,
+		client.MatchingFields{machineNodeNameIndex: node.Name},
```
bison: I guess this doesn't matter that much, but most of these have a function that returns the type, so I think you're kind of expected to instead write: `client.MatchingField(machineNodeNameIndex, node.Name)`

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -271,38 +295,32 @@ func (r *ReconcileMachineHealthCheck) getMachinesFromMHC(mhc mhcv1beta1.MachineH
...
 }
 
 func (r *ReconcileMachineHealthCheck) mhcRequestsFromNode(o handler.MapObject) []reconcile.Request {
 	glog.V(4).Infof("Getting MHC requests from node %q", namespacedName(o.Meta).String())
 	node := &corev1.Node{}
 	if err := r.client.Get(context.Background(), namespacedName(o.Meta), node); err != nil {
-		glog.Errorf("No-op: Unable to retrieve node %q from store: %v", namespacedName(o.Meta).String(), err)
-		return nil
+		if apimachineryerrors.IsNotFound(err) {
+			node.Name = o.Meta.GetName()
```
bison: If we only ever care about the name, we should maybe just change `getMachineFromNode()` to take the name as a parameter instead of the full object.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -38,22 +37,47 @@ const (
...
+	if err != nil {
+		return fmt.Errorf("error building reconciler: %v", err)
+	}
 	return add(mgr, r, r.mhcRequestsFromMachine, r.mhcRequestsFromNode)
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager, opts manager.Options) *ReconcileMachineHealthCheck {
+func newReconciler(mgr manager.Manager, opts manager.Options) (*ReconcileMachineHealthCheck, error) {
+	if err := mgr.GetCache().IndexField(&mapiv1.Machine{},
```
bison: Don't matter really, so not a blocker, but I feel like this might belong in `Add()` or `add()` instead of here.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
enxebre: yeh I started that way but `MatchingField()` is now deprecated and it recommends to user as above

bison: Ah, I missed that. I thought it was the other way around! Cool

PR #437, opened 2019-11-13
Add manifest annotations for hosted deployment exclusions

csrwng:
Enables the CVO to exclude manifests in an externally hosted control plane deployment
See https://github.com/openshift/cluster-version-operator/pull/252


csrwng:
/assign @bison 
This should be ready for merge

enxebre:
thanks @csrwng, do we have any docs that we can reference in this commit desc yet?
Why aren't we excluding the manifests not included in this PR

csrwng:
@enxebre not yet, given we’re still in the beta phase on hosted deployments of Openshift.

csrwng:
We’re only excluding manifests that cause the operator to come up. In some cases, namespaces and other artifacts are still used by other components.

PR #446, opened 2019-12-12
add exclusion annotations to all resources

KodieGlosserIBM:
**- What I did**
Added exclusion annotation so the 4.3 CVO will ignore these operators to stay consistent with 4.2 CVO. As seen here, the 4.3 CVO looks at the value when looking at exclusions https://github.com/openshift/hypershift-toolkit/blob/release-4.3/assets/cluster-version-operator/cluster-version-operator-deployment.yaml#L76-L77

The 4.2 CVO excluded the following operators and resources with those operators https://github.com/openshift/hypershift-toolkit/blob/release-4.2/assets/cluster-version-operator/cluster-version-operator-deployment.yaml#L60-L72
 
**- How to verify it**
Deploy the hypershift toolkit https://github.com/openshift/hypershift-toolkit/tree/release-4.2 and verify that all the resources with this exclusion (exclude.release.openshift.io/internal-openshift-hosted: "true") are not overwriting the ones being deployed into the cluster. 
**- Description for the changelog**

added exclusion to the rest of the resources that the operator manages.

### DIFF
--- a/install/0000_30_machine-api-operator_00_namespace.yaml
+++ b/install/0000_30_machine-api-operator_00_namespace.yaml
@@ -3,6 +3,7 @@ kind: Namespace
 metadata:
   annotations:
     openshift.io/node-selector: ""
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
   name: openshift-machine-api
   labels:
     name: openshift-machine-api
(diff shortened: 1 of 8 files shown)


enxebre:
this should be sorted by https://github.com/openshift/machine-api-operator/search?p=1&q=exclude.release.openshift.io%2Finternal-openshift-hosted&unscoped_q=exclude.release.openshift.io%2Finternal-openshift-hosted
Please reopen if still relevant. Thanks!

PR #458, opened 2020-01-09
vsphere: Add support for passing ignition user-data

bison:
This adds support for fetching the user-data secret referenced in the
ProviderSpec of a Machine using the vsphere provider, and passing that
data as ignition config via guestinfo variables.

### DIFF
--- a/pkg/controller/vsphere/reconciler.go
+++ b/pkg/controller/vsphere/reconciler.go
@@ -2,6 +2,7 @@ package vsphere
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
 	"strings"
 
@@ -23,6 +24,13 @@ const (
 	minCPU       = 2
 	diskMoveType = string(types.VirtualMachineRelocateDiskMoveOptionsMoveAllDiskBackingsAndConsolidate)
 	ethCardType  = "vmxnet3"
+)
+
+// These are the guestinfo variables used by Ignition.
+// https://access.redhat.com/documentation/en-us/openshift_container_platform/4.1/html/installing/installing-on-vsphere
+const (
+	GuestInfoIgnitionData     = "guestinfo.ignition.config.data"
+	GuestInfoIgnitionEncoding = "guestinfo.ignition.config.data.encoding"
 )
 
 // Reconciler runs the logic to reconciles a machine resource towards its desired state
@@ -216,6 +224,11 @@ func isRetrieveMONotFound(taskRef string, err error) bool {
 }
 
 func clone(s *machineScope) error {
+	userData, err := s.GetUserData()
+	if err != nil {
+		return err
+	}
+
 	vmTemplate, err := s.GetSession().FindVM(*s, s.providerSpec.Template)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


bison:
/hold

Needs further testing.

bison:
Still open question: Ignition uses guestinfo keys specific to Ignition, so this *only* supports Ignition as-is. I guess we're only interested in supporting Ignition in OpenShift, so should we maybe rename the `userDataSecret` field to `ignitionSecret`?

pkg/controller/vsphere/reconciler.go
```diff
@@ -507,3 +512,43 @@ func (vm *virtualMachine) getNetworkStatusList(client *vim25.Client) ([]NetworkS
...
+		},
+		&types.OptionValue{
+			Key:   "guestinfo.ignition.config.encoding",
+			Value: "base64",
+		},
+	}
+}
+
+// EncodeIgnitionConfig attempts to decode the given data until it looks to be
+// plain-text, then returns a base64 encoded version of that plain-text.
```
enxebre: nit: this is not ignition specific, could be may be called `base64DataToString`, `sanitiseBase64` or similar.

pkg/controller/vsphere/reconciler.go
```diff
@@ -507,3 +512,43 @@ func (vm *virtualMachine) getNetworkStatusList(client *vim25.Client) ([]NetworkS
...
+func EncodeIgnitionConfig(data []byte) string {
+	if len(data) == 0 {
+		return ""
+	}
+
+	for {
+		decoded, err := base64.StdEncoding.DecodeString(string(data))
+		if err != nil {
+			break
```
enxebre: nit: may be log something here

pkg/controller/vsphere/reconciler.go
```diff
@@ -507,3 +512,43 @@ func (vm *virtualMachine) getNetworkStatusList(client *vim25.Client) ([]NetworkS
...
+func IgnitionConfig(data []byte) []types.BaseOptionValue {
+	config := EncodeIgnitionConfig(data)
+
+	if config == "" {
+		return nil
+	}
+
+	return []types.BaseOptionValue{
+		&types.OptionValue{
+			Key:   "guestinfo.ignition.config.data",
```
enxebre: nit: could move this to constants

enxebre: where are these magic keys documented, can we add a reference here?

pkg/controller/vsphere/reconciler.go
bison: Like what? This is expected to break. It breaks when the input looks like plain-text, i.e. can no longer be successfully decoded. And the input is considered secret, so it shouldn't be logged.

enxebre: actually don't we want to return err here?

pkg/controller/vsphere/reconciler.go
bison: I'm not sure there's good documentation to link to yet unfortunately. Maybe the old CoreOS docs, but that seems a little weird.

pkg/controller/vsphere/reconciler.go
bison: No, we want to break and then re-encode as base64. This function ensures that you can pass either plain-text, or already encoded data (e.g. straight from a k8s secret) to `IgnitionConfig()` and have it do the right thing.

pkg/controller/vsphere/reconciler.go
bison: Fixed.

pkg/controller/vsphere/reconciler.go
bison: The implementation happens to not be Ignition specific, but the function is meant to encode Ignition data for inclusion in guestinfo variables, so I think the name should reflect that.

bison:
This is working now. We just need to decide if we call the field `UserDataSecret` or not. I guess I don't really care that much, but it seems a little misleading.

PR #459, opened 2020-01-09
vsphere: set providerID

enxebre:
Introduce support to set the providerID on the machine spec based on the biosUUID consistently with https://github.com/kubernetes/cloud-provider-vsphere/

/hold

### DIFF
--- a/pkg/controller/vsphere/machine_scope.go
+++ b/pkg/controller/vsphere/machine_scope.go
@@ -59,6 +59,9 @@ func newMachineScope(params machineScopeParams) (*machineScope, error) {
 	if err != nil {
 		return nil, fmt.Errorf("%v: error getting credentials: %v", params.machine.GetName(), err)
 	}
+	if providerSpec.Workspace == nil {
+		return nil, fmt.Errorf("%v: no workspace provided", params.machine.GetName())
+	}
 	authSession, err := session.GetOrCreate(context.TODO(),
 		providerSpec.Workspace.Server, providerSpec.Workspace.Datacenter,
 		user, password)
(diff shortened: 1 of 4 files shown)


pkg/controller/vsphere/machine_scope.go
```diff
@@ -78,19 +81,25 @@ func newMachineScope(params machineScopeParams) (*machineScope, error) {
...
+	// patch status
 	providerStatus, err := apivshpere.RawExtensionFromProviderStatus(s.providerStatus)
 	if err != nil {
 		return machineapierros.InvalidMachineConfiguration("failed to get machine provider status: %v", err.Error())
 	}
-
 	s.machine.Status.ProviderStatus = providerStatus
-	if err := s.client.Status().Patch(context.Background(), s.machine, s.machineToBePatched); err != nil {
-		klog.Errorf("Failed to update machine %q: %v", s.machine.GetName(), err)
+	if err := s.client.Status().Patch(context.Background(), s.machine.DeepCopy(), s.machineToBePatched); err != nil {
```
bison: Not sure I understand the design of the machineScope. Why do we deep copy the machine here? That means we discard the results of the update, and now both the `machine` and `machineToBePatched` fields are out of date.

pkg/controller/vsphere/reconciler.go
```diff
@@ -135,12 +140,41 @@ func (r *Reconciler) delete() error {
...
+	}
+	r.machine.Spec.ProviderID = &providerID
+	return nil
+}
+
+// convertUUIDToProviderID transforms a UUID string into a provider ID.
+// If the supplied UUID is empty or invalid an error is returned.
+// A valid UUID should adhere to the format specified by UUIDPattern.
+func convertUUIDToProviderID(UUID string) (string, error) {
+	pattern := regexp.MustCompile(UUIDPattern)
```
bison: I don't really have an objection to this, but we could also just `uuid.Parse()` the UUID and then `parsedUUID.String()` to make sure its in a known format for the provider ID. Might be a little more forgiving, and does away with the big regex.

PR #460, opened 2020-01-09
Adding a new CRD for configuration needed for Baremetal provisioning

sadasu:
A new CRD is being added to metal3.io API group that contains config
necessary to provision baremetal hosts.

### DIFF
--- a/install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
+++ b/install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
@@ -1,0 +1,148 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: provisionings.metal3.io
+spec:
+  group: metal3.io
+  names:
+    kind: Provisioning
+    listKind: ProvisioningList
+    plural: provisionings
+    singular: provisioning
+  scope: Cluster
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      description: 'Provisioning contains configuration used by the Provisioning
+        service (Ironic) to provision baremetal hosts. \n Provisioning is created
+        by the Openshift installer using admin or user provided information about
+        the provisioning network and the NIC on the server that can be used to PXE
+        boot it. \n This CR is a singleton, created by the installer and currently
+        only consumed by the machine-api-operator to bring up and update containers
+        in a metal3 cluster.'
+      type: object
+      required:
+      - spec
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: Specification of provisioning configuration for Metal3.
+          type: object
+          properties:
+            provisioningInterface:
+              description: provisioningInterface is the name of the network interface
+                on a baremetal server to the provisioning network. It can have values
+                like eth1 or ens3.
+              type: string
+            provisioningIP:
+              description: provisioningIP is the IP address assigned to the provisioningInterface
+                of the baremetal server. This IP address should be within the provisioning
+                subnet, and outside of the DHCP range.
+              type: string
+            provisioningNetworkCIDR:
+              description: provisioningNetworkCIDR is the network on which the baremetal
+                nodes are provisioned. The provisioningIP and the IPs in the dhcpRange
+                all come from within this network.
+              type: string
+            provisioningDHCPExternal:
+              description: provisioningDHCPExternal indicates whether the DHCP server for IP
+                addresses in the provisioning DHCP range is present within the metal3 cluster
+                or external to it.
+              type: boolean
+            provisioningDHCPRange:
+              description: Needs to be interpreted along with provisioningDHCPExternal.
+                If the value of provisioningDHCPExternal is set to False, then
+                provisioningDHCPRange represents the range of IP addresses that the DHCP server
+                running within the metal3 cluster can use while provisioning baremetal servers.
+                If the value of provisioningDHCPExternal is set to True, then the value of
+                provisioningDHCPRange will be ignored. When the value of provisioningDHCPExternal
+                is set to False, indicating an internal DHCP server and the value of
+                provisioningDHCPRange is not set, then the DHCP range is taken to be the default
+                range which goes from .10 to .100 of the provisioningNetworkCIDR. This is the only
+                value in all of the provisioning configuration that can be changed after the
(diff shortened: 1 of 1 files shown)


install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
```diff
@@ -0,0 +1,134 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: metal3provisionings.metal3.io
```
enxebre: I don't have strong opinions on how you want to shape this api and, mao just happens to be the integration point for this today so I'm good merging this as soon as test goes green but as an observation `metal3provisionings` seems odd t me,`metal3` is already in the domain so just `provisionings.metal3.io` would be more meaningful imho

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
```diff
@@ -0,0 +1,134 @@
...
+              description: observedGeneration is the last generation change you've
+                dealt with
+              type: integer
+              format: int64
+            readyReplicas:
+              description: readyReplicas indicates how many replicas are ready and
+                at the desired state
+              type: integer
+              format: int32
+  version: v1alpha1
```
enxebre: are you planning go `alpha` or promoting `beta` later on?

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
hardys: Having a plural "provisionings" also seems strange. should it be just provisioning.metal3.io?

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
```diff
@@ -0,0 +1,134 @@
...
+metadata:
+  name: metal3provisionings.metal3.io
+spec:
+  group: metal3.io
+  names:
+    kind: Metal3Provisioning
+    listKind: Metal3ProvisioningList
+    plural: metal3provisionings
+    shortNames:
+    - m3p
```
deads2k: not a blocker, but we generally encourage people to only use shortnames for resources that are frequently gotten

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
```diff
@@ -0,0 +1,134 @@
...
+    - m3ps
+    singular: metal3provisioning
+  scope: ""
+  subresources:
+    scale:
+      labelSelectorPath: .status.labelSelector
+      specReplicasPath: .spec.replicas
```
deads2k: Again, not a blocker for your own API, but we haven't generally placed scaling options here for operator-y things.  I just encourage you to be sure you want to.

deads2k:
A couple optional suggestions for consideration from me. Nothing that I would consider blocking.

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
sadasu: @enxebre I was stuck here. Since this is a singleton CR, plural really does not make sense but the unit tests fail without a value for plural.
Without Metal3, the CRD would be called Provisioning. I am pushing another version with these changes.

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
sadasu: @deads2k @enxebre 
I have a few questions regarding naming.
1. As Alberto rightly pointed out, metal3provisioning.metal3.io looks odd. Better to call it, provisioning.metal3.io. So, without the API group name, the CRD would be called just **provisioning**?
2. This is a singleton CR as of now, so it is odd to call the plural "provisionings"! But, not having a value for plural is causing the unti tests to fail.
3. If we did call the CRD "provisioning", then can I keep the "shortNames" the same as the regular names? Or would you prefer "prov" ?

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
sadasu: Not sure we want them so removing the scaling options.

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
sadasu: Promoting beta later on. But at this point want to leave it as alpha for this PR.

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
```diff
@@ -0,0 +1,152 @@
...
+metadata:
+  name: provisionings.metal3.io
+spec:
+  group: metal3.io
+  names:
+    kind: Provisioning
+    listKind: ProvisioningList
+    plural: provisionings
+    singular: provisioning
+  scope: ""
```
deads2k: I recommend using Cluster scope

deads2k:
I'm generally in favor of allowing you to create whatever API you wish, but are you sure this works?

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
sadasu: Done

sadasu:
> I'm generally in favor of allowing you to create whatever API you wish, but are you sure this works?

I was able to instantiate a CR and "get" it.

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
```diff
@@ -0,0 +1,148 @@
...
+    listKind: ProvisioningList
+    plural: provisionings
+    singular: provisioning
+  scope: Cluster
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      description: 'Provisioning contains configuration used by the Provisioning
+        service (Ironic) to provision baremetal hosts. \n Provisioning is created
```
hardys: nit I would just say "the metal3 provisioning services", Ironic is supposed to be an implementation detail.

hardys:
Small nit in the description but this lgtm, we can adjust the description in some subsequent PR if this is otherwise good to merge.

enxebre:
let's move on and do any farther interaction in follow ups
/lgtm

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
```diff
@@ -0,0 +1,148 @@
...
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: Specification of provisioning configuration for Metal3.
+          type: object
+          properties:
+            provisioningInterface:
+              description: provisioningInterface is the name of the network interface
+                on a baremetal server to the provisioning network. It can have values
```
imain: This sentence seems awkward to me.  I think I would say "proisioningInterface is the name of the network interface on a baremetal server that is used to provision new machines"

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
```diff
@@ -0,0 +1,148 @@
...
+                like eth1 or ens3.
```
imain: I like the examples here.

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
sadasu: Thank you for the comment @imain . I will take care of it in the next MAO patch containing changes to read this CR.

PR #466, opened 2020-01-14
vsphere: Rename UserDataSecret to IgnitionSecret

bison:
This renames the UserDataSecret field in VSphereMachineProviderSpec to
IgnitionSecret as the provider only supports passing Ignition data
currently and not other types of configuration.

### DIFF
--- a/pkg/controller/vsphere/machine_scope.go
+++ b/pkg/controller/vsphere/machine_scope.go
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	userDataSecretKey         = "userData"
+	ignitionSecretKey         = "ignition"
 	credentialsSecretUser     = "user"
 	credentialsSecretPassword = "password"
 )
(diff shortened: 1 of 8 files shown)


pkg/controller/vsphere/machine_scope.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	userDataSecretKey         = "userData"
+	ignitionSecretKey         = "ignition"
```
enxebre: just thinking today this is common to all platforms, this would require specific changes to the installer https://github.com/openshift/installer/blob/4bb7986830ec15de43f6f0b6a5f40d9b761925bf/pkg/asset/machines/worker.go#L313

https://github.com/openshift/installer/blob/316688eafae84eb850a20429ad2ab72856c6d851/pkg/asset/machines/userdata.go#L11-L22

bison: Ugh. Okay. Not worth it then. I'll close this.

PR #468, opened 2020-01-15
Update readme to reflect controllers

enxebre:
Update readme to reflect controllers


elmiko:
i don't know the source material that well, but the language in this change looks ok to me.

PR #469, opened 2020-01-16
Expose a dynamic client in the Operator struct

enxebre:
This is so baremetal can fetch the `provisioning.metal3.io` object and parameterize their deployment. Related openshift/api#540 and #460


enxebre:
cc @sadasu you can then just fetch the provisioning object and parameterise the baremetal deployment at your convenience.
Once all get merged please make sure to reflect the latest discussions here https://github.com/openshift/enhancements/blob/master/enhancements/baremetal/baremetal-provisioning-config.md

enxebre:
Alternative in case we wanted to watch these resources https://github.com/openshift/machine-api-operator/compare/master...enxebre:dynamic-informer?expand=1

PR #470, opened 2020-01-16
Read Metal3 config from a new CRD

sadasu:
Add functionality for MAO to read provisionings.metal3.io using a dynamic client and updating the baremetal config with config items needed to provision baremetal hosts.


install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
```diff
@@ -15,9 +15,9 @@ spec:
   validation:
     openAPIV3Schema:
       description: 'Provisioning contains configuration used by the Provisioning
```
enxebre: this should go into a different commit

pkg/operator/baremetal_config.go
```diff
@@ -0,0 +1,67 @@
+package operator
+
+import (
+	"github.com/golang/glog"
+
```
enxebre: extra spaces here and 8

pkg/operator/baremetal_test.go
```diff
@@ -3,10 +3,37 @@ package operator
 import (
 	"testing"
 
+	"sigs.k8s.io/yaml"
+
```
enxebre: extraspace

pkg/operator/baremetal_test.go
```diff
@@ -3,10 +3,37 @@ package operator
...
+	fakedynamic "k8s.io/client-go/dynamic/fake"
 	fakekube "k8s.io/client-go/kubernetes/fake"
 )
 
+var yamlContent = `
+apiVersion: metal3.io/v1alpha1
+kind: Provisioning
+metadata:
+  name: test
+  labels:
```
enxebre: please remove this label, it has no meaning

pkg/operator/baremetal_config.go
```diff
@@ -0,0 +1,67 @@
...
+func getBaremetalProvisioningConfig(dc dynamic.Interface, configName string, usingBareMetal bool) BaremetalConfig {
+	if !usingBareMetal {
+		return BaremetalConfig{}
+	}
+
+	provisioningClient := dc.Resource(provisioningGVR)
+
+	provisioningConfig, err := provisioningClient.Get(configName, metav1.GetOptions{})
+
```
enxebre: remove extra space

pkg/operator/baremetal_config.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	provisioningClient := dc.Resource(provisioningGVR)
+	provisioningConfig, err := provisioningClient.Get(configName, metav1.GetOptions{})
+	if err != nil {
+		glog.V(1).Infof("Error getting provisioning config %q", configName)
+		return BaremetalConfig{}
+	}
+	provisioningSpec, found, err := unstructured.NestedMap(provisioningConfig.UnstructuredContent(), "spec")
+	if !found {
+		glog.Errorf("Nested Spec not found in Baremetal provisioning CR %q", configName)
```
enxebre: don't you want to error here for all the bad cases? I'd suggest you move this function to be called by https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/sync.go#L77 instead of the current caller you can handle errors as you wish without impacting other platforms.
Also how is this going to handle defaults?

vendor/modules.txt
```diff
@@ -347,33 +347,34 @@ k8s.io/apimachinery/pkg/util/runtime
 k8s.io/apimachinery/pkg/watch
```
enxebre: please move vendoring into its own commit

pkg/operator/baremetal_config.go
sadasu: Done.

pkg/operator/baremetal_test.go
sadasu: Done.

pkg/operator/baremetal_test.go
sadasu: Done.

pkg/operator/baremetal_config.go
sadasu: Done.

install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml
sadasu: Agreed.

vendor/modules.txt
enxebre: also this is missing packages to be vendored /go/src/github.com/openshift/machine-api-operator/vendor/k8s.io/client-go/dynamic/fake, please include them via
go mod tidy, go mod vendor

pkg/operator/baremetal_config.go
sadasu: The current implementation will have to support reading from the CR first and if that fails, fall back to reading from the ConfigMap to support backwards compatibility among other things. So, when we hit errors, we plan to just return an empty BaremetalConfig struct. Logic that is coming in a following PR would see that we were not able to get config from the CR and fallback to  using the configmap.
https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/baremetal_pod.go#L17.

Regarding calling getBaremetalProvisioningConfig() from sync.go (https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/sync.go#L77 ). We added BaremetalConfig to OperatorConfig and ideally before we try to deploy metal3, OperatorConfig should be fully populated.
We could potentially, leave BaremetalConfig uninitialized until we try to deploy metal3. I was keeping the flow similar to the other baremetal related config, the BaremetalControllers.

vendor/modules.txt
sadasu: https://github.com/openshift/machine-api-operator/pull/472 has been created to handling vendoring for k8s.io/client-go/dynamic/fake. Will mark this PR as WIP until the vendoring merges.

pkg/operator/baremetal_config.go
enxebre: I see, so if you fail to read the CRD, or there's a missing field in the CRD, then you fallback to the configMap and if anything fails you return the empty struct. What if there's a transient failure to read from the API you just set the empty struct as well?
In any case I'd still prefer this to be in https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/sync.go#L77 so this is completely isolated from other platforms and you can change how to handle those errors anytime.

vendor/modules.txt
enxebre: please include any vendoring required by this PR in this PR

pkg/operator/baremetal_config.go
sadasu: That is correct. Updating to accommodate comment.

sadasu: Refactored to remove BaremetalConfig from OperatorConfig and moved initialization of BaremetalConfig to within https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/sync.go#L77 .

pkg/operator/baremetal_config.go
```diff
@@ -0,0 +1,57 @@
...
+	dynamic "k8s.io/client-go/dynamic"
+)
+
+var provisioningGVR = schema.GroupVersionResource{Group: "metal3.io", Resource: "provisionings", Version: "v1alpha1"}
+
+func getBaremetalProvisioningConfig(dc dynamic.Interface, configName string) BaremetalConfig {
+	provisioningClient := dc.Resource(provisioningGVR)
+	provisioningConfig, err := provisioningClient.Get(configName, metav1.GetOptions{})
+	if err != nil {
+		glog.V(1).Infof("Error getting provisioning config %q", configName)
```
enxebre: I find it weird to return `BaremetalConfig{}` here, at a minimum this should log error/warning rather than info

pkg/operator/baremetal_config.go
```diff
@@ -0,0 +1,57 @@
...
+	if !found {
+		glog.Errorf("provisioningDHCPExternal not found in Baremetal provisioning CR %q", configName)
+		return BaremetalConfig{}
+	}
+	provisioningDHCPRange, found, err := unstructured.NestedString(provisioningSpec, "provisioningDHCPRange")
+	if !found {
+		glog.Errorf("provisioningDHCPRange not found in Baremetal provisioning CR %q", configName)
+		return BaremetalConfig{}
+	}
+	return BaremetalConfig{
```
enxebre: wouldn't `BaremetalProvisioningConfig` be more appropriate?

pkg/operator/baremetal_test.go
```diff
@@ -71,3 +95,31 @@ func TestCreateMariadbPasswordSecret(t *testing.T) {
...
+		t.Errorf("failed to unmarshall input yaml content:%v", err)
+	}
+	dynamicClient := fakedynamic.NewSimpleDynamicClient(runtime.NewScheme(), u)
+	baremetalConfig := getBaremetalProvisioningConfig(dynamicClient, "test")
+	if baremetalConfig.ProvisioningInterface != expectedProvisioningInterface ||
+		baremetalConfig.ProvisioningIp != expectedProvisioningIP ||
+		baremetalConfig.ProvisioningNetworkCIDR != expectedProvisioningNetworkCIDR ||
+		baremetalConfig.ProvisioningDHCPExternal != expectedProvisioningDHCPExternal ||
+		baremetalConfig.ProvisioningDHCPRange != expectedProvisioningDHCPRange {
+		t.Errorf("failed getBaremetalProvisioningConfig. One or more BaremetalConfig items do not match the expected config.")
```
enxebre: can you render here the expected vs provisioned?

pkg/operator/config.go
```diff
@@ -58,6 +58,15 @@ type Images struct {
 	BaremetalStaticIpManager     string `json:"baremetalStaticIpManager"`
 }
```
enxebre: can we please declare this in `baremetal_config.go`

enxebre:
@bison PTAL

pkg/operator/baremetal_config.go
sadasu: Do you suggest a retry mechanism (good for transient errors) or a mechanism to bubble up an error (should not ignore errors) from this method?

pkg/operator/baremetal_pod.go
```diff
@@ -98,7 +98,7 @@ func createMariadbPasswordSecret(client coreclientv1.SecretsGetter, config *Oper
 	return err
 }
 
-func newMetal3Deployment(config *OperatorConfig) *appsv1.Deployment {
+func newMetal3Deployment(config *OperatorConfig, baremetalConfig BaremetalConfig) *appsv1.Deployment {
```
bison: I feel like I must be missing something. This reads the config from a new CRD so that it can pass it to this function, but then it appears to not be used.

sadasu: @bison yes, that is coming in a following PR.

pkg/operator/config.go
sadasu: Done.

pkg/operator/baremetal_config.go
sadasu: Done.

pkg/operator/baremetal_test.go
sadasu: Done.

pkg/operator/baremetal_pod.go
sadasu: Using it is coming in a subsequent PR. Adding TODO to indicate that in code.

pkg/operator/baremetal_test.go
```diff
@@ -71,3 +95,40 @@ func TestCreateMariadbPasswordSecret(t *testing.T) {
...
+func TestGetBaremetalProvisioningConfig(t *testing.T) {
+	u := &unstructured.Unstructured{Object: map[string]interface{}{}}
+	if err := yaml.Unmarshal([]byte(yamlContent), &u); err != nil {
+		t.Errorf("failed to unmarshall input yaml content:%v", err)
+	}
+	dynamicClient := fakedynamic.NewSimpleDynamicClient(runtime.NewScheme(), u)
+	baremetalConfig, err := getBaremetalProvisioningConfig(dynamicClient, "test")
+	if err != nil {
+		t.Logf("Unstructed Config:  %+v", u)
+		t.Fatalf("Failed to get Baremetal Provisioning Interface from CR %s", "test")
```
enxebre: no need to use %s here since you are passing a hardcoded string

pkg/operator/baremetal_test.go
```diff
@@ -71,3 +95,40 @@ func TestCreateMariadbPasswordSecret(t *testing.T) {
...
+	}
+	if baremetalConfig.ProvisioningInterface != expectedProvisioningInterface ||
+		baremetalConfig.ProvisioningIp != expectedProvisioningIP ||
+		baremetalConfig.ProvisioningNetworkCIDR != expectedProvisioningNetworkCIDR ||
+		baremetalConfig.ProvisioningDHCPExternal != expectedProvisioningDHCPExternal ||
+		baremetalConfig.ProvisioningDHCPRange != expectedProvisioningDHCPRange {
+		t.Logf("Actual BaremetalProvisioningConfig: %+v", baremetalConfig)
+		t.Logf("Expected : ProvisioningInterface: %s, ProvisioningIP: %s, ProvisioningNetworkCIDR: %s, ProvisioningDHCPExternal: %t, expectedProvisioningDHCPRange: %s", expectedProvisioningInterface, expectedProvisioningIP, expectedProvisioningNetworkCIDR, expectedProvisioningDHCPExternal, expectedProvisioningDHCPRange)
```
enxebre: it'd be nice to follow `got: n%v, expected: %v` format as everywhere else

pkg/operator/sync.go
```diff
@@ -74,13 +74,18 @@ func (optr *Operator) syncClusterAPIController(config *OperatorConfig) error {
 }
 
 func (optr *Operator) syncBaremetalControllers(config *OperatorConfig) error {
-	// First create a Secret needed for the Metal3 deployment
+	// Try to get baremetal provisioning config from a CR
+	baremetalProvisioningConfig, err := getBaremetalProvisioningConfig(optr.dynamicClient, baremetalProvisioningCR)
+	if err != nil {
+		glog.Infof("Unable to read Baremetal Provisioning config from CR %s.", baremetalProvisioningCR)
```
enxebre: I find it weird not returning err here, at minimum this should log.error, and make clear in the message the function will continue nevertheless

pkg/operator/baremetal_test.go
```diff
@@ -71,3 +95,40 @@ func TestCreateMariadbPasswordSecret(t *testing.T) {
...
+}
+
+func TestGetIncorrectBaremetalProvisioningCR(t *testing.T) {
+	u := &unstructured.Unstructured{Object: map[string]interface{}{}}
+	if err := yaml.Unmarshal([]byte(yamlContent), &u); err != nil {
+		t.Errorf("failed to unmarshall input yaml content:%v", err)
+	}
+	dynamicClient := fakedynamic.NewSimpleDynamicClient(runtime.NewScheme(), u)
+	baremetalConfig, err := getBaremetalProvisioningConfig(dynamicClient, "test1")
+	if err != nil {
```
enxebre: wouldn't be the client expected to error here and so `getBaremetalProvisioningConfig`

For units we prefer to have one test per function i.e `TestGetBaremetalProvisioningConfig` and loop over multiple test cases within it. e.g https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/vsphere/machine_scope_test.go#L139-L261

enxebre:
nit: we use the imperative mood in the subject line for commits rather than -ing, https://chris.beams.io/posts/git-commit/

pkg/operator/baremetal_config.go
sadasu: Done.

pkg/operator/sync.go
sadasu: In this release, we cannot return error and stop deploying Metal3 because during discussions around upgrade strategy, we had decided to have one release where the config could be obtained from either sources, the ConfigMap or the newly created Config resource.
In the current implementation, we are prioritizing reading config from the new Config Resource. When that fails, (could be for a simple reason that the user did not want to provide a CR and instead still wanted to use a configmap), then we should be able to proceed with the deployment. 
The other part of the comment talks about logging an error. Again, I worry that this might be interpreted as something went wrong. If you think seeing an error msg in the logs will not be interpreted as I think it will, I am fine with making this Info statement into an Error msg.

PR #472, opened 2020-01-17
Updating vendoring for k8s.io/client-go/dynamic/fake

sadasu:


sadasu:
@enxebre I am not sure if we need a separate PR for vendoring changes.  Please approve if this works. All required CIs are passing.

enxebre:
Hey @sadasu please let's just include whatever needs to be vendored in its own commit for the PR requiring it. I assume this is for https://github.com/openshift/machine-api-operator/pull/470

PR #474, opened 2020-01-22
Vsphere provider

enxebre:
Add support for mao to run the vsphere machine controller.

The machineset controller development was moved into this repo by #441 and the vsphere actuator lives in this repo, therefore the binaries live in the mao image unlike other providers with their own image e.g aws

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,4 +8,7 @@ COPY --from=builder /go/src/github.com/openshift/machine-api-operator/install ma
 COPY --from=builder /go/src/github.com/openshift/machine-api-operator/bin/machine-api-operator .
 COPY --from=builder /go/src/github.com/openshift/machine-api-operator/bin/nodelink-controller .
 COPY --from=builder /go/src/github.com/openshift/machine-api-operator/bin/machine-healthcheck .
+COPY --from=builder /go/src/github.com/openshift/machine-api-operator/bin/machineset ./manager
+COPY --from=builder /go/src/github.com/openshift/machine-api-operator/bin/vsphere ./machine-controller-manager
+
 LABEL io.openshift.release.operator true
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -8,4 +8,7 @@ COPY --from=builder /go/src/github.com/openshift/machine-api-operator/install ma
 COPY --from=builder /go/src/github.com/openshift/machine-api-operator/bin/machine-api-operator .
 COPY --from=builder /go/src/github.com/openshift/machine-api-operator/bin/nodelink-controller .
 COPY --from=builder /go/src/github.com/openshift/machine-api-operator/bin/machine-healthcheck .
+COPY --from=builder /go/src/github.com/openshift/machine-api-operator/bin/machineset ./manager
+COPY --from=builder /go/src/github.com/openshift/machine-api-operator/bin/vsphere ./machine-controller-manager
```
bison: This seems pretty confusing. Are we sure we don't want to build a separate image for the vSphere actuator? If the plan is to move more of these into this repo, we have to figure that out eventually anyway.

patrickdillon:
Do we need to create a cloud credential request for vsphere similar to other platforms? https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_00_credentials-request.yaml

When I run this I get this error:
```
I0123 18:10:09.749055       1 controller.go:162] padillon-8wclx-worker-zs2ff: reconciling Machine
I0123 18:10:09.749137       1 actuator.go:76] padillon-8wclx-worker-zs2ff: actuator checking if machine exists
E0123 18:10:09.749422       1 controller.go:254] padillon-8wclx-worker-zs2ff: failed to check if machine exists: padillon-8wclx-worker-zs2ff: failed to create scope for machine: padillon-8wclx-worker-zs2ff: error getting credentials: error getting credentials secret openshift-machine-api/vsphere-cloud-credentials: Secret "vsphere-cloud-credentials" not found
```

patrickdillon:
When I manually create the credentialrequest the workers are created in vsphere but no node is created: 
```
I0123 21:35:03.151990       1 controller.go:275] padillon-sx25d-worker-2jb4w: has no node yet, requeuing
```

Dockerfile
enxebre: I think I'd rather live with this oddity for now and get the benefits of it with no need to revendor at all for this provider. If we add another one in tree we'd could go (machine-controller-%v, platform) for naming the binaries

bison: Yeah, I can see the point about not having another image. I do think we should come up with some kind of wrapper that determines the platform and executes the correct controller / actuator though, but that can definitely wait until a follow-up.

PR #475, opened 2020-01-23
Bug 1800423: Apply the reboot annotation to the machine

beekhof:
Discussing with the metal3 folks, we’d like to watch the machine for the annotation instead if the node. This avoids the creation of a controller that just relays the annotation from the node to the machine.


beekhof:
Re-executed the tests, the failures do not appear related to this PR

beekhof:
/cc @bison or @michaelgugino - could i trouble one of you for a review and lgtm please?
Test failures are not related to this code (mostly etcd failures)

PR #476, opened 2020-01-23
Bug 1800425: More appropriate annotation name

beekhof:
Since MHC has no control over how the machine is remediated, it would be
better not to imply that it will (only) be via reboot.

### DIFF
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -488,7 +488,7 @@ func (t *target) remediationStrategyExternal(r *ReconcileMachineHealthCheck) err
 			&t.Machine,
 			corev1.EventTypeWarning,
 			EventExternalAnnotationFailed,
-			"Requesting external of node associated with machine %v failed: %v",
+			"Requesting external remediation of node associated with machine %v failed: %v",
 			t.string(),
 			err,
 		)
@@ -498,7 +498,7 @@ func (t *target) remediationStrategyExternal(r *ReconcileMachineHealthCheck) err
 		&t.Machine,
 		corev1.EventTypeNormal,
 		EventExternalAnnotationAdded,
-		"Requesting external of node associated with machine %v",
+		"Requesting external remediation of node associated with machine %v",
 		t.string(),
 	)
 	return nil
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -28,7 +28,7 @@ import (
 
 const (
 	machineAnnotationKey          = "machine.openshift.io/machine"
-	machineExternalAnnotationKey  = "machine.openshift.io/external-remediation"
+	machineExternalAnnotationKey  = "host.metal3.io/external-remediation"
 	ownerControllerKind           = "MachineSet"
 	nodeMasterLabel               = "node-role.kubernetes.io/master"
 	machineRoleLabel              = "machine.openshift.io/cluster-api-machine-role"
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -28,14 +28,14 @@ import (
 
 const (
 	machineAnnotationKey          = "machine.openshift.io/machine"
-	machineRebootAnnotationKey    = "healthchecking.openshift.io/external-remediation"
+	machineExternalAnnotationKey  = "machine.openshift.io/external-remediation"
 	ownerControllerKind           = "MachineSet"
 	nodeMasterLabel               = "node-role.kubernetes.io/master"
 	machineRoleLabel              = "machine.openshift.io/cluster-api-machine-role"
 	machineMasterRole             = "master"
 	machinePhaseFailed            = "Failed"
 	remediationStrategyAnnotation = "healthchecking.openshift.io/strategy"
-	remediationStrategyReboot     = mapiv1.RemediationStrategyType("reboot")
+	remediationStrategyExternal   = mapiv1.RemediationStrategyType("external")
 	timeoutForMachineToHaveNode   = 10 * time.Minute
 	machineNodeNameIndex          = "machineNodeNameIndex"
 	controllerName                = "machinehealthcheck-controller"
@@ -48,7 +48,7 @@ const (
 	// machine was detected unhealthy
 	EventDetectedUnhealthy string = "DetectedUnhealthy"
 	// EventSkippedMaster is emitted in case an unhealthy node (or a machine
-	// associated with the node) has Master role and reboot remediation strategy
+	// associated with the node) has Master role and external remediation strategy
 	// is not enabled
 	EventSkippedMaster string = "SkippedMaster"
 	// EventMachineDeletionFailed is emitted in case remediation of a machine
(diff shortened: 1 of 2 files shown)


pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -28,7 +28,7 @@ import (
 
 const (
 	machineAnnotationKey          = "machine.openshift.io/machine"
-	machineRebootAnnotationKey    = "healthchecking.openshift.io/machine-remediation-reboot"
```
enxebre: can you also change `healthchecking.openshift.io` -> `machine.openshift.io`

enxebre: wouldn't you want to chance the variable name as well? and the func `remediationStrategyReboot` name and its events and logging?

beekhof: good point, I'll do that too

beekhof: also `healthchecking.openshift.io/strategy` -> `machine.openshift.io/remediation-strategy` ?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -471,24 +471,24 @@ func (t *target) remediate(r *ReconcileMachineHealthCheck) error {
...
+	glog.Infof("Machine %s has been unhealthy for too long, adding external annotation", t.Machine.Name)
+	t.Node.Annotations[machineExternalAnnotationKey] = ""
 	if err := r.client.Update(context.TODO(), t.Node); err != nil {
 		r.recorder.Eventf(
 			&t.Machine,
 			corev1.EventTypeWarning,
-			EventRebootAnnotationFailed,
-			"Requesting reboot of node associated with machine %v failed: %v",
+			EventExternalAnnotationFailed,
+			"Requesting external of node associated with machine %v failed: %v",
```
markmc: s/external of node/external remediation of node/

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -497,8 +497,8 @@ func (t *target) remediationStrategyReboot(r *ReconcileMachineHealthCheck) error
 	r.recorder.Eventf(
 		&t.Machine,
 		corev1.EventTypeNormal,
-		EventRebootAnnotationAdded,
-		"Requesting reboot of node associated with machine %v",
+		EventExternalAnnotationAdded,
+		"Requesting external of node associated with machine %v",
```
markmc: Same here

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -28,14 +28,14 @@ import (
 
 const (
 	machineAnnotationKey          = "machine.openshift.io/machine"
-	machineRebootAnnotationKey    = "healthchecking.openshift.io/machine-remediation-reboot"
+	machineExternalAnnotationKey  = "machine.openshift.io/external-remediation"
```
beekhof:
```suggestion
	machineExternalAnnotationKey  = "host.metal3.io/external-remediation"
```

beekhof:
/bugzilla refresh

@enxebre I was going to wait until the other PR merged and then rebase this one

bison:
I agree with the intent here, but now the strategy name remains generic (reboot -> external) while the annotation is provider-specific: `host.metal3.io/external-remediation`

Could we make the strategy name match more closely?

Also, I forget, is this still feature gated in a way that blocks upgrades? If not, even if it's still tech-preview, can we update it like this? If I update an existing cluster that uses the `reboot` strategy, suddenly that's not recognized and I get the "delete" strategy with no warning.

beekhof:
Rebased now that #475 has been merged.

> I agree with the intent here, but now the strategy name remains generic (reboot -> external) while the annotation is provider-specific: `host.metal3.io/external-remediation`
> 
> Could we make the strategy name match more closely?

Are we just talking about changing
``remediationStrategyExternal   = mapiv1.RemediationStrategyType("external”)``
to something like
``remediationStrategyExternal   = mapiv1.RemediationStrategyType(“baremetal-external”)``
to indicate that this only works for baremetal?

Otherwise, the challenge of using `machine.openshift.io/external-remediation` is that metal3 see's `openshift.io` in the annotation name and thinks "that's openshift specific, we shouldn't be implementing that".  

I don't think we can both have `machine.openshift.io/external-remediation` as a platform agnostic annotation name, _and_ push the implementation onto providers that need to run elsewhere.

> Also, I forget, is this still feature gated in a way that blocks upgrades? If not, even if it's still tech-preview, can we update it like this? If I update an existing cluster that uses the `reboot` strategy, suddenly that's not recognized and I get the "delete" strategy with no warning.

I don't believe it is feature gated in 4.3, but we've also never shipped anything that depended these annotations.  That does not exclude the possibility that customers have dug into the MAO internals and implemented something - but are annotations held to the same versioning standards as CRDs? I thought that was part of the attraction of using annotations when the "final" interface is still an open question.

beekhof:
Resolved goof when merging with #475
Remaining test failures seem to be from etcd-operator

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -28,14 +28,14 @@ import (
...
+	machineExternalAnnotationKey  = "host.metal3.io/external-remediation"
 	ownerControllerKind           = "MachineSet"
 	nodeMasterLabel               = "node-role.kubernetes.io/master"
 	machineRoleLabel              = "machine.openshift.io/cluster-api-machine-role"
 	machineMasterRole             = "master"
 	machinePhaseFailed            = "Failed"
-	remediationStrategyAnnotation = "healthchecking.openshift.io/strategy"
-	remediationStrategyReboot     = mapiv1.RemediationStrategyType("reboot")
+	remediationStrategyAnnotation = "machine.openshift.io/remediation-strategy"
+	remediationStrategyExternal   = mapiv1.RemediationStrategyType("external")
```
beekhof:
```suggestion
	remediationStrategyExternal   = mapiv1.RemediationStrategyType("external-baremetal")
```

bison:
> Are we just talking about changing
> remediationStrategyExternal = mapiv1.RemediationStrategyType("external”)
> to something like
> remediationStrategyExternal = mapiv1.RemediationStrategyType(“baremetal-external”)
> to indicate that this only works for baremetal?

Yeah, sorry, that wasn't very clear. More or less what you described, yeah. If different strategies result in provider specific annotations going onto Machine objects, I just think the `RemediationStrategyType` string should reflect the provider in some way. So `baremetal-external` or `metal3` or whatever would be fine.

> I don't believe it is feature gated in 4.3, but we've also never shipped anything that depended these annotations. That does not exclude the possibility that customers have dug into the MAO internals and implemented something - but are annotations held to the same versioning standards as CRDs? I thought that was part of the attraction of using annotations when the "final" interface is still an open question.

Fair enough. I think you're probably right. I mostly wanted to bring up that the current behavior for an unknown strategy is to just silently fallback to "delete the Machine", which I think is a separate issue. I'll file a bug on that.

beekhof:
@bison happy enough with the change to give lgtm?

beekhof:
@enxebre what do you think, can we target this for 4.4?

PR #480, opened 2020-01-26
Bug 1795037: Fix cloud-provider-config-reader role

mandre:
Due to a typo, the `cloud-provider-config-reader` role introduced in
[1] to allow CAPO to read from the `openshift-config/cloud-provider-config` configmap didn't work.

This commit fixes the typo.

[1] https://github.com/openshift/machine-api-operator/pull/473/

### DIFF
--- a/install/0000_30_machine-api-operator_09_rbac.yaml
+++ b/install/0000_30_machine-api-operator_09_rbac.yaml
@@ -384,7 +384,7 @@ rules:
   - apiGroups:
       - ""
     resources:
-      - configmap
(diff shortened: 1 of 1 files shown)


enxebre:
This is now being superseded by https://github.com/openshift/machine-api-operator/pull/471

enxebre:
This can be closed now since https://github.com/openshift/machine-api-operator/pull/471 got in and it gives looser perms for configmaps so any name referenced by the infra object is accessible.

PR #488, opened 2020-02-06
Bug 1733474: Add log message when draining the unreachable node

alexander-demichev:
Add log message when draining the unreachable node.

### DIFF
--- a/pkg/controller/machine/controller.go
+++ b/pkg/controller/machine/controller.go
@@ -347,6 +347,8 @@ func (r *ReconcileMachine) drainNode(machine *machinev1.Machine) error {
 	}
 
 	if nodeIsUnreachable(node) {
+		klog.Infof("%q: Node %q is unreachable, draining will wait %q seconds after pod is signalled for deletion and skip after it",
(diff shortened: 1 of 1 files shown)


pkg/controller/machine/controller.go
```diff
@@ -347,6 +347,7 @@ func (r *ReconcileMachine) drainNode(machine *machinev1.Machine) error {
 	}
 
 	if nodeIsUnreachable(node) {
+		klog.Infof("Node %q is unreachable, draining will wait for pod deletion for %q seconds and skip after it", node.Name, skipWaitForDeleteTimeoutSeconds)
```
enxebre: may be somehow clarify this is `skipWaitForDeleteTimeoutSeconds` starting when the pod is signalled for deletion, otherwise every time this is logged cause the impression there's `skipWaitForDeleteTimeoutSeconds` left to timeout

enxebre:
/lgtm
if it does for @bison and @michaelgugino

michaelgugino:
We can always add an event later if needed, this is good enough for now, it's not really info we should need all that much and would be in any must-gather thanks to logs.

PR #490, opened 2020-02-11
Bug 1800314: Check if instance exists before proceeding with deletion

alexander-demichev:
From BZ description:
Deleting the Node object in Kubernetes signals to the cluster that no pod on that node is running and that it is safe to release any storage or process locks that ensure two processes can't be running on different nodes with the same name or volumes.  The machine controller was deleting the node before the machine was fully terminated, which means that a stateful set controller would be able to launch two pods with the same name running on the cluster at the same time, which violates our cluster safety guarantees.

Fix is to wait for machine is confirmed shut down by cloud provider before deleting the Node object.

### DIFF
--- a/pkg/controller/machine/controller.go
+++ b/pkg/controller/machine/controller.go
@@ -226,6 +226,17 @@ func (r *ReconcileMachine) Reconcile(request reconcile.Request) (reconcile.Resul
 			}
 		}
 
+		instanceExists, err := r.actuator.Exists(ctx, m)
+		if err != nil {
+			klog.Errorf("%v: failed to check if machine exists: %v", machineName, err)
+			return reconcile.Result{}, err
+		}
+
+		if instanceExists {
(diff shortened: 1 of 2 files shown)


pkg/controller/machine/controller.go
```diff
@@ -226,6 +226,17 @@ func (r *ReconcileMachine) Reconcile(request reconcile.Request) (reconcile.Resul
...
 		}
 
+		instanceExists, err := r.actuator.Exists(ctx, m)
+		if err != nil {
+			klog.Errorf("%v: failed to check if machine exists: %v", machineName, err)
+			return reconcile.Result{}, err
+		}
+
+		if instanceExists {
+			klog.Errorf("%v: can't proceed deleting machine while cloud instance is being terminated, requeuing", machineName)
```
enxebre: shouldn't this be `v(3).info`?

enxebre:
/approve
PTAL @bison

enxebre:
@alexander-demichev can we please reference this PR from all the revendoring PRs once this merge and create a 4.3/4.2 back port

PR #491, opened 2020-02-11
Bug 1802179: use generic image placeholders for install

elmiko:
these placeholders will get replaced by the release-creation tooling[0].
this change is inspired by a previous commit[1] to the
machine-config-operator.

[0]: https://github.com/openshift/cluster-version-operator/blob/33079c7732b99394f78c4fecb8d7d277c87ef02c/docs/dev/operators.md#how-do-i-ensure-the-right-images-get-used-by-my-manifests
[1]: https://github.com/openshift/machine-config-operator/pull/750

### DIFF
--- a/install/0000_30_machine-api-operator_01_images.configmap.yaml
+++ b/install/0000_30_machine-api-operator_01_images.configmap.yaml
@@ -6,19 +6,19 @@ metadata:
 data:
   images.json: >
     {
-      "machineAPIOperator": "docker.io/openshift/origin-machine-api-operator:v4.0.0",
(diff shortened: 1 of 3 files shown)


elmiko:
@enxebre should i file a bz for this?

enxebre:
thanks! there's also https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_01_images.configmap.yaml that needs to be updated to match the placeholder so they can be replaced by the right image reference coming from the imageStream.
We'll also need a bugzilla created an referenced for this.

elmiko:
thanks Alberto, i'll fix up the other file and make a bz

enxebre:
/retest
/approve
@wking PTAL

PR #492, opened 2020-02-12
Bug 1803736: Revert "Vsphere provider"

alexander-demichev:
Reverts openshift/machine-api-operator#474

vSphere IPI was dropped from 4.4, therefore we don't want the controller to run in this release. We'll enable it back once master opens for 4.5


tux-o-matic:
What's the reason?

enxebre:
Rebased
/lgtm

PR #494, opened 2020-02-13
update deployment link on readme

elmiko:
in the dev section there are instructions for running on vanilla
kubernetes, this change fixes the broken link for the deployment
manifest step.


elmiko:
the only question i have is, do we need an extra step for the kube-rbac-proxy configmap since we've added that to the deployment pod?

elmiko:
thanks @bison, maybe i'll just make a followup to remove these instructions after talking with the group a little.

enxebre:
Agree with https://github.com/openshift/machine-api-operator/pull/494#pullrequestreview-359077051
/approve
/test e2e-aws-operator

PR #497, opened 2020-02-24
Fix make targets

markmc:
Couple of small fixes - removing obsolete check-pkg target and fixing running goimports target locally


enxebre:
Thanks @markmc! this needs rebase.

PR #502, opened 2020-03-02
BUG 1806438: Remove explicit securityContext and add granular securitycontextconstraints "use" permissions in machine-api-controllers clusterRole

enxebre:
Without the runlabel https://github.com/openshift/machine-api-operator/pull/496, we’ll run as a high user by default, no need to say run me as non root.

Otherwise when removing the runlevel completely for the `openshift-machine-api` namespace https://github.com/openshift/cluster-autoscaler-operator/pull/133 the kube controller manager complains with ```Error creating: pods "machine-api-operator-75c887884f-" is forbidden: unable to validate against any security context constraint: [spec.containers[0].securityContext.securityContext.runAsUser: Invalid value: 65534: must be in the ranges: [1000340000, 1000349999] spec.containers[1].securityContext.securityContext.runAsUser: Invalid value: 65534: must be in the ranges: [1000340000, 1000349999]]``` https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-autoscaler-operator/133/pull-ci-openshift-cluster-autoscaler-operator-master-e2e-aws/496/artifacts/e2e-aws/pods/openshift-kube-controller-manager_kube-controller-manager-ip-10-0-133-251.us-east-2.compute.internal_kube-controller-manager.log"

Only after this PR makes it to payload then https://github.com/openshift/cluster-autoscaler-operator/pull/133/files should go green as no pods within the `openshift-machine-api` namespace will set an invalid user.

### DIFF
--- a/install/0000_30_machine-api-operator_09_rbac.yaml
+++ b/install/0000_30_machine-api-operator_09_rbac.yaml
@@ -167,6 +167,16 @@ rules:
     - list
     - watch
 
+# the baremetal pod deployment uses hostNetwork, hostPort, and privileged
+  - apiGroups:
+      - security.openshift.io
+    resources:
+      - securitycontextconstraints
+    verbs:
+      - use
+    resourceNames:
+      - privileged
+
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
--- a/install/0000_30_machine-api-operator_11_deployment.yaml
+++ b/install/0000_30_machine-api-operator_11_deployment.yaml
@@ -72,9 +72,6 @@ spec:
       nodeSelector:
         node-role.kubernetes.io/master: ""
       restartPolicy: Always
-      securityContext:
-        runAsNonRoot: true
-        runAsUser: 65534
       tolerations:
       - key: "node-role.kubernetes.io/master"
         operator: "Exists"
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -194,13 +194,9 @@ func newPodTemplateSpec(config *OperatorConfig, features map[string]bool) *corev
 			},
 		},
 		Spec: corev1.PodSpec{
-			Containers:        containers,
-			PriorityClassName: "system-node-critical",
-			NodeSelector:      map[string]string{"node-role.kubernetes.io/master": ""},
-			SecurityContext: &corev1.PodSecurityContext{
-				RunAsNonRoot: pointer.BoolPtr(true),
-				RunAsUser:    pointer.Int64Ptr(65534),
-			},
+			Containers:         containers,
+			PriorityClassName:  "system-node-critical",
+			NodeSelector:       map[string]string{"node-role.kubernetes.io/master": ""},
 			ServiceAccountName: "machine-api-controllers",
 			Tolerations:        tolerations,
 		},


markmc:
@enxebre As per comment in openshift/cluster-autoscaler-operator#133 the baremetal pod deployment uses hostNetwork, hostPort, and privileged

/cc @sadasu @stbenjam @hardys

enxebre:
@markmc good catch, I believe adding the specifics SCC to the clusterRole used by the name: machine-api-controllers should be enough as per openshift/cluster-ingress-operator@1409f54
i.e

- apiGroups:
  - security.openshift.io
  resources:
  - securitycontextconstraints
  verbs:
  - use
  resourceNames:
  - hostnetwork
  - privileged
cc @Miciah @smarterclayton can you confirm?

/hold

smarterclayton:
Yes, although ONLY the component that needs those should have a reference to the SCC via a unique service account.

enxebre:
>Yes, although ONLY the component that needs those should have a reference to the SCC via a unique service account.

Agreed, currently our granularity let us give this to the pod which run the multiple machine controllers. In near future we might consider breaking this pod down into multiple components.

/hold cancel
/retest

sadasu:
@enxebre @markmc I would like to try these changes locally.

enxebre:
>@enxebre @markmc I would like to try these changes locally.

@sadasu  You could do by rebuilding a payload which includes this PR and https://github.com/openshift/cluster-autoscaler-operator/pull/133/files

markmc:
> > @enxebre @markmc I would like to try these changes locally.
> 
> @sadasu You could do by rebuilding a payload which includes this PR and https://github.com/openshift/cluster-autoscaler-operator/pull/133/files

Here's what I just tried:

1. Added CVO overrides for the openshift-machine-api namespace, the machine-api-controllers role, and the machine-api-operator and cluster-autoscaler-operator deployments
2. Scaled the four openshift-machine-api deployments to zero
3. Updated the openshift-machine-api namespace to remove the runlevel
4. Scaled up the metal3 deployment to confirm it's broken with `unable to validate against any security context constraint` and scaled it back to zero
5. Added RBAC for the machine-api-controllers Role (not ClusterRole) to use the privileged SCC (just privileged, since it allows hostNetwork)
6. Scaled metal3 deployment back up to confirm the pods now get admitted

Seems to work fine

install/0000_30_machine-api-operator_09_rbac.yaml
```diff
@@ -167,6 +167,17 @@ rules:
...
 
+# the baremetal pod deployment uses hostNetwork, hostPort, and privileged
+  - apiGroups:
+      - security.openshift.io
+    resources:
+      - securitycontextconstraints
+    verbs:
+      - use
+    resourceNames:
+      - hostnetwork
```
markmc: Just privileged should be enough since it allows hostNetwork and hostPorts

install/0000_30_machine-api-operator_09_rbac.yaml
```diff
@@ -167,6 +167,17 @@ rules:
     - list
     - watch
 
+# the baremetal pod deployment uses hostNetwork, hostPort, and privileged
```
markmc: Any reason to include it in the ClusterRole and not the Role? The latter worked fine for me

smarterclayton:
Is there a reason this isn't part of the other PR?  So we can verify them together?

enxebre:
>Is there a reason this isn't part of the other PR? So we can verify them together?

If you mean https://github.com/openshift/cluster-autoscaler-operator/pull/133/files note that is against a different repo cluster-autoscaler-operator which shares `openshift-machine-api` namespace

install/0000_30_machine-api-operator_09_rbac.yaml
enxebre: I opted for the clusterRole as scc are cluster scoped resources.

enxebre:
@markmc you good to lgtm?

enxebre:
@bison @JoelSpeed PTAL

markmc:
> @markmc you good to lgtm?

sure

/lgtm

pkg/operator/baremetal_pod.go
```diff
@@ -173,15 +173,12 @@ func newMetal3PodTemplateSpec(config *OperatorConfig, baremetalProvisioningConfi
...
 		Spec: corev1.PodSpec{
-			Volumes:           volumes,
-			InitContainers:    initContainers,
-			Containers:        containers,
-			HostNetwork:       true,
-			PriorityClassName: "system-node-critical",
-			NodeSelector:      map[string]string{"node-role.kubernetes.io/master": ""},
-			SecurityContext: &corev1.PodSecurityContext{
-				RunAsNonRoot: pointer.BoolPtr(false),
-			},
```
markmc: Oh wait, I missed this bit - why did you remove this?

enxebre: I thought this was delegated to whatever level of isolation give the SCC in use now, so no need to explicitly set this?

markmc: I haven't looked into this part at all, but I guess I understood that RunAsNonRoot: true is the default when we remove the run-level ... so we'd still explicitly need to explicitly set it where we want RunAsNonRoot: false

enxebre: I was reading `RunAsNonRoot: true` :/. I'll update

markmc:
/lgtm

Would be good to get feedback from @sadasu's testing too

sadasu:
/approve
I was able to patch in the latest fix for this PR with the fix for https://github.com/openshift/cluster-autoscaler-operator/pull/133 by e2e baremetal (virtual) deployment. Happy to report that the metal3 pod came up fine and so did all the workers.

install/0000_30_machine-api-operator_09_rbac.yaml
```diff
@@ -167,6 +167,16 @@ rules:
...
+  - apiGroups:
+      - security.openshift.io
+    resources:
+      - securitycontextconstraints
+    verbs:
+      - use
+    resourceNames:
+      - privileged
```
stlaz: why does the operator require privileged SCC to run?

stlaz:
/hold
> [spec.containers[0].securityContext.securityContext.runAsUser: Invalid value: 65534: must be in the ranges: [1000340000, 1000349999] spec.containers[1].securityContext.securityContext.runAsUser: Invalid value: 65534: must be in the ranges: [1000340000, 1000349999]]
if this was the only problem then just the removal of `runAsUser` should do

PR #503, opened 2020-03-02
Bug 1809333: ROKS - remove machine related CRDs from cluster

csrwng:
In a ROKS cluster, the console UI should not show machine related views since
machine management is done outside the cluster. The UI will remove the
views as long as the CRDs are not present.

### DIFF
--- a/hack/gen-crd.sh
+++ b/hack/gen-crd.sh
@@ -2,6 +2,15 @@
 
 set -eu
 
+function annotate_crd() {
+  script='/^metadata:/a\
+\ \ annotations:\
+\ \ \ \ exclude.release.openshift.io/internal-opernshift-hosted: "true"'
+  input="${1}"
+  output="${2}"
+  sed -e "${script}" "${input}" > "${output}"
+}
+
 echo "Building controller-gen tool..."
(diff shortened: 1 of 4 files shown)


enxebre:
mm @csrwng is unfortunate that this would be overridden every time the CRDs are autogenerated. How are others solving that problem?

csrwng:
@enxebre machine related CRDs are the only ones we need to omit for ROKS. Can we make it part of the CRD generation script? (https://github.com/openshift/machine-api-operator/blob/master/hack/gen-crd.sh)

csrwng:
@enxebre I updated the gen-crd script to add the annotation for the hosted control plane. If this is acceptable, I'll do the same for the autoscaler PR.

enxebre:
/approve
PTAL @bison @JoelSpeed

csrwng:
Thanks @JoelSpeed. Submitting an alternate implementation that doesn't alter the encoding of the resulting file. It relies on generated CRDs not having any annotations, which I think for now is safe to assume. I don't believe the generate script is used in CI. The checked-in CRD files is what is used everywhere.

enxebre:
We don't validate on CI atm, we should though. I just created OCPCLOUD-809 Jira ticket to track this.

Issue #522, opened 2020-03-17
Precedence in Cloud Identification


srirammageswarank:
The machine api now is getting the cloud it is running in infrastructure object. 

https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/operator.go#L253-L258

https://github.com/openshift/machine-api-operator/blob/94398d11715b820f0bd79377b3cc702754f2ea02/pkg/operator/config.go#L63-L68

Can it be configured to be read from a config map and default to the infrastructure Object .?

This would be help full in bare metal install in any cloud which wants to use machines and machinesets.?

PR #525, opened 2020-03-18
Precedence in Cloud Identification

srirammageswarank:
The machine api now is getting the cloud it is running in infrastructure object.

The change is about setting a precedence for the with a config map to support for bare metal installation on any cloud.

When the operator finds the config map it will  load the cloud from the config map and default to infrastructure object when there is an error in reading the configmap or the configmap not configured properly.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -260,11 +260,11 @@ func (optr *Operator) maoConfigFromInfrastructure() (*OperatorConfig, error) {
 		return nil, err
 	}
 
-	cloudapiconfig, err := optr.kubeClient.CoreV1().ConfigMaps(optr.namespace).Get("cluster-api-provider-config", metav1.GetOptions{})
+	machineapiconfig, err := optr.kubeClient.CoreV1().ConfigMaps(optr.namespace).Get("machine-api-controller-config", metav1.GetOptions{})
 	if err != nil {
-		glog.V(4).Infof("Error in reading config map cluster-api-provider-config, %v", err)
+		glog.V(4).Infof("Error in reading config map machine-api-controller-config, %v", err)
 	} else {
-		provider, err = getProviderFromConfigMap(cloudapiconfig)
+		provider, err = getProviderFromConfigMap(machineapiconfig)
 		if err != nil {
 			glog.V(4).Infof("Unable able to find cloud in configmap, %v", err)
 		}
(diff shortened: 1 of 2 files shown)


enxebre:
hey @srirammageswarank could you share more context about your particular use case and the scenario where you'd see this?

srirammageswarank:
Sure @enxebre , For Openshift bare metal installs on any cloud the infrastructure object is being set to None. The operator looks up this value and does not load any machine-api-controller operator. 

The change is to override that using an config map, If the config map is not found the operator will default to the infrastructure Object. IAlso, f found will take precedence and resolve the type of cloud from the config map So the platform can use the machine controller even if deployed as baremetal.

srirammageswarank:
#522 is the issue.

enxebre:
thanks @srirammageswarank! could you elaborate on what you mean by "bare metal installs on any cloud"? we do support [metal3](https://github.com/metal3-io) baremetal integration
https://github.com/openshift/machine-api-operator/blob/a3345f5a437a8ca82be683c7c38fe49de64c2b5f/pkg/operator/operator.go#L268-L273
https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/sync.go#L44

srirammageswarank:
Hi @enxebre , Thanks for the info.  By bare metal install on AWS i mean setting the below values in install-config.yaml 

platform:
  none: {}

pkg/operator/operator.go
```diff
@@ -260,6 +260,16 @@ func (optr *Operator) maoConfigFromInfrastructure() (*OperatorConfig, error) {
 		return nil, err
 	}
 
+	cloudapiconfig, err := optr.kubeClient.CoreV1().ConfigMaps(optr.namespace).Get("cluster-api-provider-config", metav1.GetOptions{})
```
michaelgugino: We don't really use 'cluster-api' for object names, we use 'machine-api' in OpenShift.

srirammageswarank: Hi @michaelgugino , Thanks for the review. I corrected it.

srirammageswarank:
No @michaelgugino , It is not related to any other work. And , yes the cloud credentials operator is being deployed. But the cloud credentials operator will not be able to detect the type of cloud as the infrastructure object will be platform status : None

srirammageswarank:
Hi @michaelgugino , 

Wanted to follow up on this ? Does the change look good ? 

Thanks 
@srirammageswarank

PR #530, opened 2020-03-20
Bug 1811161: ROKS - remove machine related CRDs from cluster [4.4]

csrwng:
In a ROKS cluster, the console UI should not show machine related views since
machine management is done outside the cluster. The UI will remove the
views as long as the CRDs are not present.

### DIFF
--- a/hack/gen-crd.sh
+++ b/hack/gen-crd.sh
@@ -2,6 +2,15 @@
 
 set -eu
 
+function annotate_crd() {
+  script='/^metadata:/a\
+\ \ annotations:\
+\ \ \ \ exclude.release.openshift.io/internal-openshift-hosted: "true"'
+  input="${1}"
+  output="${2}"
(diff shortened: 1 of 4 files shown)


hack/gen-crd.sh
```diff
@@ -2,6 +2,15 @@
 
 set -eu
 
+function annotate_crd() {
```
michaelgugino: Is there a way to do this in go code rather than here?

csrwng: You mean a go program to invoke the original gen tool?

csrwng: Definitely, if there's a better way to do it, then we should create an issue for it. However, this already went in to master and all we're trying to do is backport it to 4.4 and 4.3.

enxebre: yeh there's OCPCLOUD-809 to follow up on this separately. We should probably use kustomize or similar for the long term.

PR #537, opened 2020-03-24
vsphere ipi: add extra config key with guestinfo hostname

jcpowermac:
To support setting the hostname of the RHCOS guest
add a extra config key/value `guestinfo.hostname`.

### DIFF
--- a/pkg/controller/vsphere/reconciler.go
+++ b/pkg/controller/vsphere/reconciler.go
@@ -36,6 +36,7 @@ const (
 const (
 	GuestInfoIgnitionData     = "guestinfo.ignition.config.data"
 	GuestInfoIgnitionEncoding = "guestinfo.ignition.config.data.encoding"
+	GuestInfoHostname         = "guestinfo.hostname"
 )
 
 // Reconciler runs the logic to reconciles a machine resource towards its desired state
@@ -389,6 +390,14 @@ func clone(s *machineScope) error {
 	var deviceSpecs []types.BaseVirtualDeviceConfigSpec
 	deviceSpecs = append(deviceSpecs, networkDevices...)
 
+	extraConfig := []types.BaseOptionValue{}
+
+	extraConfig = append(extraConfig, IgnitionConfig(userData)...)
+	extraConfig = append(extraConfig, &types.OptionValue{
+		Key:   GuestInfoHostname,
+		Value: s.machine.GetName(),
+	})
+
 	spec := types.VirtualMachineCloneSpec{
 		Config: &types.VirtualMachineConfigSpec{
 			Annotation: s.machine.GetName(),
@@ -397,7 +406,7 @@ func clone(s *machineScope) error {
 			// the VM's UUID.
 			InstanceUuid: string(s.machine.UID),
 			Flags:        newVMFlagInfo(),
-			ExtraConfig:  IgnitionConfig(userData),
+			ExtraConfig:  extraConfig,
 			// TODO: set disk devices
 			DeviceChange:      deviceSpecs,
 			NumCPUs:           numCPUs,


pkg/controller/vsphere/reconciler.go
```diff
@@ -389,6 +390,13 @@ func clone(s *machineScope) error {
 	var deviceSpecs []types.BaseVirtualDeviceConfigSpec
 	deviceSpecs = append(deviceSpecs, networkDevices...)
 
+	extraConfig := IgnitionConfig(userData)
+
+	extraConfig = append(extraConfig, &types.OptionValue{
+		Key:   GuestInfoHostname,
+		Value: s.machine.GetName(),
+	})
+
```
abhinavdahiya: personally i think this reads misleading and it would be better to write it like,
```suggestion
	extraConfig := []types.BaseOptionValue{}
	extranConfig = append(extraConfig, IgnitionConfig(userData)...)
	extraConfig = append(extraConfig, &types.OptionValue{
		Key:   GuestInfoHostname,
		Value: s.machine.GetName(),
	})

```

this makes it clear that we are collecting the list of options, instead of previously when using the `IgnitionConfig` function it seems like extraConfig is related to ignition configs...

jcpowermac: Thanks @abhinavdahiya updated

pkg/controller/vsphere/reconciler.go
```diff
@@ -389,6 +390,14 @@ func clone(s *machineScope) error {
 	var deviceSpecs []types.BaseVirtualDeviceConfigSpec
 	deviceSpecs = append(deviceSpecs, networkDevices...)
 
+	extraConfig := []types.BaseOptionValue{}
+
+	extraConfig = append(extraConfig, IgnitionConfig(userData)...)
+	extraConfig = append(extraConfig, &types.OptionValue{
+		Key:   GuestInfoHostname,
+		Value: s.machine.GetName(),
```
enxebre: why is this not the same than https://github.com/openshift/machine-api-operator/commit/eb9152c368e62bed358f2aacc3e5d67707361a7d#diff-11b6610f82e5d47fdfb8dda44b90dd1eR283

pkg/controller/vsphere/reconciler.go
```diff
@@ -389,6 +390,14 @@ func clone(s *machineScope) error {
 	var deviceSpecs []types.BaseVirtualDeviceConfigSpec
 	deviceSpecs = append(deviceSpecs, networkDevices...)
 
+	extraConfig := []types.BaseOptionValue{}
+
+	extraConfig = append(extraConfig, IgnitionConfig(userData)...)
```
michaelgugino: While it shouldn't happen because every machine will have user data, this might attempt to unpack a nil, which I think would cause the reconciler to crash.

pkg/controller/vsphere/reconciler.go
jcpowermac: @enxebre the vm doesn't exist yet.  Is that not the name of the machine that will be created?

pkg/controller/vsphere/reconciler.go
abhinavdahiya: hmm.. based on https://play.golang.org/p/GVxF2yAVWKE
i don't this this should crash @michaelgugino  ?

michaelgugino: You are right, I tried it out myself.  Go is a hacky language.

pkg/controller/vsphere/reconciler.go
enxebre: right, It's passed here so should be https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/vsphere/reconciler.go#L425
It'd be nice to use the same reference consistently everywhere though.

enxebre:
/approve
@alexander-demichev PTAL

PR #547, opened 2020-04-02
add a podman option to Makefile

elmiko:
This change allows a user to set `USE_PODMAN=1` when running the make
command to force usage of podman instead of docker during the build and
image targets.

Also adds a note to the README about the `NO_DOCKER` and `USE_PODMAN`
options.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,13 +19,19 @@ endif
 .PHONY: all
 all: check build test
 
+USE_PODMAN ?= 0
 NO_DOCKER ?= 0
 ifeq ($(NO_DOCKER), 1)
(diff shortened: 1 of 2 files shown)


elmiko:
this is something i've run into several times, i thought it would be nice to have an option for systems that don't have docker available.

i intentionally kept this pr simple, i had thought about adding more automation to detect the presence of podman or absence of docker, but i think this solution is clear and easy to use while preserving the default behavior.

elmiko:
i'm gonna close this and just keep my local changes for podman, perhaps in the future it will make sense to add something like this.

PR #560, opened 2020-04-17
Bug 1840782: Update conditions under which metal3 pod would be deployed

sadasu:
Changes to accommodate Baremetal UPI deployments or assisted baremetal deployments where the platform type would be Baremetal without a need to deploy the metal3 pod.

### DIFF
--- a/pkg/operator/baremetal_config.go
+++ b/pkg/operator/baremetal_config.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net"
 
-	"github.com/golang/glog"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -14,6 +13,7 @@ import (
 )
 
 var provisioningGVR = schema.GroupVersionResource{Group: "metal3.io", Resource: "provisionings", Version: "v1alpha1"}
+var provisioningGR = schema.GroupResource{Group: "metal3.io", Resource: "provisionings"}
 
 const (
 	baremetalProvisioningCR        = "provisioning-configuration"
@@ -48,47 +48,47 @@ func reportError(found bool, err error, configItem string, configName string) er
 	return fmt.Errorf("Unknown Error while reading %s from Baremetal provisioning CR %s", configItem, configName)
 }
 
-func getBaremetalProvisioningConfig(dc dynamic.Interface, configName string) (BaremetalProvisioningConfig, error) {
+func getBaremetalProvisioningConfig(dc dynamic.Interface, configName string) (*BaremetalProvisioningConfig, error) {
 	provisioningClient := dc.Resource(provisioningGVR)
 	provisioningConfig, err := provisioningClient.Get(context.Background(), configName, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
-		glog.V(3).Infof("Baremetal provisioning CR %s is not found", configName)
-		return BaremetalProvisioningConfig{}, nil
+		// The provisioning CR is not present and that is not considered an error.
+		return nil, nil
 	}
 	if err != nil {
-		return BaremetalProvisioningConfig{}, reportError(true, err, "provisioning configuration", configName)
+		return nil, reportError(true, err, "provisioning configuration", configName)
 	}
 	provisioningSpec, found, err := unstructured.NestedMap(provisioningConfig.UnstructuredContent(), "spec")
 	if !found || err != nil {
-		return BaremetalProvisioningConfig{}, reportError(found, err, "Spec field", configName)
+		return nil, reportError(found, err, "Spec field", configName)
 	}
 	provisioningInterface, found, err := unstructured.NestedString(provisioningSpec, "provisioningInterface")
 	if !found || err != nil {
-		return BaremetalProvisioningConfig{}, reportError(found, err, "provisioningInterface", configName)
+		return nil, reportError(found, err, "provisioningInterface", configName)
 	}
 	provisioningIP, found, err := unstructured.NestedString(provisioningSpec, "provisioningIP")
 	if !found || err != nil {
-		return BaremetalProvisioningConfig{}, reportError(found, err, "provisioningIP", configName)
+		return nil, reportError(found, err, "provisioningIP", configName)
 	}
 	provisioningDHCPRange, found, err := unstructured.NestedString(provisioningSpec, "provisioningDHCPRange")
 	if !found || err != nil {
-		return BaremetalProvisioningConfig{}, reportError(found, err, "provisioningDHCPRange", configName)
+		return nil, reportError(found, err, "provisioningDHCPRange", configName)
 	}
 	provisioningOSDownloadURL, found, err := unstructured.NestedString(provisioningSpec, "provisioningOSDownloadURL")
 	if !found || err != nil {
-		return BaremetalProvisioningConfig{}, reportError(found, err, "provisioningOSDownloadURL", configName)
+		return nil, reportError(found, err, "provisioningOSDownloadURL", configName)
 	}
 	// If provisioningNetwork is not provided, set its value based on provisioningDHCPExternal
 	provisioningNetwork, foundNetworkState, err := unstructured.NestedString(provisioningSpec, "provisioningNetwork")
 	if err != nil {
-		return BaremetalProvisioningConfig{}, reportError(true, err, "provisioningNetwork", configName)
+		return nil, reportError(true, err, "provisioningNetwork", configName)
 	}
 	if !foundNetworkState {
 		// Check if provisioningDHCPExternal is present in the config
 		provisioningDHCPExternal, foundDHCP, err := unstructured.NestedBool(provisioningSpec, "provisioningDHCPExternal")
 		if !foundDHCP || err != nil {
 			// Both the new provisioningNetwork and the old provisioningDHCPExternal configs are not found.
-			return BaremetalProvisioningConfig{}, reportError(foundDHCP, err, "provisioningNetwork and provisioningDHCPExternal", configName)
+			return nil, reportError(foundDHCP, err, "provisioningNetwork and provisioningDHCPExternal", configName)
 		}
 		if !provisioningDHCPExternal {
 			provisioningNetwork = provisioningNetworkManaged
@@ -100,23 +100,24 @@ func getBaremetalProvisioningConfig(dc dynamic.Interface, configName string) (Ba
 	// The CIDR of the network needs to be extracted to form the provisioningIPCIDR
 	provisioningNetworkCIDR, found, err := unstructured.NestedString(provisioningSpec, "provisioningNetworkCIDR")
 	if !found || err != nil {
-		return BaremetalProvisioningConfig{}, reportError(found, err, "provisioningNetworkCIDR", configName)
+		return nil, reportError(found, err, "provisioningNetworkCIDR", configName)
 	}
 	// Check if the other config values make sense for the provisioningNetwork configured.
 	if provisioningInterface == "" && provisioningNetwork == provisioningNetworkManaged {
-		return BaremetalProvisioningConfig{}, fmt.Errorf("provisioningInterface cannot be empty when provisioningNetwork is Managed.")
+		return nil, fmt.Errorf("provisioningInterface cannot be empty when provisioningNetwork is Managed.")
 	}
 	if provisioningDHCPRange == "" && provisioningNetwork == provisioningNetworkManaged {
-		return BaremetalProvisioningConfig{}, fmt.Errorf("provisioningDHCPRange cannot be empty when provisioningNetwork is Managed or when the DHCP server needs to run with the metal3 cluster.")
+		return nil, fmt.Errorf("provisioningDHCPRange cannot be empty when provisioningNetwork is Managed or when the DHCP server needs to run with the metal3 cluster.")
 	}
-	return BaremetalProvisioningConfig{
+	baremetalConfig := BaremetalProvisioningConfig{
 		ProvisioningInterface:     provisioningInterface,
 		ProvisioningIp:            provisioningIP,
 		ProvisioningNetworkCIDR:   provisioningNetworkCIDR,
 		ProvisioningDHCPRange:     provisioningDHCPRange,
 		ProvisioningOSDownloadURL: provisioningOSDownloadURL,
 		ProvisioningNetwork:       provisioningNetwork,
-	}, nil
+	}
+	return &baremetalConfig, nil
 }
 
 func getProvisioningIPCIDR(baremetalConfig BaremetalProvisioningConfig) *string {
--- a/pkg/operator/baremetal_pod.go
+++ b/pkg/operator/baremetal_pod.go
@@ -18,7 +18,6 @@ import (
 )
 
 const (
-	baremetalConfigmap         = "metal3-config"
 	baremetalSharedVolume      = "metal3-shared"
 	baremetalSecretName        = "metal3-mariadb-password"
 	baremetalSecretKey         = "password"
--- a/pkg/operator/baremetal_test.go
+++ b/pkg/operator/baremetal_test.go
@@ -5,6 +5,7 @@ import (
 
 	. "github.com/onsi/gomega"
 	"golang.org/x/net/context"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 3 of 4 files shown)


sadasu:
@hardys and @yevgeny-shnaidman could this be a way to handle the other baremetal mode that you were discussing here https://github.com/openshift/enhancements/pull/212/files#r409393934

yevgeny-shnaidman:
Yes, it might.
One suggestion: if it is not too late, maybe you can move the call to
isBaremetalProvisioningConfigAvailable
 from  syncAll to maoConfigFromInfrastructure, to where the value of
usingBareMetal is set. It can be something like:
usingBareMetal := (provider == osconfigv1.BareMetalPlatformType &&
isBaremetalProvisioningConfigAvailable())
then you don't need to change syncAll and syncBaremetalControllers functions
at all.

pkg/operator/baremetal_config.go
```diff
@@ -34,6 +34,15 @@ type BaremetalProvisioningConfig struct {
 	ProvisioningOSDownloadURL string
 }
 
+func isBaremetalProvisioningConfigAvailable(dc dynamic.Interface, configName string) bool {
+	provisioningClient := dc.Resource(provisioningGVR)
+	_, err := provisioningClient.Get(context.Background(), configName, metav1.GetOptions{})
+	if err != nil {
```
JoelSpeed, line 40: In a baremetal cluster, will the provisioning CRDs always be deployed, or will they only be deployed in the case that Metal3 is desired? Just wondering if we should distinguish the behaviour of, for example, a failed connection vs the GVR missing from the cluster vs the configmap not existing

pkg/operator/sync.go
```diff
@@ -52,9 +52,12 @@ func (optr *Operator) syncAll(config *OperatorConfig) error {
 	}
 	glog.V(3).Info("Synced up all machine-api-controller components")
 
-	// In addition, if the Provider is BareMetal, then bring up
-	// the baremetal-operator pod
+	// In addition, if the Provider is BareMetal, then imay have to deploy metal3 pod.
```
JoelSpeed, line 55: Might be better to include the condition rather than saying "may have to"
```suggestion
	// In addition, if the Provider is BareMetal and a provisioning config is available, deploy a metal3 pod.
```

pkg/operator/sync.go
```diff
@@ -52,9 +52,12 @@ func (optr *Operator) syncAll(config *OperatorConfig) error {
...
 	if config.BaremetalControllers.BaremetalOperator != "" {
+		if !isBaremetalProvisioningConfigAvailable(optr.dynamicClient, baremetalProvisioningCR) {
+			glog.V(3).Info("Not necessary to deploy metal3 pod for this baremetal install.")
```
JoelSpeed, line 58: This could be misleading if there are transient connection issues, perhaps this should be reworded to say something more similar to
```suggestion
			glog.V(3).Info("Unable to read Baremetal Provisioning config, skipping Metal3 install.")
```

sadasu:
> Yes, it might. One suggestion: if it is not too late, maybe you can move the call to isBaremetalProvisioningConfigAvailable from syncAll to maoConfigFromInfrastructure, to where the value of usingBareMetal is set. It can be something like: usingBareMetal := (provider == osconfigv1.BareMetalPlatformType && isBaremetalProvisioningConfigAvailable()) then you don't need to change syncAll and syncBaremetalControllers functions at all.

@yevgeny-shnaidman yes, we can improve the implementation. I want confirmation that you are Ok with the approach that I am suggesting here i.e not creating the provisioning CR when we do not want metal3 to be deployed. If that is OK, then we need a corresponding change in the installer to not create the provisioning CR for this other baremetal mode (UPI ?). There we will need a configuration flag to distinguish between the 2 baremetal installs. cc @hardys @stbenjam  @imain

yevgeny-shnaidman:
i am OK with this approach. Regarding installer change - since this is a
temporary solution until BMO becomes full-fledge operator, we will be able
to remove the provisioning CR from bootstrap ignition after the installer
has created it, so we won't need any change in the installer code

pkg/operator/baremetal_config.go
sadasu: @JoelSpeed Currently, provisioning CR is always deployed when the platform is baremetal. With the need to support this new BareMetal installation type where the platform type is Baremetal but there is no need for the metal3 pod to manage the cluster, we would not need the provisioning CR.
With this change, within the MAO, we will not be able to distinguish between the cases where there was some failure deploying the provisioning CR or if we do not need the CR at all. And, that is OK I think. 
I am not sure I understand what you mean by "failed connection vs the GVR missing". Regarding the config map, since we are blocking metal3 deployment from happening without the provisioning config, the only instance where the configmap would be the case when the provisioning CR exists but is missing some values.

pkg/operator/sync.go
sadasu: Done.

pkg/operator/sync.go
sadasu: Refactored code so do not need this Info message. I think with the current refactor, even if there are transient connection issues, the sync operation would re-try deployment and eventually succeed but without these messages in the log.

sadasu:
> Yes, it might. One suggestion: if it is not too late, maybe you can move the call to isBaremetalProvisioningConfigAvailable from syncAll to maoConfigFromInfrastructure, to where the value of usingBareMetal is set. It can be something like: usingBareMetal := (provider == osconfigv1.BareMetalPlatformType && isBaremetalProvisioningConfigAvailable()) then you don't need to change syncAll and syncBaremetalControllers functions at all.

@yevgeny-shnaidman I would like to keep the check for the presence of the provisioning CR within the sync logic because it is possible that we are unable to read the CR due to some transient connectivity issues. We want to keep retrying to read that. With the current implementation we will not be flooding the logs with info messages when the provisioning CR is not really present.

sadasu:
Would like to point out that this fix is not temporary. This same logic would be carried forward to the new operator when it is ready.

yevgeny-shnaidman:
that's ok, because eventually we would like to use baremetal operator, only
with our provisioner. Not using it at all is for the intermediate stage of
our product

romfreiman:
@sadasu any eta on this commit?

sadasu:
@enxebre and @JoelSpeed could you please take a look?

romfreiman:
@enxebre any update when this pr will be merged?
In the assisted flow it's a requirement so we will not run ironic on day1.

hardys:
@sadasu @romfreiman this now needs a bz to get bugzilla/valid-bug before we can chase for reviews again - do we have an existing bz that can be added to the title?

sadasu:
> @sadasu @romfreiman this now needs a bz to get bugzilla/valid-bug before we can chase for reviews again - do we have an existing bz that can be added to the title?

@hardys Thanks for pointing that out. I am working on creating the BZ. Will update title as soon as I am done.

pkg/operator/baremetal_config.go
enxebre: >With this change, within the MAO, we will not be able to distinguish between the cases where there was some failure deploying the provisioning CR or if we do not need the CR at all. And, that is OK I think

Why is that ok? here https://github.com/openshift/machine-api-operator/pull/560/files#diff-fa45321336db7ad1cedc28bf643a4f97R60 you are using this function to decide if the pod needs to be deployed. But the function as it is, could be lying to you. Unless I'm missing anything we need to check `if err != nil  and errIsNotFound` here

sadasu: When there are intermittent failures reading the provisioning CR, then the call to isBaremetalProvisioningConfigAvailable() will succeed in a subsequent call to sync(). And that would result in the metal3 pod being deployed. 
If the provisioning CR is really not present, then in every sync() call, the decision would be made not to deploy the metal3 pod.
That is what I tried to convey in my previous comment.

enxebre: I see that it accidentally would work, that is brittle. You should account for the error semantics that are relevant here (notFound vs other failures) and let the callers react appropriately.

Also I see there's also already a func `getBaremetalProvisioningConfig` which returns an error from `provisioningClient.Get(context.Background(), configName, metav1.GetOptions{})` so I see no need for a new func, you can reuse that and react to the notFound error.

Also this code as it is is completely dropping the ability to fall back to the configMap if the ProvisioningConfig CRD does not exists so can we  please make sure to remove all that dead code?

sadasu: I added isBaremetalProvisioningConfigAvailable() as a helper function because the assisted-install team wanted a silent check and getBaremetalProvisioningConfig() is not silent when it is not able to read provisioning config. Based on previous comment, I am removing the helper function and updating getBaremetalProvisioningConfig() to not report anything when it is not able to read the provisioning CR. More code reuse for sure with slight loss of debuggability.

Also, removing any code trying to read config from a configmap as suggested above since we would not hit that scenario at all with this change.

pkg/operator/sync.go
```diff
@@ -54,10 +54,10 @@ func (optr *Operator) syncAll(config *OperatorConfig) error {
 	}
 	glog.V(3).Info("Synced up all machine-api-controller components")
 
-	// In addition, if the Provider is BareMetal, then bring up
-	// the baremetal-operator pod
-	if config.BaremetalControllers.BaremetalOperator != "" {
-		if err := optr.syncBaremetalControllers(config); err != nil {
+	// In addition, if the Provider is BareMetal and a provisioning config is available,
+	// then bring up the baremetal-operator pod
+	if baremetalProvisioningConfig, err := getBaremetalProvisioningConfig(optr.dynamicClient, baremetalProvisioningCR); err != nil && apierrors.IsNotFound(err) && config.BaremetalControllers.BaremetalOperator != "" {
```
enxebre, line 59: so this would only run `syncBaremetalControllers` when the `baremetalProvisioningConfig` is not found? is that the intention? I thought it was the other way around.

enxebre: There's already a place where you are making this decission -> https://github.com/openshift/machine-api-operator/blob/a52b81d08668851f8faf8db37b38d93fdb15ec7b/pkg/operator/operator.go#L289

Any farther validation should be put in the same place e.g `func usingBareMetal`.

sadasu: I don't like the check I have right now. Please look at the latest version. 

> There's already a place where you are making this decision ->
> 
> https://github.com/openshift/machine-api-operator/blob/a52b81d08668851f8faf8db37b38d93fdb15ec7b/pkg/operator/operator.go#L289
> 
> Any farther validation should be put in the same place e.g `func usingBareMetal`, then you can leave this logic as it is now.

The reasoning for adding the check in the sync is to take care of situations where the Provisioning CR is temporarily unavailable.

Also, there is an additional requirement to provide for a way to opt-in to have a metal3 deployment as part of day-2 operations. So, adding the Provisioning CR at any (later) time would allow the metal3 deployment to proceed.

pkg/operator/baremetal_pod.go
```diff
@@ -37,29 +37,11 @@ var volumeMounts = []corev1.VolumeMount{
...
-	if value != nil {
-		return corev1.EnvVar{
-			Name:  name,
-			Value: *value,
-		}
-	} else {
-		return buildEnvVarFromConfigMap(name, key)
+	return corev1.EnvVar{
+		Name:  name,
+		Value: *value,
```
enxebre, line 44: wouldn't this panic if value is nil?

pkg/operator/baremetal_config.go
```diff
@@ -37,8 +38,12 @@ type BaremetalProvisioningConfig struct {
 func getBaremetalProvisioningConfig(dc dynamic.Interface, configName string) (BaremetalProvisioningConfig, error) {
 	provisioningClient := dc.Resource(provisioningGVR)
 	provisioningConfig, err := provisioningClient.Get(context.Background(), configName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		glog.Infof("Baremetal provisioning CR %s is not found", configName)
+		return BaremetalProvisioningConfig{}, nil
+	}
 	if err != nil {
-		glog.Errorf("Error getting config from Baremetal provisioning CR %s", configName)
+		glog.Infof("Baremetal config not available from Baremetal provisioning CR %s", configName)
```
enxebre, line 46: this should log Errorf

pkg/operator/baremetal_config.go
```diff
@@ -37,8 +38,12 @@ type BaremetalProvisioningConfig struct {
 func getBaremetalProvisioningConfig(dc dynamic.Interface, configName string) (BaremetalProvisioningConfig, error) {
 	provisioningClient := dc.Resource(provisioningGVR)
 	provisioningConfig, err := provisioningClient.Get(context.Background(), configName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		glog.Infof("Baremetal provisioning CR %s is not found", configName)
```
enxebre, line 42: can this log in v(3)

pkg/operator/sync.go
```diff
@@ -116,13 +116,18 @@ func (optr *Operator) syncTerminationHandler(config *OperatorConfig) error {
 	return nil
 }
 
-func (optr *Operator) syncBaremetalControllers(config *OperatorConfig) error {
-	// Try to get baremetal provisioning config from a CR
-	baremetalProvisioningConfig, err := getBaremetalProvisioningConfig(optr.dynamicClient, baremetalProvisioningCR)
-	if err != nil {
-		glog.Errorf("Unable to read Baremetal Provisioning config from CR %s.", baremetalProvisioningCR)
-		glog.Infof("Will try to read Baremetal Provisioning config from ConfigMap %s instead", baremetalConfigmap)
+func (optr *Operator) checkIfBaremetalAndGetConfig(config *OperatorConfig) (BaremetalProvisioningConfig, bool) {
```
enxebre, line 119: this function is doing too many things and its not reused anywhere. This logic can just be inlined here https://github.com/openshift/machine-api-operator/pull/560/files#diff-fa45321336db7ad1cedc28bf643a4f97R59

pkg/operator/baremetal_config.go
sadasu: We are returning error as nil for this condition, so thought that Infof was better than Errorf. Will change it back.

sadasu: Done.

pkg/operator/baremetal_config.go
sadasu: Done.

pkg/operator/sync.go
sadasu: Done.

pkg/operator/baremetal_pod.go
sadasu: getBaremetalProvisioningConfig() will return BaremetalProvisioningConfig{} if it is unable to read any config value from the Provisioning CR. If we end up with BaremetalProvisioningConfig{}, metal3 deployment will not proceed. So, when get to this point, we will not have value of nil.

sadasu:
Design for this is undergoing some changes and is being targeted for the next release.

hardys:
@sadasu I think we may want to reopen this - there is $customer interest in using `platform: baremetal` in a UPI fashion so enable the hosted LoadBalancer and DNS components via the MCO, and this PR would enable doing that while disabling the provisioning components (e.g the metal3 pod).

Further changes will likely be needed in future to fully support this (e.g some way to control if the provisioning CR is created by the installer via the install-config) but as a first step removing the manifest is probably sufficient to enable testing.

pkg/operator/baremetal_pod.go
sadasu: Fixed.

pkg/operator/sync.go
```diff
@@ -195,6 +195,9 @@ func (optr *Operator) syncMutatingWebhook() error {
 func (optr *Operator) syncBaremetalControllers(config *OperatorConfig) error {
 	// Try to get baremetal provisioning config from a CR
 	baremetalProvisioningConfig, err := getBaremetalProvisioningConfig(optr.dynamicClient, baremetalProvisioningCR)
+	if err == nil && (BaremetalProvisioningConfig{}) == baremetalProvisioningConfig {
```
dhellmann, line 198: Comparing the empty struct feels a bit odd. How big of a change would it be to make `getBaremetalProvisioningConfig` return a pointer?

sadasu:
> Can we update the tests to add a test case that covers this so that we don't regress?

Done.

pkg/operator/baremetal_test.go
```diff
@@ -206,13 +206,13 @@ func TestGetMetal3DeploymentConfig(t *testing.T) {
 		t.Logf("Unstructed Config:  %+v", u)
 		t.Errorf("Failed to get Baremetal Provisioning Config from CR %s", testConfigResource)
 	}
-	actualCacheURL := getMetal3DeploymentConfig("CACHEURL", baremetalConfig)
+	actualCacheURL := getMetal3DeploymentConfig("CACHEURL", *baremetalConfig)
```
stbenjam, line 209: Is it a big deal to change the function to take a pointer? Dereferencing the pointer in every call is a bit expensive, since you'll cause an entire copy of the config to be passed each time, instead of a copy of the address.

sadasu: > Is it a big deal to change the function to take a pointer? Dereferencing the pointer in every call is a bit expensive, since you'll cause an entire copy of the config to be passed each time, instead of a copy of the address.

This is the unit test code.

michaelgugino, line 209: I realize this is just a test, but there is a possible null pointer deref here.

pkg/operator/sync.go
```diff
@@ -204,7 +212,7 @@ func (optr *Operator) syncBaremetalControllers(config *OperatorConfig) error {
 		return err
 	}
 
-	metal3Deployment := newMetal3Deployment(config, baremetalProvisioningConfig)
+	metal3Deployment := newMetal3Deployment(config, *baremetalProvisioningConfig)
```
michaelgugino, line 215: It doesn't make great sense to dereference the pointer here, maybe pass the address to newMetal3Deployment and dereference where it's actually used?  That would require re-checking non-nil, though (to be safe, anyway).  Anyway, just a suggestion.

sadasu: Yes, I am working on some improvements in baremetal_pod.go that includes what you just suggested.

pkg/operator/baremetal_test.go
sadasu: Yes. Fixing this in the follow-up PR.

PR #565, opened 2020-04-20
FAQ: Expand on in-place updates

cgwalters:
Link to the MCO toplevel as well as the specific upgrades doc
since they allow the user to find out more.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -142,7 +142,7 @@ Yes, though consider carefully before doing so.  Inspect the Match Labels from t
 # Machine Deployments
 
 ## Upstream kubernetes cluster-api project is utilizing MachineDeployments, why isn’t OpenShift?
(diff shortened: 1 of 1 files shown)


cgwalters:
Aside, VS Code's markdown mode insisted on doing this:

```
diff --git a/FAQ.md b/FAQ.md
index 06d37016..603d3878 100644
--- a/FAQ.md
+++ b/FAQ.md
@@ -3,6 +3,7 @@
 This document consists of frequently asked questions intended for users.
 
 <!-- toc -->
+- [User FAQ](#user-faq)
 - [Important Terminology](#important-terminology)
   - [Machine vs Node vs VM/Instance](#machine-vs-node-vs-vminstance)
 - [Machines](#machines)
```

I ended up using `git commit --interactive` in order to only get my change.

cgwalters:
(Also forgot to note in the commit message, I removed the "VM" term since the MCO/ostree model also applies to bare metal, and it's important to remember that)

PR #573, opened 2020-04-28
Bug 1829333: vSphere - Don't reconcile power state

alexander-demichev:
We should support the ability to switch off VM but current implementation turns on VM in every machine update. 
I removed the reconciliation of power state and set `PowerOn:  true` on machine creation. We have this part copied from upstream and powerOn set to true shouldn't be an issue for us.


enxebre:
/approve
Can you please tag the relevant BZ
/hold

PR #575, opened 2020-04-29
Bug 1808971: Machine status shows "running" when an instance was terminated

Danil-Grigorev:
Setting machine instance state to unknown on machine failure

### DIFF
--- a/pkg/controller/machine/controller.go
+++ b/pkg/controller/machine/controller.go
@@ -87,6 +87,9 @@ const (
 	// Machine has a deletion timestamp
 	phaseDeleting = "Deleting"
 
+	// Hardcoded instance state set on machine failure
+	unknownInstanceState = "Unknown"
+
 	skipWaitForDeleteTimeoutSeconds = 60 * 5
 )
 
@@ -426,9 +429,17 @@ func (r *ReconcileMachine) setPhase(machine *machinev1.Machine, phase string, er
 		machine.Status.LastUpdated = &now
 		if phase == phaseFailed && errorMessage != "" {
 			machine.Status.ErrorMessage = &errorMessage
+			if machine.Annotations == nil {
+				machine.Annotations = map[string]string{}
+			}
+			machine.Annotations[MachineInstanceStateAnnotationName] = unknownInstanceState
+			if err := r.Client.Patch(context.Background(), machine, baseToPatch); err != nil {
+				klog.Errorf("Failed to update machine %q: %v", machine.GetName(), err)
+				return err
+			}
 		}
 		if err := r.Client.Status().Patch(context.Background(), machine, baseToPatch); err != nil {
-			klog.Errorf("Failed to update machine %q: %v", machine.GetName(), err)
+			klog.Errorf("Failed to update machine status %q: %v", machine.GetName(), err)
 			return err
 		}
 	}
--- a/pkg/controller/machine/controller_test.go
+++ b/pkg/controller/machine/controller_test.go
@@ -356,7 +356,7 @@ func TestSetPhase(t *testing.T) {
 		t.Fatal("Got phase nil")
 	}
 	if *got.Status.Phase != phaseRunning {
-		t.Errorf("Got: %v, expected: %v", *got.Status.Phase, phaseRunning)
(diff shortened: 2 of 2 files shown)


alexander-demichev:
This can break the consistency because we set same annotation in 2 different places. However, the idea of setting this in one place is nice.

pkg/controller/machine/controller.go
```diff
@@ -426,9 +426,18 @@ func (r *ReconcileMachine) setPhase(machine *machinev1.Machine, phase string, er
 		machine.Status.LastUpdated = &now
 		if phase == phaseFailed && errorMessage != "" {
 			machine.Status.ErrorMessage = &errorMessage
+			if machine.Annotations == nil {
+				machine.Annotations = map[string]string{}
```
enxebre, line 430: you could save the if:
```
annotations := m.GetAnnotations()
annotations["key"] = "value"
m.SetAnnotations(annotations)
```

pkg/controller/machine/controller.go
```diff
@@ -426,9 +426,18 @@ func (r *ReconcileMachine) setPhase(machine *machinev1.Machine, phase string, er
...
+			}
+			machine.Annotations[MachineInstanceStateAnnotationName] = "Unknown"
```
enxebre, line 432: can we put this into a constant?

enxebre, line 432: I think we should only set unknown here https://github.com/openshift/machine-api-operator/pull/575/files#diff-b96ce5bf8686fb02d855188000ae1cb4R299-R304

pkg/controller/machine/controller.go
Danil-Grigorev: This will cause a panic on `Annotations` being nil, especially in tests. I'm not sure how to avoid nil check here.

pkg/controller/machine/controller.go
Danil-Grigorev: Here is another place, where the machine will get into Failed pahse, and the VM would not be monitored/updated anymore: https://github.com/openshift/machine-api-operator/blob/46410db09aae40009c72d3070ff7dc0bce6c4478/pkg/controller/machine/controller.go#L313-L316 I was thinking to handle it generically.

pkg/controller/machine/controller.go
enxebre: I'm missing where exactly would this panic?

pkg/controller/machine/controller.go
enxebre: make sense, in the other scenario the provider actually have a chance to set a valid state as it goes failed so might good to leave it. Happy to hear others thoughts.

pkg/controller/machine/controller_test.go
```diff
@@ -401,12 +403,21 @@ func TestSetPhase(t *testing.T) {
...
 	}
 	// validate passed object
 	if expecterErrorMessage != *machine.Status.ErrorMessage {
 		t.Errorf("Expected: %v, got: %v", expecterErrorMessage, *machine.Status.ErrorMessage)
 	}
+	if got.Status.Phase == nil {
+		t.Fatal("Got phase nil")
+	}
+	if *got.Status.Phase != phaseFailed {
+		t.Errorf("Got: %v, expected: %v", *got.Status.Phase, phaseRunning)
```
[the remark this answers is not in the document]
Danil-Grigorev: Oh I missed the fact that all those fields are pointers :D Thanks for the tip, will revert it.

pkg/controller/machine/controller.go
```diff
@@ -426,9 +429,17 @@ func (r *ReconcileMachine) setPhase(machine *machinev1.Machine, phase string, er
 		machine.Status.LastUpdated = &now
 		if phase == phaseFailed && errorMessage != "" {
 			machine.Status.ErrorMessage = &errorMessage
+			if machine.Annotations == nil {
+				machine.Annotations = map[string]string{}
+			}
+			machine.Annotations[MachineInstanceStateAnnotationName] = unknownInstanceState
```
enxebre, line 435: this might cause skew if the actuators set something in the provider.Status. We might consolidate to consolidate that.

enxebre:
/approve
@JoelSpeed PTAL

PR #586, opened 2020-05-12
Bug 1796412: cluster-reader is unable to view machine resources

Danil-Grigorev:
Added cluster-reader rbac for machine CRDs

### DIFF
--- a/config/cluster-reader-rbac.yaml
+++ b/config/cluster-reader-rbac.yaml
@@ -1,0 +1,18 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: machine-api-operator:cluster-reader
+  labels:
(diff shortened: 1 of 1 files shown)


config/cluster-reader-rbac.yaml
```diff
@@ -0,0 +1,20 @@
...
+metadata:
+  name: machine-api-operator:cluster-reader
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
+rules:
+ apiGroups:
+  - machine.openshift.io
+  resources:
+  - machineconfigpools
+  - machineconfigs
```
enxebre, line 12: please remove machineconfigs and machineconfigpools, those should be handle by the machine config operator https://github.com/openshift/machine-config-operator

PR #587, opened 2020-05-13
Add dgrigore to approvers

Danil-Grigorev:


enxebre:
Hey @Danil-Grigorev We need to either create a BZ or hold it until master opens for 4.6
/hold
to stop test spamming

PR #589, opened 2020-05-14
Bug 1827050: Update to use revised ApplyDeployment

marun:
The new method hashes the spec of the deployment which removes the need for forcing rollout.


enxebre:
Thanks @marun!
I thought we have a test for this. We should add some tests before merging.

Should probably break it down  `syncClusterAPIController` and have something only responsible for what we want to test the likes of

```
func (optr *Operator) ensureDeployment(deployment *appsv1.Deployment) error {
	expectedGeneration := resourcemerge.ExpectedDeploymentGeneration(deployment, optr.generations)

	d, updated, err := resourceapply.ApplyDeployment(optr.kubeClient.AppsV1(),
		events.NewLoggingEventRecorder(optr.name), deployment, expectedGeneration, false)
	if err != nil {
		return err
	}

	if updated {
		resourcemerge.SetDeploymentGeneration(&optr.generations, d)
	}
}
```
Then we can validate with the steps:
ensureDeployment.
updateDeployment out of band.
ensureDeployment. Check it matches original spec.

To avoid kube fakeClient "misbehaviour" we should also probably also test against an envtest environment, just similar to what do here https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/machineset/machineset_controller_test.go

I also created https://issues.redhat.com/browse/OCPCLOUD-867 to make sure we cover with an e2e test.

michaelgugino:
@enxebre the operator is enforcing the expected deployment today.  This change only impacts if the operator somehow wants to deploy something different (outside of an upgrade).

michaelgugino:
/hold cancel

Looks like origin is on kube 1.18.2, and this bumps us to that as well.  Changes in library-go were merged 20+ days ago, so this is probably safe.

enxebre:
>@enxebre the operator is enforcing the expected deployment today. This change only impacts if the operator somehow wants to deploy something different (outside of an upgrade).

Yes we have a test to validate the operator is enforcing the expected deployment today https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/operators/machine-api-operator.go#L25

So we need another one to validate if the deployment gets modified out of band it's reconciled by the operator.
So if anything change the details of `syncClusterAPIController` as in this PR. We know does not break the two expectations above. https://github.com/openshift/machine-api-operator/pull/589#issuecomment-628783262

PR #597, opened 2020-05-25
Bug 1840665: [vSphere] Get insecure flag from provider config

alexander-demichev:
We should get the insecure flag from provider config and default to secure connection if flag is missing

https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/existing.html#single-vcenter

### DIFF
--- a/pkg/controller/vsphere/machine_scope_test.go
+++ b/pkg/controller/vsphere/machine_scope_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"testing"
 
+	configv1 "github.com/openshift/api/config/v1"
 	machinev1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	vspherev1 "github.com/openshift/machine-api-operator/pkg/apis/vsphereprovider/v1beta1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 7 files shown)


pkg/controller/vsphere/machine_scope_test.go
```diff
@@ -279,8 +281,13 @@ func TestPatchMachine(t *testing.T) {
 	model, _, server := initSimulator(t)
 	defer model.Remove()
 	defer server.Close()
-	credentialsSecretUsername := fmt.Sprintf("%s.username", server.URL.Host)
-	credentialsSecretPassword := fmt.Sprintf("%s.password", server.URL.Host)
+	host, port, err := net.SplitHostPort(server.URL.Host)
```
enxebre, line 284: how is this change related to this PR?

alexander-demichev: We need to pass the test provider config where insecure connection is allowed to all tests because simulator doesn't support secure connections.

pkg/controller/vsphere/util_test.go
```diff
@@ -67,4 +69,8 @@ func TestGetVSphereConfig(t *testing.T) {
 	if vSphereConfig.Global.Port != testPort {
 		t.Errorf("Expected zone %s, got %s", testZone, vSphereConfig.Global.Port)
 	}
+
+	if vSphereConfig.Global.InsecureFlag != testInsecureFlag {
+		t.Errorf("Expected zone %s, got %s", testZone, vSphereConfig.Global.Port)
```
elmiko, line 74: this error message is the same as the previous message, also it is reporting with the port. is this accurate or should it emit something about the insecure flag?

alexander-demichev: oh, right :D ctrl+C, ctrl+V

PR #605, opened 2020-06-01
Bug 1843327: [vSphere] Improve provider spec validation

alexander-demichev:
Return invalid machine configuration error on missing VM template and missing user data secret

### DIFF
--- a/pkg/controller/vsphere/machine_scope.go
+++ b/pkg/controller/vsphere/machine_scope.go
@@ -8,6 +8,7 @@ import (
 	machinev1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	apivsphere "github.com/openshift/machine-api-operator/pkg/apis/vsphereprovider/v1beta1"
 	machineapierros "github.com/openshift/machine-api-operator/pkg/controller/machine"
+	machinecontroller "github.com/openshift/machine-api-operator/pkg/controller/machine"
 	"github.com/openshift/machine-api-operator/pkg/controller/vsphere/session"
 	apicorev1 "k8s.io/api/core/v1"
 	apimachineryerrors "k8s.io/apimachinery/pkg/api/errors"
@@ -133,7 +134,7 @@ func (s *machineScope) GetSession() *session.Session {
 // provider spec, if one is set.
 func (s *machineScope) GetUserData() ([]byte, error) {
 	if s.providerSpec == nil || s.providerSpec.UserDataSecret == nil {
-		return nil, nil
+		return nil, machinecontroller.InvalidMachineConfiguration("user data secret is missing in provider spec")
 	}
 
 	userDataSecret := &apicorev1.Secret{}
--- a/pkg/controller/vsphere/reconciler.go
+++ b/pkg/controller/vsphere/reconciler.go
@@ -454,7 +454,9 @@ func clone(s *machineScope) (string, error) {
 
 	vmTemplate, err := s.GetSession().FindVM(*s, "", s.providerSpec.Template)
 	if err != nil {
-		return "", err
+		const notFoundMsg = "template not found, specify valid value"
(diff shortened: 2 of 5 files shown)


pkg/controller/vsphere/reconciler.go
```diff
@@ -410,7 +410,9 @@ func clone(s *machineScope) (string, error) {
 
 	vmTemplate, err := s.GetSession().FindVM(*s, "", s.providerSpec.Template)
 	if err != nil {
-		return "", err
+		notFoundMsg := "template not found, specify valid value"
```
elmiko, line 413: given that `notFoundMsg` doesn't change, i wonder if it's better to make it a const?

```suggestion
		const notFoundMsg = "template not found, specify valid value"
```

pkg/controller/vsphere/reconciler.go
```diff
@@ -589,7 +591,10 @@ func getNetworkDevices(s *machineScope, devices object.VirtualDeviceList) ([]typ
 
 		ref, err := s.GetSession().Finder.Network(s.Context, netSpec.NetworkName)
 		if err != nil {
-			return nil, fmt.Errorf("unable to find network %q: %w", netSpec.NetworkName, err)
+			multipleFoundMsg := "multiple networks found, specify one in config"
+			notFoundMsg := "network not found, specify valid value"
```
elmiko, line 595: similar question here for `multipleFoundMsg` and `notFoundMsg`

alexander-demichev:
@JoelSpeed @elmiko all fixed

/retest

pkg/controller/vsphere/reconciler.go
```diff
@@ -456,15 +458,15 @@ func clone(s *machineScope) (string, error) {
...
+		const multipleFoundMsg = "multiple folders found, specify one in config"
 		notFoundMsg := "folder not found, specify valid value"
 		defaultError := fmt.Errorf("unable to get folder for %q: %w", folderPath, err)
 		return "", handleVSphereError(multipleFoundMsg, notFoundMsg, defaultError, err)
 	}
 
 	datastore, err := s.GetSession().Finder.DatastoreOrDefault(s, datastorePath)
 	if err != nil {
-		multipleFoundMsg := "multiple datastores found, specify one in config"
+		const multipleFoundMsg = "multiple datastores found, specify one in config"
```
enxebre, line 469: why is this using const and := unconsistently??
https://github.com/openshift/machine-api-operator/pull/605/files#diff-11b6610f82e5d47fdfb8dda44b90dd1eR469
https://github.com/openshift/machine-api-operator/pull/605/files#diff-11b6610f82e5d47fdfb8dda44b90dd1eR470
https://github.com/openshift/machine-api-operator/pull/605/files#diff-11b6610f82e5d47fdfb8dda44b90dd1eR594
https://github.com/openshift/machine-api-operator/pull/605/files#diff-11b6610f82e5d47fdfb8dda44b90dd1eR595

pkg/controller/vsphere/machine_scope.go
```diff
@@ -133,7 +134,7 @@ func (s *machineScope) GetSession() *session.Session {
 // provider spec, if one is set.
 func (s *machineScope) GetUserData() ([]byte, error) {
 	if s.providerSpec == nil || s.providerSpec.UserDataSecret == nil {
-		return nil, nil
+		return nil, machinecontroller.InvalidMachineConfiguration("user data secret is missing in provider spec")
```
enxebre, line 137: Although I think this is fine and it makes sense as we don't currently support any model where this could be empty fwiw I don't think we are doing the same in other providers.
In the near future with a model where MCO generate secrets for machines on demand this might need to change.
https://github.com/openshift/enhancements/pull/368
https://github.com/openshift/enhancements/pull/201

enxebre:
"Improve" seems too wide, Can we include "by returning invalid machine configuration error on missing VM template and missing user data secret" as in the desc?
/approve

alexander-demichev:
@enxebre done

PR #618, opened 2020-06-18
Bug 1845610: Windows VM MachineSet error on Azure

Danil-Grigorev:
Fix for offer selection

Add missing option to specify offer/image/sku for azure.

### DIFF
--- a/pkg/apis/machine/v1beta1/machine_webhook.go
+++ b/pkg/apis/machine/v1beta1/machine_webhook.go
@@ -433,7 +433,7 @@ func defaultAzure(h *defaulterHandler, m *Machine) (bool, utilerrors.Aggregate) 
 		}
 	}
 
-	if providerSpec.Image.ResourceID == "" {
+	if providerSpec.Image == (azure.Image{}) {
 		providerSpec.Image.ResourceID = defaultAzureImageResourceID(h.clusterID)
 	}
 
@@ -516,16 +516,14 @@ func validateAzure(h *validatorHandler, m *Machine) (bool, utilerrors.Aggregate)
 		errs = append(errs, field.Required(field.NewPath("providerSpec", "networkResourceGroup"), "must provide a network resource group when a virtual network or subnet is specified"))
 	}
 
-	if providerSpec.Image.ResourceID == "" {
-		errs = append(errs, field.Required(field.NewPath("providerSpec", "image", "resourceID"), "resourceID must be provided"))
-	}
+	errs = append(errs, validateAzureImage(providerSpec.Image)...)
 
 	if providerSpec.ManagedIdentity == "" {
 		errs = append(errs, field.Required(field.NewPath("providerSpec", "managedIdentity"), "managedIdentity must be provided"))
 	}
 
 	if providerSpec.ResourceGroup == "" {
-		errs = append(errs, field.Required(field.NewPath("providerSpec", "resourceGropu"), "resourceGroup must be provided"))
+		errs = append(errs, field.Required(field.NewPath("providerSpec", "resourceGroup"), "resourceGroup must be provided"))
 	}
 
 	if providerSpec.UserDataSecret == nil {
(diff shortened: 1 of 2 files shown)


pkg/apis/machine/v1beta1/machine_webhook.go
```diff
@@ -517,15 +517,34 @@ func validateAzure(h *validatorHandler, m *Machine) (bool, utilerrors.Aggregate)
 	}
 
 	if providerSpec.Image.ResourceID == "" {
-		errs = append(errs, field.Required(field.NewPath("providerSpec", "image", "resourceID"), "resourceID must be provided"))
+		imageProvided := true
```
enxebre, line 520: I'm not following this, why `imageProvided := true` when `providerSpec.Image.ResourceID == ""`?

enxebre: Now this won't error when providerSpec.Image.ResourceID == ""? Why?

Danil-Grigorev: Let me rewrite this, it actually does not seem clear. But this is due to logic in 

https://github.com/openshift/cluster-api-provider-azure/blob/master/pkg/cloud/azure/services/virtualmachines/virtualmachines.go#L105-L115

Danil-Grigorev: I'm going to adjust the support in https://github.com/openshift/cluster-api-provider-azure/pull/140, so the options will be equivalent, and you could either specify `imageReference` by offer etc. or give it a recourceID.

enxebre: Could we may be make this logic less clever and more explicit so is easier to read?
And present the output as one atomic error similar to what we do here -> https://github.com/openshift/machine-api-operator/blob/master/pkg/apis/machine/v1beta1/machine_webhook.go#L331-L339

pkg/apis/machine/v1beta1/machine_webhook.go
```diff
@@ -516,16 +516,35 @@ func validateAzure(h *validatorHandler, m *Machine) (bool, utilerrors.Aggregate)
 		errs = append(errs, field.Required(field.NewPath("providerSpec", "networkResourceGroup"), "must provide a network resource group when a virtual network or subnet is specified"))
 	}
 
+	if providerSpec.Image.ResourceID == "" &&
```
JoelSpeed, line 519: How about we break this into a seperate function so that we can return early, similar to what was done for the GCP provider? 

It would be good to make the resourceID mutually exclusive from the rest of the fields, ie, if resource ID is set and any other field is, this should be an error, I don't think that happens currently

JoelSpeed:
@Danil-Grigorev We need to make sure we only default the resource ID if the other fields are not set, currently we don't
https://github.com/openshift/machine-api-operator/blob/ba983a5a8c794d819e3cd7f19afc055a52de7d9e/pkg/apis/machine/v1beta1/machine_webhook.go#L436-L438

Can we include this in this PR as it seems very related
```
if providerSpec.Image == azure.Image{} {
 	// Image is empty, so default to installer created image
 	providerSpec.Image.ResourceID = defaultAzureImageResourceID(h.clusterID) 
 } 
```

JoelSpeed:
/approve

Looks great, thanks @Danil-Grigorev

pkg/apis/machine/v1beta1/machine_webhook.go
```diff
@@ -562,6 +562,41 @@ func validateAzure(h *validatorHandler, m *Machine) (bool, utilerrors.Aggregate)
...
+		if !(image.Offer == "" &&
+			image.Publisher == "" &&
+			image.SKU == "" &&
+			image.Version == "") {
+			return field.Required(field.NewPath("providerSpec", "image", "resourceID"), "resourceID is already specified, other fields such as [Offer, Publisher, SKU, Version] should not be set")
+		}
+		return nil
+	}
+
+	if image.Offer == "" {
```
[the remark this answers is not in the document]
enxebre: wouldn't this also be exactly what you'd expect to see as a user? Similar to what we do https://github.com/openshift/machine-api-operator/blob/master/pkg/apis/machine/v1beta1/machine_webhook.go#L342-L350
As far as we are consistent everywhere I'm good.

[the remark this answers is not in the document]
JoelSpeed: I think this is kind of covered by the above cases, there's three paths here
- No resourceID + No offer/publisher - covered by L#565-L#567, user gets `an image reference must be provided`
- Resource ID provided + other fields - covered by L#570-L#572, user gets `resourceID is already specified, other fields such as [Offer, Publisher, SKU, Version] should not be set`
- No Resource ID + other fields - covered here down, if any of them are missing a message is returned

We could make this more obvious perhaps by changing the error on L#566 to read `either a resource ID must be provided or the Offer, Publisher, SKU and Version must be provided`

Or we drop these checks individually below and just check for all four being set and return a single error, WDYT?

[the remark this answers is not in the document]
enxebre: >either a resource ID must be provided or the Offer, Publisher, SKU and Version must be provided
Or we drop these checks individually below and just check for all four being set and return a single error, WDYT?

I'd be in favour of that.

PR #620, opened 2020-06-21
WIP: sig controller-runtime grace period testing

michaelgugino:
Do not merge

/hold

### DIFF
--- a/pkg/apis/machine/v1beta1/machineset_webhook_test.go
+++ b/pkg/apis/machine/v1beta1/machineset_webhook_test.go
@@ -1,6 +1,7 @@
 package v1beta1
 
 import (
+	"context"
 	"fmt"
 	"testing"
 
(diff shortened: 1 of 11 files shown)


pkg/apis/machine/v1beta1/machineset_webhook_test.go
```diff
@@ -0,0 +1,941 @@
...
+			expectedError: "Requested DELETE of Control Plane MachineSet Not Allowed.",
+			objectMeta: ObjectMeta{
+				Labels: map[string]string{
+					"machine.openshift.io/cluster-api-machine-role": "master",
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
```
michaelgugino, line 677: This is the test that fails most often without this patch.  With this patch, the tests run indefinitely until they hit the 10 minute testing timeout panic built into golang.

PR #625, opened 2020-06-23
OWNERS: Add a component link for BZ

cgwalters:
There is apparently tooling that uses this, came up in chat.
See e.g. the MCO:
https://github.com/openshift/machine-config-operator/blob/master/OWNERS#L12


enxebre:
thanks!
/approve
/lgtm

PR #628, opened 2020-06-30
Enable termination handler for GCP

alexander-demichev:
Enable termination handler for GCP


enxebre:
We are defining a daemonsSet with a binary command which does not exist yet in that image get and operatorStatus is transparently reporting degraded=false https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_machine-api-operator/628/pull-ci-openshift-machine-api-operator-master-e2e-gcp/1277953390990594048/artifacts/e2e-gcp/gather-extra/clusteroperators.json

In a follow up we need to include a fix to go degraded when the daemonSet is unable to run pods. I captured this in https://issues.redhat.com/browse/OCPCLOUD-903.

PR #630, opened 2020-07-02
Add MachineHealthCheck and BareMetalHost to related objects

n1r1:
[must-gather ](https://docs.openshift.com/container-platform/4.4/support/gathering-cluster-data.html#about-must-gather_gathering-cluster-data)is a tool which captures data from a cluster for debugging purposes.
One way for it to collect data is to look for related objects.
Since MAO installs baremetalhost CRD, we want to have it in the related objects.
This PR also adds MachineHealthCheck object so we can have MachineHealthCheck CRs in must-gather.


enxebre:
thanks!
/approve

PR #639, opened 2020-07-14
Show error message when fail to get clusterId from infrastructure

Danil-Grigorev:
Adds missing infrastructure log error message, causing non-representative machine reconcile error loops
when there is a re-occuring error with  `client.Get` on the infrastructure resource.

### DIFF
--- a/pkg/controller/machine/controller.go
+++ b/pkg/controller/machine/controller.go
@@ -180,6 +180,7 @@ func (r *ReconcileMachine) Reconcile(request reconcile.Request) (reconcile.Resul
 
 	// Add clusterID label
 	if err := r.setClusterIDLabel(ctx, m); err != nil {
+		klog.Errorf("%v: failed to set ClusterID for machine: %v", machineName, err)
 		return reconcile.Result{}, err
 	}
 


pkg/controller/machine/controller.go
```diff
@@ -180,6 +180,7 @@ func (r *ReconcileMachine) Reconcile(request reconcile.Request) (reconcile.Resul
 
 	// Add clusterID label
 	if err := r.setClusterIDLabel(ctx, m); err != nil {
+		klog.Errorf("%v: failed to get infrastructure for machine: %v", machineName, err)
```
enxebre, line 183: This error message does not corresponds with the caller that is returning it.
We should either rephrase or move it inside the function where trying to get the infra resource.
Also wouldn't this error be logged by controller runtime https://github.com/kubernetes-sigs/controller-runtime/blob/bb96678bedc9d98361be219cb21562eb99ab9e15/pkg/internal/controller/controller.go#L235?

Danil-Grigorev: It didn't show any error in CI runs: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-api-provider-openstack/106/pull-ci-openshift-cluster-api-provider-openstack-master-e2e-openstack/1282763600154333184/artifacts/e2e-openstack/pods/openshift-machine-api_machine-api-controllers-66d8c68c7c-zlwx4_machine-controller.log

enxebre: how di you conclude that's failing then?

Danil-Grigorev: `configv1` was not imported into the scheme. I just spent some time, executing the openstack provider manually :roll_eyes: and mentioned the conclusion here: https://github.com/openshift/cluster-api-provider-openstack/pull/106#issuecomment-657973741 As you can see, it's the first successful CI pass for that PR.

enxebre:
closing in favour of https://github.com/openshift/machine-api-operator/pull/644

@Danil-Grigorev Is there an answer for https://github.com/openshift/machine-api-operator/pull/639#pullrequestreview-447974915 so we can capture an action item for this?

Danil-Grigorev:
> closing in favour of https://github.com/openshift/machine-api-operator/pull/644
> 
> @Danil-Grigorev Is there an answer for https://github.com/openshift/machine-api-operator/pull/639#pullrequestreview-447974915 so we can capture an action item for this?

I'll look for conformation later.

PR #641, opened 2020-07-15
update the baremetalhost crd for bootmode and rootdevicehints

dhellmann:
Update the CRD for BareMetalHost to the version that includes bootMode
and rootDeviceHints, 2 features added for the 4.6 release.


zaneb:
I guess we should make it part of our process to update this whenever we sync the baremetal-operator from upstream.

dhellmann:
> I guess we should make it part of our process to update this whenever we sync the baremetal-operator from upstream.

Ideally we could pull the CRD from the baremetal-operator repo as part of a build step instead, but yeah.

dhellmann:
This has a LGTM but needs an approval.

PR #642, opened 2020-07-15
Refactor webhook syncronization to work with library-go implementation

Danil-Grigorev:
This PR removes our webhook management code in favor of https://github.com/openshift/library-go/pull/836 

It also removes unit tests, which are currently impossible to mock as it depends on generation increments, which is not supported even by `testenv` standards. But we have all the necessary infrastructure in 
https://github.com/openshift/cluster-api-actuator-pkg/blob/master/pkg/operators/machine-api-operator.go#L75-L193 which extensively covers the code and typical use cases, and it is covered in https://github.com/openshift/library-go/pull/836 which this PR is pulling.


elmiko (approved):
/approve

PR #645, opened 2020-07-16
[WIP] Pass through of config variables to ironic

derekhiggins:
Mark as a WIP as further discussion is needed, to be used with https://github.com/openshift/installer/pull/3887


derekhiggins:
OK will close this, looks like we'd rather add things when needed.

PR #646, opened 2020-07-16
add trusted ca bundle to machine-controllers

abhinavdahiya:
### Why

The installer team is investigating running CI for AWS GovCloud https://issues.redhat.com/browse/CORS-1315 . Because of various restrictions the team and DPP are working to enable Gov Cloud emulator from SHIFT https://issues.redhat.com/browse/DPP-4708 as our solution to testing these restricted environment on AWS.

The SHIFT emulator works almost like Man-In-Middle-Terminating SSL Proxy capturing and transporting all connections from the chosen VPC to their own backbone and then emulating the AWS calls to say GovCloud region onto the source VPC's region which can be any commercial region.

So the allow testing of ipi workflow in the above mentioned environment we need the machine-controller talking to aws apis like other operators and components pulling content from internet to trust the certificate of the mimt-ssl-proxy. 

### add trusted ca bundle to machine-controllers 

The machine-controllers communicate with AWS APIs that are outside the openshift cluster, therefore
it is possible that if the user has setup a Man-In-Middle-Terminating SSL Proxy for external traffic for reasons like auditing,
the machine-controllers will fail to communicate with AWS APIs as they do not trust the mimt-ssl-proxy's certificate.

The goal is to support just the case of mimt-proxy because this usually is applied using traffic shaping that directs all external traffic to it,
and therefore it makes sense to support it and still keep the previous stance that openshift expects to reach out to cloud APIs directly.

The implementation is very similar to the other cluster operators like cluster-ingress-operator in https://github.com/openshift/cluster-ingress-operator/pull/334/files

It creates a configmap in operator namespace with the annotation `config.openshift.io/inject-trusted-cabundle: "true"` which will include the complete trust for the container
and will be updated when user changes their configuration to add or change the user-defined trust for mimt-proxy.

### So the trust bundle is loaded into the machine-controller's deployment as a config map volume, so the contents of the file
will automatically be syned when the config map contents change due to external factors.

But all the http clients that have already been created by the machine-controller like the cloud SDK clients, these will not see the
new trust unless re-created. So therefore to reload the trust for these clients we have 2 options,

- update all the cluster-api-provider implementations to include an ability to reload/restart when a file changes on disk.
  This requires a lot of work in each of the implementations and also creates a lot of duplicated code when we cannot use the same
  code for the behavior because we use upstream-downstream patches.

- track external dependencies and update the deployment object to rollout the changes as part of management in the machine-api-operator itself.
  This keeps the code to track this and others if we do end up extending it in one place in the operator. This method is extensively used
  by various other library-go based operators like openshift-apiserver-oeprator here https://github.com/openshift/cluster-openshift-apiserver-operator/blob/master/pkg/operator/workload/workload_openshiftapiserver_v311_00_sync.go#L340-L351
  and many others for similar behavior.
Here I have decided to use the latter option.

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -392,6 +392,18 @@ func newPodTemplateSpec(config *OperatorConfig, features map[string]bool) *corev
 				},
 			},
 		},
+		{
+			Name: "trusted-ca",
+			VolumeSource: corev1.VolumeSource{
+				ConfigMap: &corev1.ConfigMapVolumeSource{
+					Items: []corev1.KeyToPath{{Key: "ca-bundle.crt", Path: "tls-ca-bundle.pem"}},
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: "machine-api-operator-trusted-ca",
+					},
+					Optional: pointer.BoolPtr(true),
+				},
+			},
+		},
 	}
 
 	return &corev1.PodTemplateSpec{
@@ -501,6 +513,13 @@ func newContainers(config *OperatorConfig, features map[string]bool) []corev1.Co
 						Path: "/readyz",
 						Port: intstr.Parse("healthz"),
 					},
+				},
+			},
+			VolumeMounts: []corev1.VolumeMount{
+				{
+					MountPath: "/etc/pki/ca-trust/extracted/pem",
+					Name:      "trusted-ca",
+					ReadOnly:  true,
 				},
 			},
 		},
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -7,6 +7,7 @@ import (
 	"github.com/golang/glog"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcehash"
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 	mapiv1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	machinecontroller "github.com/openshift/machine-api-operator/pkg/controller/machine"
@@ -37,6 +38,7 @@ const (
 	defaultMachineHealthCheckHealthPort = 9442
 	kubeRBACConfigName                  = "config"
 	certStoreName                       = "machine-api-controllers-tls"
+	externalTrustBundleConfigMapName    = "mao-trusted-ca"
 )
 
 func (optr *Operator) syncAll(config *OperatorConfig) error {
@@ -101,6 +103,17 @@ func (optr *Operator) syncAll(config *OperatorConfig) error {
 
 func (optr *Operator) syncClusterAPIController(config *OperatorConfig) error {
 	controllersDeployment := newDeployment(config, nil)
+
+	// we watch some resources so that our deployment will redeploy without explicitly and carefully ordered resource creation
+	inputHashes, err := resourcehash.MultipleObjectHashStringMapForObjectReferences(
+		optr.kubeClient,
+		resourcehash.NewObjectRef().ForConfigMap().InNamespace(config.TargetNamespace).Named(externalTrustBundleConfigMapName),
+	)
+	if err != nil {
+		return fmt.Errorf("invalid dependency reference: %q", err)
+	}
+	ensureDependecyAnnotations(inputHashes, controllersDeployment)
+
 	expectedGeneration := resourcemerge.ExpectedDeploymentGeneration(controllersDeployment, optr.generations)
 	d, updated, err := resourceapply.ApplyDeployment(optr.kubeClient.AppsV1(),
 		events.NewLoggingEventRecorder(optr.name), controllersDeployment, expectedGeneration)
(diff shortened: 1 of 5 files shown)


pkg/operator/sync.go
```diff
@@ -392,6 +392,18 @@ func newPodTemplateSpec(config *OperatorConfig, features map[string]bool) *corev
 				},
 			},
 		},
+		{
+			Name: "trusted-ca",
+			VolumeSource: corev1.VolumeSource{
+				ConfigMap: &corev1.ConfigMapVolumeSource{
+					Items: []corev1.KeyToPath{{Key: "ca-bundle.crt", Path: "tls-ca-bundle.pem"}},
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: "machine-api-operator-trusted-ca",
```
enxebre: `machine-api-operator-trusted-ca` is also used here https://github.com/openshift/machine-api-operator/pull/646/commits/145fae715a77084768d974a1a9c8124aa62b950a#diff-fa45321336db7ad1cedc28bf643a4f97R109. Can we put it in a constant?

pkg/operator/sync.go
```diff
@@ -101,6 +102,24 @@ func (optr *Operator) syncAll(config *OperatorConfig) error {
...
+	inputHashes, err := resourcehash.MultipleObjectHashStringMapForObjectReferences(
+		optr.kubeClient,
+		resourcehash.NewObjectRef().ForConfigMap().InNamespace(config.TargetNamespace).Named("machine-api-operator-trusted-ca"),
+	)
+	if err != nil {
+		return fmt.Errorf("invalid dependency reference: %q", err)
+	}
+	for k, v := range inputHashes {
+		annotationKey := fmt.Sprintf("operator.openshift.io/dep-%s", k)
+		controllersDeployment.Annotations[annotationKey] = v
```
enxebre, line 116: shouldn't this also safe initialise `controllersDeployment.Annotations`?

pkg/operator/sync.go
```diff
@@ -101,6 +102,24 @@ func (optr *Operator) syncAll(config *OperatorConfig) error {
 
 func (optr *Operator) syncClusterAPIController(config *OperatorConfig) error {
 	controllersDeployment := newDeployment(config, nil)
+
+	// we watch some resources so that our deployment will redeploy without explicitly and carefully ordered resource creation
+	inputHashes, err := resourcehash.MultipleObjectHashStringMapForObjectReferences(
```
enxebre, line 107: so is this logic to rollout the deployment replicas if the configMap changes?
If so why is that necessary?
shouldn't this be updated automatically https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically

enxebre: can we put this logic in a unit tested func that we just invoke here?

enxebre:
thanks @abhinavdahiya! dropped some questions. Can we also please include a PR desc possibly referencing the story/enhancement this is supporting?

pkg/operator/sync.go
abhinavdahiya: So i created the `forceSyncDeployment` function that takes in inputHashes and then applies that to the deployment, and added a unit test for that.

generation of the inputhashes is function from library-go, so personally it didn't make sense to test it's behavior again as part of unit testing here.

abhinavdahiya:
> thanks @abhinavdahiya! dropped some questions. Can we also please include a PR desc possibly referencing the story/enhancement this is supporting?

Thanks @enxebre for looking at the PR, i was still testing this so didn't put some text. I have updated the commits and PR description to include the details on the reasoning..

pkg/operator/sync.go
```diff
@@ -674,3 +706,19 @@ func newTerminationContainers(config *OperatorConfig) []corev1.Container {
 		},
 	}
 }
+
+// forceSyncDeployment uses inputHash map of external dependencies to force new generation of the deployment
+// triggering the Kubernetes rollout as defined when the inputHash changes.
+func forceSyncDeployment(inputHashes map[string]string, deployment *appsv1.Deployment) {
+	for k, v := range inputHashes {
+		annotationKey := fmt.Sprintf("dep-%s", k)
```
michaelgugino, line 714: We should prefix any annotations with an openshift domain prefix.

"x.openshift.io/dep-%s"

Annotation keys without prefix are supposed to be reserved for the user.

enxebre:
>The machine-controllers communicate with AWS APIs that are outside the openshift cluster, therefore
it is possible that if the user has setup a Man-In-Middle-Terminating SSL Proxy for external traffic for reasons like auditing,
the machine-controllers will fail to communicate with AWS APIs as they do not trust the mimt-ssl-proxy's certificate.

Wouldn't we need to make this containers honour the http(s)_proxy variables for this to work at all?

pkg/operator/sync.go
enxebre: ok, thanks!
personally I see `// forceSyncDeployment uses inputHash map of external dependencies to force new generation of the deployment
// triggering the Kubernetes rollout as defined when the inputHash changes.`
as a circumstantial consequence and nice to have a comment but I would name the function for what it actually does `ensureDependecyAnnotations` or something like that

PR #649, opened 2020-07-20
BUG 1858400: [Performance] Lease refresh period for machine-api-controllers is too high, causes heavy writes to etcd at idle

Danil-Grigorev:
The machine-api-controller components are refreshing their lease more
than all other components combined. Bringing this to 90s each, will
decrease etcd writes at idle.

All relevant PRs:
- https://github.com/openshift/cluster-api-provider-aws/pull/339
- https://github.com/openshift/cluster-api-provider-azure/pull/152
- https://github.com/openshift/cluster-api-provider-baremetal/pull/88
- https://github.com/openshift/cluster-api-provider-gcp/pull/104
- https://github.com/openshift/cluster-api-provider-openstack/pull/109
- https://github.com/openshift/cluster-api-provider-ovirt/pull/56
- https://github.com/openshift/machine-api-operator/pull/649

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -423,6 +423,7 @@ func newContainers(config *OperatorConfig, features map[string]bool) []corev1.Co
 		"--logtostderr=true",
 		"--v=3",
 		"--leader-elect=true",
+		"--leader-elect-lease-duration=90s",
 		fmt.Sprintf("--namespace=%s", config.TargetNamespace),
 	}
 
--- a/cmd/machine-healthcheck/main.go
+++ b/cmd/machine-healthcheck/main.go
@@ -59,7 +59,7 @@ func main() {
 
 	leaderElectLeaseDuration := flag.Duration(
 		"leader-elect-lease-duration",
-		15*time.Second,
+		90*time.Second,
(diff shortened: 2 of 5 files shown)


enxebre:
@Danil-Grigorev can we also please enforce this via https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/sync.go#L425

Danil-Grigorev:
> @Danil-Grigorev can we also please enforce this via https://github.com/openshift/machine-api-operator/blob/master/pkg/operator/sync.go#L425
> Setting defaults is sane but let’s enforce our opinions on mao.

Agree, this will help when MAO will be revendored in providers next time - https://github.com/openshift/machine-api-operator/pull/649/commits/828c4539b9025a9c48b397037d6b8cc0acacea8a

pkg/operator/sync.go
```diff
@@ -423,6 +423,7 @@ func newContainers(config *OperatorConfig, features map[string]bool) []corev1.Co
 		"--logtostderr=true",
 		"--v=3",
 		"--leader-elect=true",
+		"-leader-elect-lease-duration=90s",
```
enxebre, line 426: why only one dash here while any other command has `--`?

enxebre:
/hold
https://github.com/openshift/machine-api-operator/pull/649#discussion_r457310179 so there's no change to merge by accident as it has both labels now.
@Danil-Grigorev please feel free to unhold after addressing the comment above.

pkg/operator/sync.go
Danil-Grigorev: Copy pasting from CLI is an unhealthy habit 😅Even when it works both ways, will fix in a second

PR #656, opened 2020-07-24
Lock in place golang.org/x/text on v0.3.3

Danil-Grigorev:
Prevent accedential revendoring of security fix, by manual replacement in all subsequent vendor libraries.

Original commit: e07424843170ad44a4a1eee11b2f8817759bd6c4


Danil-Grigorev:
Vendor does not change, replace is not needed.

PR #672, opened 2020-08-07
Move to context driven reconcile thread for MAO

Danil-Grigorev:
Refactor MAO to be non-blocking, and process each event as it appears.
Previous thread would be canceled, and a new one started once a new
event enters the queue.

Previous 5 minutes timeouts on Deployment and DaemonSet rollout are
now merged into global 10 minute context timeout, accounting for both.

The main difference is that there is no longer a soft lock on changes, if
the operator is currently busy processing previous event, while not knowing
about anything for a period ~10 minutes in the worst case.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -33,12 +33,16 @@ const (
 	// a machineconfig pool is going to be requeued:
 	//
 	// 5ms, 10ms, 20ms, 40ms, 80ms, 160ms, 320ms, 640ms, 1.3s, 2.6s, 5.1s, 10.2s, 20.4s, 41s, 82s
-	maxRetries         = 15
+	// maxRetries = 15
+	// Reconcile should take up to 10 minutes to sync. When changed we first have to wait for a full Deployment
+	// rollout (5 minutes) and DaemonSet rollout (another 5 minutes) before everything is synced.
+	maoSyncTimeout     = 10 * time.Minute
 	maoOwnedAnnotation = "machine.openshift.io/owned"
 )
 
 // Operator defines machine api operator.
 type Operator struct {
+	context         context.Context
 	namespace, name string
 
 	imagesFile string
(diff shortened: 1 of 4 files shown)


pkg/operator/operator.go
```diff
@@ -265,35 +269,40 @@ func isMachineWebhook(obj interface{}) bool {
 }
 
 func (optr *Operator) worker() {
-	for optr.processNextWorkItem() {
+	var cancel context.CancelFunc
+	processNext := true
+	optr.context, cancel = context.WithCancel(context.Background())
+	for processNext {
+		processNext, cancel = optr.processNextWorkItem(cancel)
```
michaelgugino, line 276: What is the purpose of getting a cancel function here?

pkg/operator/operator.go
```diff
@@ -265,35 +269,40 @@ func isMachineWebhook(obj interface{}) bool {
...
 	}
 	defer optr.queue.Done(key)
 
 	glog.V(4).Infof("Processing key %s", key)
-	err := optr.syncHandler(key.(string))
-	optr.handleErr(err, key)
+	// Cancel previous reconcile
+	cancel()
+	optr.context, cancel = context.WithTimeout(context.Background(), maoSyncTimeout)
+	go func() {
```
enxebre, line 292: Why is this running in a subroutine??
The code this PR is changing is pretty standard controller skeleton and we have no intent to change it.

If we needed async/multiple calls we could run just more workers https://github.com/openshift/machine-api-operator/blob/c9a1f7584b4bd025e1e99444d0f6281c86a18a1a/cmd/machine-api-operator/start.go#L132 but I see no need for it.

If anything the reconciling loop could be modified to reconcile the different resources here asynchronously https://github.com/openshift/machine-api-operator/blob/0ce2439a9f0ec28dbc45987dc5c28b223e2cfd05/pkg/operator/sync.go#L44
But also I don't see a real need for that. We'll only introduce complexity here is there's a really strong argument in favour of it.

pkg/operator/operator.go
Danil-Grigorev: The current worker gets canceled on a long-running task by the next one, like waiting for rollout, if he arrives earlier then it finishes. Just making sure we account for the changes indicated by the event. 1 worker for one request, but with that, we are giving priority to the state maintenance, rather than a full flow completion. Exposing the `cancel` here provides the next worker an ability to stop his predecessor and take its place.

Danil-Grigorev:
/close

This PR is deprecated by https://github.com/openshift/machine-api-operator/pull/691

PR #675, opened 2020-08-12
BUG 1858400: [Performance] Lease refresh period for machine-api-controllers is too high, causes heavy writes to etcd at idle

Danil-Grigorev:
Reduce default lease retry rate on `30s` which will prevent from heavy writes into etcd at idle, and constrain renew deadline on `90s`.

Inspired by https://github.com/openshift/cloud-credential-operator/pull/231

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -446,7 +446,7 @@ func newContainers(config *OperatorConfig, features map[string]bool) []corev1.Co
 		"--logtostderr=true",
 		"--v=3",
 		"--leader-elect=true",
-		"--leader-elect-lease-duration=90s",
+		"--leader-elect-lease-duration=120s",
 		fmt.Sprintf("--namespace=%s", config.TargetNamespace),
 	}
 
--- a/cmd/machine-healthcheck/main.go
+++ b/cmd/machine-healthcheck/main.go
@@ -20,6 +20,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
 )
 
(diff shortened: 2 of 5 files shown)


Danil-Grigorev:
/hold cancel
Set this on `60s`, so it will try once per renew deadline.

Danil-Grigorev:
After some realization, set it on `30s` and `90s`, so it would retry at least one to twice.

elmiko:
in general i think this is a good patch, but i agree with @michaelgugino that the values should be higher.

michaelgugino:
Okay, after some more discussion, we determined 120/110/90 might be a better fit.

Don't want the leadership to be too long as if the pod gets moved (eg, upgrades) we don't want operations suspended for too long.  120 seconds would be adequate.

pkg/operator/sync.go
```diff
@@ -446,7 +446,7 @@ func newContainers(config *OperatorConfig, features map[string]bool) []corev1.Co
 		"--logtostderr=true",
 		"--v=3",
 		"--leader-elect=true",
-		"--leader-elect-lease-duration=90s",
+		"--leader-elect-lease-duration=120s",
```
elmiko, line 449: this is the only confusing part to me. we set the lease duration in the controller-runtime config options, do we also need to set on the command line?

michaelgugino: We use the default if not specified on the CLI.  This is useful for development/debugging.

elmiko: ahh, perfect. thanks for the explanation Mike!

PR #681, opened 2020-08-19
Bug 1878880: Fix confusing log messages about deleting Node

zaneb:
These two log messages were inconsistent with the format of all the others. In particular it is very confusing to see the message `deleting node <machine name> for machine`.


zaneb:
/retitle Bug 1878880: Fix confusing log messages about deleting Node

PR #689, opened 2020-08-27
Bug 1873234: Manage ownership hand-off of metal3 between MAO and CBO for upgrade and downgrade

sadasu:
In an upgrade scenario, check if CBO is available to take ownership of the metal3 deployment. In that case, back off and let cluster-baremetal-operator (CBO) manage it.

In a downgrade/rollback scenario where the metal3 deployment is
annotated as being owned by cluster-baremetal-operator, but the CBO
has been removed, take back control of the metal3
deployment.

### DIFF
--- a/pkg/operator/baremetal_pod.go
+++ b/pkg/operator/baremetal_pod.go
@@ -9,15 +9,18 @@ import (
 	"golang.org/x/crypto/bcrypt"
 
 	"github.com/golang/glog"
+	osclientset "github.com/openshift/client-go/config/clientset/versioned"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	appsclientv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/utils/pointer"
 )
 
 const (
+	baremetalDeploymentName    = "metal3"
 	baremetalSharedVolume      = "metal3-shared"
 	baremetalSecretName        = "metal3-mariadb-password"
 	baremetalSecretKey         = "password"
@@ -234,13 +237,40 @@ func createMetal3PasswordSecrets(client coreclientv1.SecretsGetter, config *Oper
 	return nil
 }
 
+// Return false on error or if "baremetal.openshift.io/owned" annotation set
+func checkMetal3DeploymentMAOOwned(client appsclientv1.DeploymentsGetter, config *OperatorConfig) (bool, error) {
+	existing, err := client.Deployments(config.TargetNamespace).Get(context.Background(), baremetalDeploymentName, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	}
+	if _, exists := existing.ObjectMeta.Annotations[cboOwnedAnnotation]; exists {
+		return false, nil
+	}
+	return true, nil
+}
+
+// Return true if the baremetal clusteroperator exists
+func checkForBaremetalClusterOperator(osClient osclientset.Interface) (bool, error) {
+	_, err := osClient.ConfigV1().ClusterOperators().Get(context.Background(), cboClusterOperatorName, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
 func newMetal3Deployment(config *OperatorConfig, baremetalProvisioningConfig BaremetalProvisioningConfig) *appsv1.Deployment {
 	replicas := int32(1)
 	template := newMetal3PodTemplateSpec(config, baremetalProvisioningConfig)
 
 	return &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "metal3",
+			Name:      baremetalDeploymentName,
 			Namespace: config.TargetNamespace,
 			Annotations: map[string]string{
 				maoOwnedAnnotation: "",
--- a/pkg/operator/baremetal_test.go
+++ b/pkg/operator/baremetal_test.go
@@ -4,7 +4,10 @@ import (
 	"testing"
 
 	. "github.com/onsi/gomega"
+	osconfigv1 "github.com/openshift/api/config/v1"
+	fakeos "github.com/openshift/client-go/config/clientset/versioned/fake"
 	"golang.org/x/net/context"
+	appsv1 "k8s.io/api/apps/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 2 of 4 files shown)


pkg/operator/baremetal_test.go
```diff
@@ -341,3 +344,159 @@ func TestBaremetalProvisionigConfig(t *testing.T) {
...
+				},
+			},
+			expected: false,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(string(tc.testCase), func(t *testing.T) {
+			var osClient *fakeos.Clientset
+			osClient = fakeos.NewSimpleClientset(tc.clusterOperator)
+			_, err := osClient.ConfigV1().ClusterOperators().Create(context.Background(), tc.clusterOperator, metav1.CreateOptions{})
```
sadasu, line 489: This unit test is unable to create the ClusterOperator for the CBO using this fake client. Please let me know if I am missing something here.

sadasu:
Marking this is as WIP because I am having trouble getting one UT to pass. I have added a comment against the test and part of the test I am having trouble with. I am continuing to work on getting that working. In the meantime, here are the rest of the changes for you to look at. Thanks!

dhellmann:
@sadasu is this the one we need to get into 4.6 to support upgrades to 4.7?

sadasu:
> @sadasu is this the one we need to get into 4.6 to support upgrades to 4.7?

Yes. @dhellmann.

dhellmann:
> > @sadasu is this the one we need to get into 4.6 to support upgrades to 4.7?
> 
> Yes. @dhellmann.

OK, we'll need a bug, as @michaelgugino pointed out.

pkg/operator/baremetal_test.go
sadasu: Fixed it.

pkg/operator/baremetal_pod.go
```diff
@@ -235,13 +238,40 @@ func createMetal3PasswordSecrets(client coreclientv1.SecretsGetter, config *Oper
 	return nil
 }
 
+// Return false on error or if "baremetal.openshift.io/owned" annotation set
+func checkMetal3DeploymentMAOOwned(client appsclientv1.DeploymentsGetter, config *OperatorConfig) (bool, error) {
+	existing, err := client.Deployments(config.TargetNamespace).Get(context.Background(), "metal3", metav1.GetOptions{})
```
dhellmann, line 243: Should the `"metal3"` here be `baremetalDeploymentName`?

sadasu: Yes. Made that change.

pkg/operator/sync.go
```diff
@@ -198,6 +198,23 @@ func (optr *Operator) syncMutatingWebhook() error {
 }
 
 func (optr *Operator) syncBaremetalControllers(config *OperatorConfig, configName string) error {
+	// Stand down if cluster-bare-metal operator has claimed the metal3 deployment
+	// and if the baremetal clusteroperator exists
+	maoOwned, err := checkMetal3DeploymentMAOOwned(optr.kubeClient.AppsV1(), config)
```
michaelgugino, line 203: This check seems redundant.  If the annotation is being applied by the BMO, there's no need to check for this.  If the BMO exists, then it owns the deployment and should implement GetOrCreate for the deployment.

pkg/operator/baremetal_test.go
```diff
@@ -375,7 +378,163 @@ func TestSyncBaremetalControllers(t *testing.T) {
...
+	client := kubeClient.AppsV1()
 
+	testCases := []struct {
+		testCase      string
+		deployment    *appsv1.Deployment
+		expected      bool
+		expectedError bool
+	}{
+		{
+			testCase: "Only maoOwnedAnnotation",
```
michaelgugino, line 397: Checking these annotations in tests and not checking them in real code is not useful.  In any case, I don't think we need the new CBO annotation in any case, at least not in this code base.

pkg/operator/sync.go
dhellmann: This code is trying to ensure a smooth hand-off between the MAO and CBO during upgrade and downgrade. The details are described in the enhancement document at https://github.com/openshift/enhancements/blob/master/enhancements/baremetal/an-slo-for-baremetal.md#upgrade--downgrade--version-skew-strategy

sadasu: We definitely need this change when we downgrade from a 4.7 MAO which does not manage metal3 to 4.6 MAO that does. We are achieving that by checking for the presence of the cboOwnedAnnotation and the presence of CBO's ClusterOperator.

sadasu: In 4.7, it will be removed along with the rest of baremetal code in MAO.

pkg/operator/baremetal_test.go
sadasu: maoOwnedAnnotation was added as part of https://github.com/openshift/machine-api-operator/pull/424. Only cboOwnedAnnotation is added as part of this change. This annotation will not be added to the metal3 deployment by any component in 4.6. It will be added by CBO in 4.7. During some part of the 4.7 dev cycle, both MAO and CBO would be capable of managing the metal3 deployment and during that time, we want CBO to take precedence.

pkg/operator/sync.go
michaelgugino: I disagree this is needed.

Scenario in 4.6:  CBO not deployed, MAO owned, done.
Scenario in 4.7:  CBO gets deployed at some point, MAO doesn't own, MAO quits syncing.
Scenario in 4.7 -> 4.6 downgrade: CBO is initially deployed, MAO does nothing. CBO goes away, MAO starts syncing again.

There's not a scenario that needs an annotation that I can see.

pkg/operator/baremetal_test.go
michaelgugino: There's nothing in the actual product code that checks for this condition, and these tests don't call any MAO function that might result in both being present.  So, these tests are unnecessary.

pkg/operator/sync.go
dhellmann: The 4.6 and 4.7 configurations for metal3 will be different. Therefore if we ever have a time when 2 components are both trying to update the Deployment, they're going to thrash making changes back and forth. There are at least 2 situations where that could come up.

First, during a production upgrade from 4.6 to 4.7 we cannot ensure the order in which the CVO launches the new version of the MAO and the CBO. If the CBO launches before the new MAO, then the definition of the metal3 Deployment would thrash until the MAO is updated. Restarting that pod is extremely disruptive. It has a DHCP server and a relational database and a bunch of other unusual stuff in it. We can't prevent it from restarting entirely, but we want to minimize the restarts, especially in production clusters.

Second, the CBO is not yet fully functional. We want to build it incrementally. As part of doing that, we need a way to run integration test jobs in which the CBO manages metal3 and the MAO does not, so we can ensure we can deploy new clusters from scratch just with the CBO. That test job will tell us when we have the CBO working well enough to be able to remove the metal3 logic from the MAO entirely. Without this change, it would not be possible to have that test job because the 2 components will compete to control the Deployment and we wouldn’t have a way to ensure that the CBO is the actor causing the right outcome (if the Deployment was even stable long enough to fully launch a Pod). We also couldn’t have a non-flakey 4.6->4.7 upgrade test job for bare metal, because we would run into the upgrade sequencing issue described in the previous paragraph.

We did consider some other approaches.

For example, we could wait until near the end of 4.7 and just remove the metal3 management code from the MAO after the CBO is thought to work. But, that leaves us with several sprints/months while the CBO is under development during which it would not be possible to have a stable metal cluster for development because of the thrashing. We also couldn't have those CI jobs. So, we would be waiting until very late in the cycle to do any integration testing for a significant change to the system.

We could also remove that metal3 management code from the MAO as soon as 4.7 opens. That would be even more disruptive, because it would leave us without the ability to stand up metal clusters *at all*, until the CBO is fully functional, because the MAO would no longer launch the metal3 components needed to complete the IPI installation. No one working with bare metal clusters would be able to get any work done until the new component was fully functional.

This is a relatively small change. Nothing in 4.6 is going to set the annotation so it should have no effect (outside of development/CI) until an upgrade occurs. We anticipate being able to remove this logic from the MAO in 4.7 at the same time that we remove the rest of the metal3 management logic, after the CBO is fully functional. But, we think it is needed for now, and the sooner we get it into 4.6 the more likely it will be that we can avoid forcing users to upgrade through a specific 4.6 version before going to 4.7.

If you have other suggestions to address the production upgrade sequencing issue and the incremental development requirement, please let us know.

michaelgugino: > Second, the CBO is not yet fully functional. We want to build it incrementally. As part of doing that, we need a way to run integration test jobs in which the CBO manages metal3 and the MAO does not, so we can ensure we can deploy new clusters from scratch just with the CBO. 

Okay, this makes some sense then.

My only other concern is there are tests for checking if both annotations are on the deployment, but nothing in the MAO making that check.  I don't think there's a scenario where this would actually happen.  Perhaps instead of a new CBO annotation, we should utilize the k/v of a singular annotation such as
"baremetal.openshift.io/managed-by: mao" / "baremetal.openshift.io/managed-by: cbo".  This will prevent PATCH updates from either source resulting in both annotations being present.

sadasu: checkMetal3DeploymentMAOOwned() givens preference to the cboOwnedAnnotation when present and if not, assumes it is owned by MAO. So, with that we are in effect taking care of the case where if both cboOwnedAnnotation and maoOwnedAnnotation are present on a deployment, the cboOwnedAnnotation takes precedence.

pkg/operator/baremetal_test.go
sadasu: As @dhellmann explained it better than I could have:
 "The 4.6 and 4.7 configurations for metal3 will be different. Therefore if we ever have a time when 2 components are both trying to update the Deployment, they're going to thrash making changes back and forth. There are at least 2 situations where that could come up."

In this scenario where MAO and CBO are both trying to make changes to metal3 Deployment, MAO would have added the maoOwnedAnnotation to metal3 deployment with configurationA. And, CBO would also try to update the Deployment with configurationB. In addition, CBO would add the cboOwnedAnnotation to its own metal3 Deployment. With the changes added to syncBaremetalControllers() we will allow CBO to take control and MAO will back off.

Yes, no MAO function will result in both these annotation to be set. But, MAO will have to react correctly to situation where both annotations are set. Hence the test.

pkg/operator/sync.go
sadasu: We could use the k/v of a single annotation just as you pointed out. But, that would essentially ignore the pre-existing annotation added to any deployment owned by MAO. We followed what was done as part of a previous MAO feature https://github.com/openshift/machine-api-operator/pull/424 and used a similar annotation for CBO.

pkg/operator/baremetal_pod.go
```diff
@@ -234,13 +237,40 @@ func createMetal3PasswordSecrets(client coreclientv1.SecretsGetter, config *Oper
 	return nil
 }
 
+// Return false on error or if "baremetal.openshift.io/owned" annotation set
+func checkMetal3DeploymentMAOOwned(client appsclientv1.DeploymentsGetter, config *OperatorConfig) (bool, error) {
+	existing, err := client.Deployments(config.TargetNamespace).Get(context.Background(), baremetalDeploymentName, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return true, nil
```
elmiko, line 245: i'm not understanding why this condition returns true. doesn't this mean that the baremetal deployment cannot be found and thus is not owned by anyone?

sadasu: Currently, in 4.6, MAO owns metal3 deployment by default. So, if MAO cannot find the deployment, it will go ahead and create one in syncBaremetalControllers().

enxebre:
@sadasu @dhellmann why do we need to do any check for the CBO to be available at all?
If baremetal operator is running in 4.6 I'd expect this PR to remove `syncBaremetalControllers` and all baremetal specific code completely. Then during a 4.6 upgrade the CBO just takes over, in a downgrade the old mao code takes over.

Also https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml and https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_04_metal3provisioning.crd.yaml would need to be removed.

markmc:
> @sadasu @dhellmann why do we need to do any check for the CBO to be available or annotations at all?

Covered in some detail by:

https://github.com/openshift/enhancements/blob/master/enhancements/baremetal/an-slo-for-baremetal.md#upgrade--downgrade--version-skew-strategy

and:

https://github.com/openshift/machine-api-operator/pull/689#discussion_r479516868

pkg/operator/sync.go
enxebre: >First, during a production upgrade from 4.6 to 4.7 we cannot ensure the order in which the CVO launches the new version of the MAO and the CBO. 

This could be ensured via prefix https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/upgrades.md#upgrades-and-order

>Second, the CBO is not yet fully functional. We want to build it incrementally...

This make sense to me. I think checking CBO existence  should be enough.

enxebre:
thanks for the pointers @markmc! https://github.com/openshift/machine-api-operator/pull/689#discussion_r480985494

pkg/operator/sync.go
markmc: > > First, during a production upgrade from 4.6 to 4.7 we cannot ensure the order in which the CVO launches the new version of the MAO and the CBO.
> 
> This could be covered and ensured via prefix https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/upgrades.md#upgrades-and-order

Yeah, but it's pretty ugly to put CBO in a later runlevel just for this temporary upgrade handover. Even that link suggests using another mechanism - "Over time, operators should be better at detecting their prerequisites without overcomplicating or risking the predictability of upgrades"

PR #699, opened 2020-09-10
Bug 1875939: vsphere - cloning a Windows 2019 (1909) template fails with Invalid operation for device '0'.

Danil-Grigorev:
Add check for template disk size with tests.

Prevents user from hitting an undescriptive error message `invalid operation for device '0'`,
while attempting to create a machine from a template with a specified disk size larger
then the `machine.spec.diskGB` is requiring.

### DIFF
--- a/pkg/controller/vsphere/reconciler.go
+++ b/pkg/controller/vsphere/reconciler.go
@@ -705,7 +705,7 @@ func taskIsFinished(task *mo.Task) (bool, error) {
 	case types.TaskInfoStateSuccess:
 		return true, nil
 	case types.TaskInfoStateError:
-		return true, nil
+		return true, errors.New(task.Info.Error.LocalizedMessage)
 	default:
 		return false, fmt.Errorf("task: %v, unknown state %v", task.Reference().Value, task.Info.State)
 	}
--- a/pkg/controller/vsphere/reconciler.go
+++ b/pkg/controller/vsphere/reconciler.go
@@ -10,6 +10,7 @@ import (
 	"github.com/google/uuid"
 	machinev1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	vspherev1 "github.com/openshift/machine-api-operator/pkg/apis/vsphereprovider/v1beta1"
+	machineapierros "github.com/openshift/machine-api-operator/pkg/controller/machine"
 	machinecontroller "github.com/openshift/machine-api-operator/pkg/controller/machine"
 	"github.com/openshift/machine-api-operator/pkg/controller/vsphere/session"
 	"github.com/openshift/machine-api-operator/pkg/metrics"
@@ -616,7 +617,13 @@ func getDiskSpec(s *machineScope, devices object.VirtualDeviceList) (types.BaseV
 	}
 
 	disk := disks[0].(*types.VirtualDisk)
-	disk.CapacityInKB = int64(s.providerSpec.DiskGiB) * 1024 * 1024
+	cloneCapacityKB := int64(s.providerSpec.DiskGiB) * 1024 * 1024
+	if disk.CapacityInKB > cloneCapacityKB {
+		return nil, machineapierros.InvalidMachineConfiguration(
+			"can't resize template disk down, initial capacity is larger: %dKiB > %dKiB",
+			disk.CapacityInKB, cloneCapacityKB)
+	}
+	disk.CapacityInKB = cloneCapacityKB
 
 	return &types.VirtualDeviceConfigSpec{
 		Operation: types.VirtualDeviceConfigSpecOperationEdit,
--- a/pkg/controller/vsphere/reconciler_test.go
+++ b/pkg/controller/vsphere/reconciler_test.go
@@ -94,7 +94,23 @@ func TestClone(t *testing.T) {
 
 	password, _ := server.URL.User.Password()
 	namespace := "test"
+	defaultSizeGiB := int32(5)
 	vm := simulator.Map.Any("VirtualMachine").(*simulator.VirtualMachine)
+
+	machine := object.NewVirtualMachine(session.Client.Client, vm.Reference())
+	devices, err := machine.Device(context.TODO())
+	if err != nil {
+		t.Fatalf("Failed to obtain vm devices: %w", err)
+	}
+	disks := devices.SelectByType((*types.VirtualDisk)(nil))
+	if len(disks) < 1 {
+		t.Fatal("Unable to find attached disk for resize")
+	}
+	disk := disks[0].(*types.VirtualDisk)
+	disk.CapacityInKB = int64(defaultSizeGiB) * 1024 * 1024 // GiB
+	if err := machine.EditDevice(context.TODO(), disk); err != nil {
+		t.Fatalf("Can't resize disk for specified size")
+	}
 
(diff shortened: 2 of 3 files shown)


enxebre:
thanks @Danil-Grigorev code changes looks good. Can you elaborate a bit on why the change is needed, why this gives an error from vSphere?
Can you please also elaborate the same raising and issue for `kubernetes-sigs/cluster-api-provider-vsphere` here https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/edd4fcfb973e0534b5f66edf4a851b543997e127/pkg/services/govmomi/vcenter/clone.go#L201

PTAL @jcpowermac

Danil-Grigorev:
> Thanks @Danil-Grigorev code changes looks good. Can you elaborate a bit on why the change is needed, why this gives an error from vSphere?
> Can you please also elaborate the same raising and issue for `kubernetes-sigs/cluster-api-provider-vsphere` here https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/edd4fcfb973e0534b5f66edf4a851b543997e127/pkg/services/govmomi/vcenter/clone.go#L201
> 
> PTAL @jcpowermac

@enxebre I'll raise the issue upstream if there is a similar use-case when a user-specified disk capacity could be given to the clone method there. VSphere does not show an error of any kind anywhere, except the one quoted in the PR description, and the documentation says nothing about that issue. There is a guide on how to shrink disc down manually: https://virtualman.wordpress.com/2016/02/24/shrink-a-vmware-virtual-machine-disk-vmdk/ and a one to do this with specialized software: https://www.vmware.com/support/ws5/doc/ws_disk_shrink.html 

When we clone a VM template we are hitting the same constraints - no way to specify a lower disc size than the one already set on the template. So this PR implements a check, which evaluates whether you could set such size on the machine disk before calling an edit operation on its definition.

Danil-Grigorev:
Raised the issue upstream: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/1030

pkg/controller/vsphere/reconciler.go
```diff
@@ -616,7 +617,13 @@ func getDiskSpec(s *machineScope, devices object.VirtualDeviceList) (types.BaseV
 	}
 
 	disk := disks[0].(*types.VirtualDisk)
-	disk.CapacityInKB = int64(s.providerSpec.DiskGiB) * 1024 * 1024
+	cloneCapacityKB := int64(s.providerSpec.DiskGiB) * 1024 * 1024
+	if disk.CapacityInKB > cloneCapacityKB {
```
JoelSpeed, line 621: Should we not be requiring them to be the same? Does it being smaller not also cause an issue?

pkg/controller/vsphere/reconciler.go
```diff
@@ -698,7 +705,7 @@ func taskIsFinished(task *mo.Task) (bool, error) {
 	case types.TaskInfoStateSuccess:
 		return true, nil
 	case types.TaskInfoStateError:
-		return true, nil
```
enxebre, line 701: why this change?

pkg/controller/vsphere/reconciler.go
```diff
@@ -616,7 +617,13 @@ func getDiskSpec(s *machineScope, devices object.VirtualDeviceList) (types.BaseV
 	}
 
 	disk := disks[0].(*types.VirtualDisk)
-	disk.CapacityInKB = int64(s.providerSpec.DiskGiB) * 1024 * 1024
+	cloneCapacityKB := int64(s.providerSpec.DiskGiB) * 1024 * 1024
+	if disk.CapacityInKB > cloneCapacityKB {
+		return nil, machineapierros.InvalidMachineConfiguration(
+			"can't resize template disk down, initial capacity is larger: %dKB > %dKB",
```
JoelSpeed, line 623: I think the values here are  in `KiB` as opposed to `KB`, we should try to make sure the error message has the correct units

pkg/controller/vsphere/reconciler.go
Danil-Grigorev: We were not reporting an error before that, so even getting this `invalid operation for device '0'` was possible only from vmware ui. I did this as otherwise we lose similar errors.

pkg/controller/vsphere/reconciler_test.go
```diff
@@ -161,6 +163,23 @@ func TestClone(t *testing.T) {
...
+		{
+			testCase: "fail on disc resize down",
+			providerSpec: vsphereapi.VSphereMachineProviderSpec{
+				CredentialsSecret: &corev1.LocalObjectReference{
+					Name: "test",
+				},
+				Workspace: &vsphereapi.Workspace{
+					Server: server.URL.Host,
+				},
+				DiskGiB:  0,
```
JoelSpeed, line 175: Can we make this non-zero or add a separate case? I would expect a zero value to be caught earlier or maybe even produce a different error. While it may not now, we may change this later and that would affect that validity of this test.

pkg/controller/vsphere/reconciler.go
Danil-Grigorev: Oh, and the following `update` call on the machine fails with the inability to find VM by UUID, because the error here causes the VM to be removed automatically. This results in a nil panic in the code, and the controller goes into a crash loop.

pkg/controller/vsphere/reconciler.go
Danil-Grigorev: No, expanding the disk is a supported operation.

pkg/controller/vsphere/reconciler.go
Danil-Grigorev: Good catch, will fix that, but keep in mind it is an upstream issue for storage size pedantic people 
image
This is what vmware UI shows on the template. We should actually multiply by `1000` and call it `GB/MB/KB` etc.

pkg/controller/vsphere/reconciler.go
enxebre: thanks, please move to a different commit.

pkg/controller/vsphere/reconciler_test.go
Danil-Grigorev: Done.

pkg/controller/vsphere/reconciler.go
JoelSpeed: I was looking at this backwards, ignore me

jcpowermac:
Just tested with @Danil-Grigorev cluster in VMC. I can confirm that Windows does clone and started up as expected.

pkg/controller/vsphere/reconciler.go
elmiko: i'm having trouble following the conversation here, should this be moved to another PR or not?

elmiko (commented):
this makes sense to me, i just have a question

pkg/controller/vsphere/reconciler.go
Danil-Grigorev: @elmiko The change is already in its own commit, so this PR is just a holder for both atm. Also Joel hit this issue, and we discussed that the fix will be contained in this PR https://coreos.slack.com/archives/GE2HQ9QP4/p1599829591211300?thread_ts=1599829162.210500&channel=GE2HQ9QP4&message_ts=1599829591.211300

PR #700, opened 2020-09-14
bug 1878129: add CSRs for oc adm inspect

deads2k:
The machine-api runs an approver for CSRs from nodes.  Some failure modes results in the CSRs not being approved.  This change ensures CSRs are collected for determining whether the CSR itself is malformed or whether the approver is malfunctioning.

### DIFF
--- a/install/0000_30_machine-api-operator_12_clusteroperator.yaml
+++ b/install/0000_30_machine-api-operator_12_clusteroperator.yaml
@@ -9,3 +9,6 @@ status:
   versions:
     - name: operator
       version: "0.0.1-snapshot"
+  relatedObjects:
+    - group: "certificates.k8s.io"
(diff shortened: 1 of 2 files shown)


enxebre:
This is actually captured by the machine-approver https://github.com/openshift/cluster-machine-approver/blob/4e5ad9c9430c6b6a383dcba86107dff3e64bd686/status.go#L37

deads2k:
> This is actually captured by the machine-approver https://github.com/openshift/cluster-machine-approver/blob/4e5ad9c9430c6b6a383dcba86107dff3e64bd686/status.go#L37.
> Not sure if we want to have it in two different places.

is the machine approver in a UPI install?

enxebre:
> is the machine approver in a UPI install?
Yes.

PR #701, opened 2020-09-15
Machine: Set ErrorReason when entering Failed phase

zaneb:
When an error returned by the actuator causing the Machine to enter the
Failed phase is of a known type, record the type in the ErrorReason
field.

In the upstream cluster-api, permanent failures were signalled by the
actuator setting the ErrorMessage and ErrorReason fields. A separate
controller reads these fields and updates the Phase.

In the OpenShift implementation
(176084ee02b546e1750a70d392190b7134694dac) this is signalled by the
actuator returning a MachineError, which contains the message and the
reason. However, the Machine controller was ignoring the reason so it
was never set.


elmiko:
looks good to me, thanks for the pr @zaneb ! apologies on the wait
/approve

PR #703, opened 2020-09-17
Bug 1880110: vSphere: Fix double create

michaelgugino:
Currently, some unknown condition can cause the
actuator to double create the clone task.  The
suspected reason is an eventual consistency problem
in the vSphere API which results in the task
object returned by the API as being nil immediately
after creation.

This commit ensures we only ever create a single
clone task.

### DIFF
--- a/pkg/controller/vsphere/reconciler.go
+++ b/pkg/controller/vsphere/reconciler.go
@@ -62,53 +62,53 @@ func (r *Reconciler) create() error {
 		return fmt.Errorf("%v: failed validating machine provider spec: %v", r.machine.GetName(), err)
 	}
 
-	if r.providerStatus.TaskRef != "" {
-		moTask, err := r.session.GetTask(r.Context, r.providerStatus.TaskRef)
+	// We only clone the VM template if we have no taskRef.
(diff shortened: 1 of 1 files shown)


enxebre:
/retest
/approve
@michaelgugino can you please open a ticket and PR for https://github.com/kubernetes-sigs/cluster-api-provider-vsphere
PTAL @alexander-demichev @Danil-Grigorev @JoelSpeed

alexander-demichev (approved):
/lgtm

michaelgugino:
> @michaelgugino can you please open a ticket and PR for https://github.com/kubernetes-sigs/cluster-api-provider-vsphere


@enxebre upstream is not affected by this on latest code: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/de70ff05039df7ab15a4ba11a3f559f4e92f9df7/pkg/services/govmomi/service.go#L61

Their reconcile loop is much different than ours, they always check for existing tasks before taking any other action.

elmiko (approved):
i like the extra comments you added, it helps readability for me.

/lgtm

PR #704, opened 2020-09-18
Bug 1880110: Vsphere: fix clone task log

michaelgugino:
Current log line does not display a task id because
it is not populated in status at this point.

This uses correct task id.


elmiko (approved):
/lgtm

PR #712, opened 2020-09-29
Change PROVISIONING_IP to an IP

derekhiggins:
Change the value of PROVISIONING_IP to an IP and pass
in PROVISIONING_CIDR where needed. This avoids confusion
by only passing in a CIDR where needed.


derekhiggins:
/hold 
until https://github.com/metal3-io/static-ip-manager-image/pull/5 makes it downstream

hardys:
@derekhiggins if we still need this it'll need to be proposed to https://github.com/openshift/cluster-baremetal-operator since https://github.com/openshift/machine-api-operator/pull/767 removed the code it's touching

PR #727, opened 2020-10-08
Turn on auth for ironic json-rpc

derekhiggins:
This wasn't turned on along with the other ironic
APIs as it only listens to localhost. But the container
it runs in is running with host networking so in theory
it could be accesses from outside the ironic pod.

### DIFF
--- a/pkg/operator/baremetal_pod.go
+++ b/pkg/operator/baremetal_pod.go
@@ -25,16 +25,19 @@ const (
 	baremetalSecretName        = "metal3-mariadb-password"
 	baremetalSecretKey         = "password"
 	ironicCredentialsVolume    = "metal3-ironic-basic-auth"
+	ironicrpcCredentialsVolume = "metal3-ironic-rpc-basic-auth"
 	inspectorCredentialsVolume = "metal3-inspector-basic-auth"
 	ironicUsernameKey          = "username"
 	ironicPasswordKey          = "password"
 	ironicHtpasswdKey          = "htpasswd"
 	ironicConfigKey            = "auth-config"
 	ironicSecretName           = "metal3-ironic-password"
+	ironicrpcSecretName        = "metal3-ironic-rpc-password"
 	ironicUsername             = "ironic-user"
 	inspectorSecretName        = "metal3-ironic-inspector-password"
 	inspectorUsername          = "inspector-user"
 	metal3AuthRootDir          = "/auth"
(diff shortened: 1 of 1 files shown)


derekhiggins:
CC @zaneb

pkg/operator/baremetal_pod.go
```diff
@@ -528,12 +554,14 @@ func createContainerMetal3IronicConductor(config *OperatorConfig, baremetalProvi
 		VolumeMounts: []corev1.VolumeMount{
 			sharedVolumeMount,
 			inspectorCredentialsMount,
+			rpcCredentialsMount,
```
zaneb, line 557: Conductor is only the server side, and should not have client credentials.

zaneb (commented):
For this to work, we need to have https://review.opendev.org/742474 in the ironic image. Do we?

pkg/operator/baremetal_pod.go
derekhiggins: Conductor appears to need the credentials to talk to itself, I believe @dtantsur is currently working on removing this requirement.

derekhiggins:
> For this to work, we need to have https://review.opendev.org/742474 in the ironic image. Do we?
I've tested this patch with the upstream metal3 images and all works as expected, I have yet to test it downstream but the inonic we are currently using has this patch.

pkg/operator/baremetal_pod.go
zaneb: TIL, thanks.
Looks like this might be the culprit, it's used by the drivers: https://opendev.org/openstack/ironic/src/branch/master/ironic/conductor/utils.py#L833-L848

dtantsur: I am indeed working on removing this requirement, but it's going to take substantial time.

dtantsur (approved):

zaneb:
This is not going to work with IPv6 yet due to metal3-io/ironic-image#210.
/hold

derekhiggins:
> This is not going to work with IPv6 yet due to [metal3-io/ironic-image#210](https://github.com/metal3-io/ironic-image/issues/210).
> /hold

I see a fix for that has merged, can we remove the hold?

zaneb:
Since everything is moving to the CBO, let's close this in favour of openshift/cluster-baremetal-operator#77.
/close

PR #735, opened 2020-10-23
vSphere: Enter Failed phase on invalid configuration

zaneb:
[According to the enhancement proposal](https://github.com/openshift/enhancements/blob/master/enhancements/machine-api/machine-instance-lifecycle.md#failed), Machine API providers should put the Machine into the `Failed` phase if the configuration is invalid. This was not working for the vSphere provider, due to the correct error being dropped in the error handling chain. This change ensures that the vSphere provider will behave consistently with the other providers in the event of invalid configuration.


alexander-demichev (approved):
Thanks!
/approve

JoelSpeed (approved):
/lgtm

PR #738, opened 2020-10-26
add current profile annotations to CVO manifests

guillaumerose:
This is matches openshift/enhancements#510 and doesn't change existing behavior


deads2k:
The annotation looks correct.

I don't really speak bash for that script.

/assign @enxebre

enxebre:
/approve
PTAL @JoelSpeed

PR #746, opened 2020-11-10
Bug 1889620: Azure disconnected reject publicIP setting

Danil-Grigorev:
Disconnected installations are using different DNS settings, and this is displayed in the resource
```yaml
kind: DNS
metadata:
  name: cluster
spec:
  baseDomain: ***
  privateZone:
    id: <id>
  publicZone: <- this field is absent in disconnected install
    id: <id>
```

The PR goal is to reject an invalid MachineSet configuration when a `publicIP: true`, to prevent Machine from failing.


elmiko (approved):
this makes sense to me
/lgtm

elmiko (approved):
/lgtm

PR #747, opened 2020-11-11
Add missing ErrorCount field to BareMetalHost CRD

andfasano:
Updating MAO BMH crd while CBO migration is still in progress. Related to https://github.com/openshift/cluster-baremetal-operator/pull/64.

### DIFF
--- a/install/0000_30_machine-api-operator_08_baremetalhost.crd.yaml
+++ b/install/0000_30_machine-api-operator_08_baremetalhost.crd.yaml
@@ -306,6 +306,11 @@ spec:
           status:
             description: BareMetalHostStatus defines the observed state of BareMetalHost
             properties:
+              errorCount:
+                description: ErrorCount records how many times the host has encoutered
+                  an error since the last successful operation
+                type: integer
+                default: 0
               errorMessage:
                 description: the last error message reported by the provisioning subsystem
                 type: string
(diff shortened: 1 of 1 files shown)


andfasano:
cc @zaneb @sadasu

install/0000_30_machine-api-operator_08_baremetalhost.crd.yaml
```diff
@@ -705,6 +709,7 @@ spec:
                     type: string
                 type: object
             required:
+            - errorCount
```
zaneb, line 712: Is this field being required going to be a problem for upgrades? i.e. does it mean that after updating the CRD, the baremetal-operator won't be able to write to the status until it is also upgraded?
If so this should probably be fixed upstream.

andfasano: As discussed, I've added a default value for the field to allow smoother upgrade

zaneb:
Upstream change is not accepted yet.
/hold

andfasano:
Change merged upstream https://github.com/metal3-io/baremetal-operator/pull/722

/hold cancel

zaneb:
/retitle Add missing ErrorCount field to BareMetalHost CRD
/retest

andfasano:
Addressed by https://github.com/openshift/machine-api-operator/pull/767

PR #755, opened 2020-11-17
add mapi_machinehealthcheck_short_circuit metric

elmiko:
This change adds the counter metric to track when a MachineHealthCheck
has been short-circuited. It adds helper functions for enabling and
disabling the metric, and an update to the metrics doc showing the new
metric.

### DIFF
--- a/docs/dev/metrics.md
+++ b/docs/dev/metrics.md
@@ -123,8 +123,12 @@ being monitored by `machine-healthcheck-controller`.
 The `mapi_machinehealthcheck_remediation_success_total` metric gives a total count of the successful
 remediation performed by a MachineHealthCheck.
 
+The `mapi_machinehealthcheck_short_circuit` metric indicates when a MachineHealthCheck has been
+short-circuited, a `0` value indicates normal operation, a `1` value indicates a short-circuit.
+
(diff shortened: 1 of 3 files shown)


elmiko:
creating this as a draft while i do some testing
it will need to go after #754

elmiko:
this is ready for review but still needs to go after the other pr
/hold on #754 being merged

elmiko:
updated to add the namespace label and change the Register... func call to Observe...

elmiko:
not sure why this is causing conflicts, once #752 and #754 have merged i will rebase

elmiko:
rebased

elmiko:
rebased after merge of 754
/hold cancel

michaelgugino (changes requested):
Are we sure we want to call it 'short_circuit' as that is a little idiomatic.  Is 'short circuit' described elsewhere?  We use "RemediationAllowed" as the condition, perhaps we should use that for the metric and reverse the boolean?  Or "TooManyUnhealthyNodes" to match the reason and keep the boolean the same?

If our docs say 'short circuit' I suppose that's fine, but we should try to be consistent if not.

michaelgugino (approved):
/lgtm

JoelSpeed (commented):
/approve

PR #759, opened 2020-11-24
Bug 1880161: Set retry timeout for actuator.Update() failures

alexander-demichev:
Set retry timeout for actuator.Update() failures in order to avoid exponential backoff for slow providers.


JoelSpeed (commented):
/approve

PR #765, opened 2020-11-30
Bug 1882608: Add warning when govcloud and spot are enabled in Azure

alexander-demichev:
Add warning when GovCloud and spot are enabled in Azure. Spot instances on the GovCloud are likely not supported.


JoelSpeed (approved):
/approve

elmiko (approved):
/lgtm

PR #766, opened 2020-11-30
Update RBAC to enable access to machineconfig object for the baremetal controller

kirankt:
Some context: The Baremetal platform has a requirement to pass full ignition to the hosts prior to their deployment. We do this to go around the chicken-and-egg issue in situations where more advanced network configurations such as interfaces on vlans and/or bonds need to be configured before the node is deployed. We can access the machineconfigserver (MCS) from everywhere except within the pod network. 

This PR updates the RBAC to enable the machine-api-controllers to gain access to machineconfig and machineconfigpool objects. This is done to enable the baremetal platform's controller, cluster-api-provider-baremetal (CAPBM) to fetch the rendered ignition from the machine-config-operator instead of the MCS. The CAPBM uses this ignition information to create a new secret which is then utilized by the baremetal-operator to build a config drive for booting the OS.

CAPBM PR https://github.com/openshift/cluster-api-provider-baremetal/pull/127

### DIFF
--- a/install/0000_30_machine-api-operator_09_rbac.yaml
+++ b/install/0000_30_machine-api-operator_09_rbac.yaml
@@ -201,6 +201,17 @@ rules:
     - list
     - watch
 
+# The baremetal controller needs access to rendered-ignition
+  - apiGroups:
+      - machineconfiguration.openshift.io
+    resources:
+      - machineconfigs
+      - machineconfigpools
+    verbs:
+      - get
+      - list
+      - watch
+
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role


elmiko (approved):
/lgtm

elmiko:
@kirankt #767 is removing the metal3 stuff from mao, do we still need to carry these permissions changes?

kirankt:
> @kirankt #767 is removing the metal3 stuff from mao, do we still need to carry these permissions changes?

Hi @elmiko . Yes this is needed because its related to the machine-controller for Baremetal (cluster-api-provider-baremetal) which now needs access to MachineConfig objects to create userData secrets per machine/node. Please see: https://github.com/openshift/cluster-api-provider-baremetal/pull/127

This PR has nothing to do with removing metal3 stuff in MAO.

hardys:
This was rebased after being previously approved/lgtm'd and is now passing CI, so re-adding the lgtm

/lgtm

PR #787, opened 2021-01-13
Bug 1915114: allow machine-api-controllers to access configmaps in openshift-config-managed

staebler:
To support custom CA bundles for AWS, the machine-api-controllers need access to the kube-cloud-config ConfigMap in the openshift-config-managed namespace.


Danil-Grigorev (approved):
/lgtm

PR #788, opened 2021-01-13
Bug 1915962: Add exclusion annotation for machine health check on ROKS

csrwng:
Until we switch to cluster profiles for ROKS, we still need to exclude manifests that do not apply. The MachineHealthCheck CRD is not defined in a ROKS cluster.


JoelSpeed (commented):
/approve

Is it documented somewhere (that we could link in the description maybe) why MHC is excluded on ROKS?

csrwng:
@JoelSpeed it is not publicly documented but I can send you links to design docs where this has been discussed (excluding machine related CRDs from ROKS clusters).

PR #795, opened 2021-02-07
4019 fencing  backport mhc external remediation template

mshitrit:
Back porting Machine Health Check external remediation template logic.
Feature Design: [here](https://github.com/openshift/enhancements/blob/8976c2d5945b697667cdcc02bc9c2cdb996af683/enhancements/machine-api/external-remediations.md)


pkg/apis/machine/v1beta1/consts.go
```diff
@@ -66,56 +48,17 @@ const (
...
-	// reconciled.
-	UnsupportedChangeClusterError ClusterStatusError = "UnsupportedChange"
-
-	// CreateClusterError indicates that an error was encountered
-	// when trying to create the cluster.
-	CreateClusterError ClusterStatusError = "CreateError"
-
-	// UpdateClusterError indicates that an error was encountered
-	// when trying to update the cluster.
-	UpdateClusterError ClusterStatusError = "UpdateError"
```
beekhof, line 97: Did you mean to remove these?

pkg/apis/machine/v1beta1/machine_types_test.go
```diff
@@ -92,8 +92,8 @@ func TestRoundTripMachine(t *testing.T) {
 		// losing data
 		spec := &MachineSpec{}
 		status := &MachineStatus{}
-		fuzzer.Fuzz(spec)
-		fuzzer.Fuzz(status)
+		machineFuzzer.Fuzz(spec)
+		machineFuzzer.Fuzz(status)
```
beekhof, line 96: Unless there was a compelling reason, I would not include cosmetic changes like this

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -179,17 +182,18 @@ func (r *ReconcileMachineHealthCheck) Reconcile(ctx context.Context, request rec
 
 	// health check all targets and reconcile mhc status
 	currentHealthy, needRemediationTargets, nextCheckTimes, errList := r.healthCheckTargets(targets, mhc.Spec.NodeStartupTimeout.Duration)
-	mhc.Status.CurrentHealthy = &currentHealthy
+	healthyCount := len(currentHealthy)
+	mhc.Status.CurrentHealthy = &healthyCount
```
beekhof, line 186: We look to be changing the meaning/contents of a status field.  Can you please reach out to the cloud team for the rules about if/when/how API breaking changes are permitted?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -686,7 +795,7 @@ func minDuration(durations []time.Duration) time.Duration {
 		return time.Duration(0)
 	}
 
-	minDuration := time.Duration(1 * time.Hour)
+	minDuration := time.Hour
```
beekhof, line 798: Are we sure these calls are exactly equivalent?

beekhof (changes requested):
Good progress, but lets not introduce any changes we don't need to.
My major concern is the changes to existing API fields

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
mshitrit: I believe so.
I actually caught it because of a go lint warning.
time.Hour is of Type Duration so conversion is not necessary, and since Duration is fundamentally an int64 I see no point for multiplying it by 1.

pkg/apis/machine/v1beta1/machine_types_test.go
mshitrit: I've noticed a go lint warning on these, and since I'm a big advocate of the [Boy Scout rule](https://biratkirat.medium.com/step-8-the-boy-scout-rule-robert-c-martin-uncle-bob-9ac839778385) :-) - I try to follow it myself.
However it is defiantly something I can change back in case there is any objection from cloud-team.

pkg/apis/machine/v1beta1/consts.go
mshitrit: Yep.
They weren't being used in MAO so I assumed it is a pretty safe bet to remove them.
Of course I could be wrong here - so looking for cloud team input.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
mshitrit: The change was not in status field but in the return value of healthCheckTargets method.
Instead of returning a count(number) of healthy targets it returns a slice of the actual objects.
The reason for this change is because those elements are needed later as an input for cleanEMR method.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -238,16 +242,9 @@ func (r *ReconcileMachineHealthCheck) Reconcile(ctx context.Context, request rec
...
-		klog.V(3).Infof("Reconciling %s: meet unhealthy criteria, triggers remediation", t.string())
-		if err := t.remediate(r); err != nil {
-			klog.Errorf("Reconciling %s: error remediating: %v", t.string(), err)
-			errList = append(errList, err)
-		}
-	}
-
+	errList = r.remediate(ctx, needRemediationTargets, errList, mhc)
+	// deletes External Machine Remediation for healthy machines - indicating remediation was successful
+	r.cleanEMR(ctx, currentHealthy, mhc)
```
Danil-Grigorev, line 247: I think it should belong entirely to `externalRemediation` logic block, so the default internal remediation flow is unchanged.

pkg/apis/machine/v1beta1/machine_types_test.go
JoelSpeed: IMO let's keep the diff minimal for now. I appreciate boy scout rules, but, this PR is big enough as it is ;)

pkg/apis/machine/v1beta1/machinehealthcheck_types.go
```diff
@@ -80,6 +80,15 @@ type MachineHealthCheckSpec struct {
 	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$"
 	// +kubebuilder:validation:Type:=string
 	NodeStartupTimeout metav1.Duration `json:"nodeStartupTimeout,omitempty"`
+
+	// RemediationTemplate is a reference to a remediation template
```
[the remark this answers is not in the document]
JoelSpeed: How will the controller be able to create new objects from the template? Who is going to be adding the appropriate RBAC? Is that going to be part of the default set of RBAC or will another component grant permission to the MHC service account in their own way?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -323,11 +437,7 @@ func (r *ReconcileMachineHealthCheck) reconcileStatus(baseToPatch client.Patch,
 
 // healthCheckTargets health checks a slice of targets
 // and gives a data to measure the average health
-func (r *ReconcileMachineHealthCheck) healthCheckTargets(targets []target, timeoutForMachineToHaveNode time.Duration) (int, []target, []time.Duration, []error) {
-	var nextCheckTimes []time.Duration
-	var errList []error
-	var needRemediationTargets []target
-	var currentHealthy int
+func (r *ReconcileMachineHealthCheck) healthCheckTargets(targets []target, timeoutForMachineToHaveNode time.Duration) (currentHealthy []target, needRemediationTargets []target, nextCheckTimes []time.Duration, errList []error) {
```
JoelSpeed, line 440: Please don't use named returns, it's confusing and error prone. Be explicit in each return statement please

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -2893,3 +2921,220 @@ func TestGetIntOrPercentValue(t *testing.T) {
...
+			t.Errorf("Test case: %s. Expected: %v, got: %v", tc.node.Name, tc.expected.result, result)
+		}
+	}
+	g := NewWithT(t)
+	if tc.expectedStatus != nil {
+		mhc := &mapiv1beta1.MachineHealthCheck{}
+		g.Expect(r.client.Get(ctx, request.NamespacedName, mhc)).To(Succeed())
+		g.Expect(tc.expectedStatus).To(MatchMachineHealthCheckStatus(&mhc.Status))
+	}
+}
```
JoelSpeed, line 3001: This all seems to have been moved without real need, cosmetic changes again? Can we move it back or at least keep cosmetic changes to a separate commit, currently this makes it hard to review with so many changes in this one commit

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -2893,3 +2921,220 @@ func TestGetIntOrPercentValue(t *testing.T) {
...
+	cw.Mock.On("Delete", ctx, obj, opts).Return(nil)
+	cw.Mock.Called(ctx, obj, opts)
+	return cw.Client.Delete(ctx, obj, opts...)
+}
+
+func (cw *clientWrap) Get(ctx context.Context, key client.ObjectKey, obj client.Object) error {
+	cw.Mock.On("Get", ctx, key, obj).Return(nil)
+	cw.Mock.Called(ctx, key, obj)
+	return cw.Client.Get(ctx, key, obj)
+}
```
JoelSpeed, line 3123: I'm not a fan of using spy clients like this, normally we would expect to just look up the object after the code under test has run. It's more explicit and helps us understand exactly what the input/output is.

Eg, a test should look something like this
```
// Create fake client
// Run remediation code
// Use fake client to check external remediation object now exists
// Check external remediation object matches the expectation 
```

pkg/util/external/util.go
```diff
@@ -0,0 +1,116 @@
...
+	// +optional
+	Labels map[string]string
+}
+
+func GenerateTemplate(in *GenerateTemplateInput) (*unstructured.Unstructured, error) {
+	template, found, err := unstructured.NestedMap(in.Template.Object, "spec", "template")
+	if !found {
+		return nil, errors.Errorf("missing Spec.Template on %v %q", in.Template.GroupVersionKind(), in.Template.GetName())
+	} else if err != nil {
+		return nil, errors.Wrapf(err, "failed to retrieve Spec.Template map on %v %q", in.Template.GroupVersionKind(), in.Template.GetName())
```
JoelSpeed, line 81: We use the standard library errors package in this repo, so by using this new errors package, you are creating incompatible errors. Also, is this error ever actually unwrapped anywhere? If not, there's no need to introduce the overhead of wrapping it.

```suggestion
		return nil, fmt.Errorf("missing Spec.Template on %v %q", in.Template.GroupVersionKind(), in.Template.GetName())
	} else if err != nil {
		return nil, fmt.Errorf("failed to retrieve Spec.Template map on %v %q: %v", in.Template.GroupVersionKind(), in.Template.GetName(), err)
```

JoelSpeed (commented):
I haven't given this a thorough review yet, but I've added a few comments. 
While I appreciate the effort to clean up the code, please either remove cosmetic only changes or put them into a separate commit. Currently it is very hard to review this as I can't tell what has genuinely change vs what has just been renamed or moved

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -238,16 +242,9 @@ func (r *ReconcileMachineHealthCheck) Reconcile(ctx context.Context, request rec
...
-	// remediate
-	for _, t := range needRemediationTargets {
-		klog.V(3).Infof("Reconciling %s: meet unhealthy criteria, triggers remediation", t.string())
-		if err := t.remediate(r); err != nil {
-			klog.Errorf("Reconciling %s: error remediating: %v", t.string(), err)
-			errList = append(errList, err)
-		}
-	}
-
+	errList = r.remediate(ctx, needRemediationTargets, errList, mhc)
```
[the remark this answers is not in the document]
mshitrit: Not sure I follow.
Before this change elements were appended to error list inside the loop, and now they are appended in the remediate method.
Creation a new list in the remediate method will cause any existing elements (prior to the remediate call) in errList to be lost.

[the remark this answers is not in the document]
JoelSpeed: It's easier to understand what is going on if you do the append here, this is what I would do in this case
```suggestion
	errList = append(errList, r.remediate(ctx, needRemediationTargets, mhc)...)
```

pkg/util/testing/testing.go
```diff
@@ -38,6 +39,48 @@ func NewSelectorFooBar() *metav1.LabelSelector {
...
+				"name":      "Machine",
+				"namespace": Namespace,
+			},
+			"spec": map[string]interface{}{
+				"size": "3xlarge",
+			},
+		},
+	}
+}
+
```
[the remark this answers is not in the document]
mshitrit: I don't follow.
Those methods are used in pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go:461

pkg/apis/machine/v1beta1/machinehealthcheck_types.go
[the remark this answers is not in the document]
mshitrit: There was no change in this logic from upstream.
In a nutshell the existence of a template means that the proper CRD exist so a CR (EMR) can be created.
If you want to read more, [here](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20191030-machine-health-checking.md) is the upstream md file

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
mshitrit: You are right. my apologies.
In my the next PR (the short-circuit back off) I've already concentrated all the tidy changes into one  commit.

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
mshitrit: I must disagree here.
In the pattern you suggest I don't like the double invocation of the client.
However I am a guest in your repo :smile: , so let me know if it something you feel strongly about - if so I'll try to change the pattern.

pkg/apis/machine/v1beta1/machinehealthcheck_types.go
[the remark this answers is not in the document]
JoelSpeed: Just because a CRD exists, doesn't mean a user has permission to create resources 😉  Something will need to give the service account for the Machine Health Check controller permissions to be able to create these resources, it could be that we just add those to the defaults (which I think is perfectly acceptable), but to do this we need to know ahead of time all of the types that could possible be referred to here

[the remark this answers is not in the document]
mshitrit: I understand.
I actually had to do it on my e2e test.
However I'm not sure if it is something we can know ahead - since I'm pretty sure that the CRDs are defined by the third  party (external).
It might be something that needs to be updated after coordination with the external user.
/cc @n1r1  what do you think ?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
mshitrit: I defiantly see your point, but I'm not sure the alternative is better.

The major thing I'm bothered with is that it'll will force us to add `currentHealthy` to the remediate method input params - which does not make a lot of sense to me.

Another minor issue - currently `externalRemediation` and `internalRemediation` takes a single target as input - this change will also cause a refactoring for those method to take the whole slice.

I don't have a strong opinion either way - but I thought it worth bringing these things up before making the change.
Let me know what you think.

pkg/util/external/util.go
mshitrit: It turns out that we do indeed unwrap the error.
We do it once we call `apierrors.IsNotFound(err)`
However the good news is that `fmt` can also wrap the error using the %w operator so we are still in the clear.

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
JoelSpeed: To each their own I guess, to me it makes sense to call the client to see what it thinks the state is (we aren't testing the client here, so we trust the fake to give us the right result)

I'd prefer we keep to the existing patterns in the repo ultimately, so there's a lower burden for maintenance of the tests in the future

pkg/apis/machine/v1beta1/machinehealthcheck_types.go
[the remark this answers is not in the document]
n1r1: Good point.
Unfortunately we can not list all the possible resources as some of them may still not exist and we want this to be generic and plugable (such that new remediation provider could add its own logic without any change in MHC/MAO)

I believe the way it works upstream is that the remediation template and remediation request are defined with apiGroup that already exists in their RBAC, and with `resources: '*'`.

I think we might want to go in a slightly different way. We could create a specific apiGroup that is only intended for external remediation, e.g. `ext-remediation.machine.openshift.io` and update the RBAC to allow creation of any resource with that apiGroup.

WDYT?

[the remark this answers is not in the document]
JoelSpeed: I guess this comes down to, how many do we realistically expect to support within OpenShift? If it's 2/3, then we can probably add those to the RBAC in this repo as we invent them. If it's more, perhaps the operators responsible for those external remediation strategies should be responsible for adding a binding for the service account?

Eg, could the cluster-baremetal-operator add a role/binding to give the correct permissions if the platform is baremetal and it's deploying some external remediation controller

Also, are external remediation templates always expected to be in a particular namespace? Or would they vary across a cluster/per external remediation implementation? This could affect how we want to handle the RBAC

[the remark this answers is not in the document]
n1r1: > I guess this comes down to, how many do we realistically expect to support within OpenShift? If it's 2/3, then we can probably add those to the RBAC in this repo as we invent them. If it's more, perhaps the operators responsible for those external remediation strategies should be responsible for adding a binding for the service account?

Yeah, but this creates unnecessary coupling. Once we get this PR merged, a new external remediation controllers could be installed via OLM and just work, without any need to update something in core OCP. 
What would be the downside to implement the suggestion I made?

> Eg, could the cluster-baremetal-operator add a role/binding to give the correct permissions if the platform is baremetal and it's deploying some external remediation controller

From my point of view, external remediation is a generic infrastructure. It is not tied to a specific platform. 
I prefer to have a complete decoupling between MHC and the external remediation controller.
There's a known api, and whoever wants can just implement it.

> Also, are external remediation templates always expected to be in a particular namespace? Or would they vary across a cluster/per external remediation implementation? This could affect how we want to handle the RBAC

Yes. They are required to be in the namespace of the unhealthy machine (which is `openshift-machine-api` in this case)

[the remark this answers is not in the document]
JoelSpeed: > What would be the downside to implement the suggestion I made?

IMO forcing all external remediations into a single API group will limit us severely. This means that only operators built by OpenShift and maintained by OpenShift would be able to be used, ie, no community remediation templates, if that's not an issue then sure, we can go down this route

> If it's more, perhaps the operators responsible for those external remediation strategies should be responsible for adding a binding for the service account?

I think my preference would be this kind of approach, it would be consistent across both openshift and community remediation methods at least

> From my point of view, external remediation is a generic infrastructure. It is not tied to a specific platform.
I prefer to have a complete decoupling between MHC and the external remediation controller.
There's a known api, and whoever wants can just implement it.

Yes I appreciate that and definitely want it to be generic, but that doesn't fit RBAC very well, so we need to come up with some solution of allowing the MHC permission for the groups that are appropriate.

pkg/util/external/util.go
JoelSpeed: Intersting point/aside, `apierrors.IsNotFound` uses the `errors` standard library package to unwrap the error https://github.com/kubernetes/apimachinery/blob/569bd20062c9d996b1b3d5fa251443413e63dcb7/pkg/api/errors/errors.go#L649-L654
The two libraries are incompatible so you have to wrap using `fmt` and `%w` for this to work properly

pkg/apis/machine/v1beta1/machinehealthcheck_types.go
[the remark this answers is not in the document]
n1r1: > > What would be the downside to implement the suggestion I made?
> 
> IMO forcing all external remediations into a single API group will limit us severely. This means that only operators built by OpenShift and maintained by OpenShift would be able to be used, ie, no community remediation templates, if that's not an issue then sure, we can go down this route

Good point.
We could take an hybrid approach, where openshift remediators can be seamlessly added while community operators would need to do some extra step (either patch the SA or update the RBAC in this repo)

> 
> > If it's more, perhaps the operators responsible for those external remediation strategies should be responsible for adding a binding for the service account?
> 
> I think my preference would be this kind of approach, it would be consistent across both openshift and community remediation methods at least

The service account could be different in different distributions, isn't it? for example, MHC SA could be `ocp-mhc-sa` and in CAPI based clusters it could be `capi-mhc-sa`, so the remediator will need to patch all "known" SAs.

Also, I'm not sure if someone will be able to remember that changing MHC service account breaks other external components.

Maybe remediators should expose some config value. it could be either the `apiGroup` to be used or the `ServiceAccount` to be patched.

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
mshitrit: NP, I've made the changes :+1:

pkg/apis/machine/v1beta1/machinehealthcheck_types.go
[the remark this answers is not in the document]
n1r1: a new idea came up to my mind - what about just having a non openshift api group which is specifically for external remediation?
for example `external-remediation.io`. We could propose that upstream as well, such that community remediators could work the same on OCP and upstream, by using that group.

WDYT?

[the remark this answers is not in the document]
elmiko: just wanted to add another idea here. we could create the default build with a tightly focused api group for the external remediation template, but give the user clear instructions on how to override/expand that initial group. basically as a day 2 operation they could create their own special EMRs and load those in, but by default it would work "out of the box" with a reasonable setting.

> We could take an hybrid approach, where openshift remediators can be seamlessly added while community operators would need to do some extra step (either patch the SA or update the RBAC in this repo)

+1, i think this is saying something similar

this situation is also similar to how we have modified the cluster-api provider in the cluster-autoscaler. it's a different problem there, but perhaps a similar solution. we compile the autoscaler to know about our resrouce groups, but it is possible for the autoscaler to be reconfigured by environment variable to use a different api group. this allows us to share code with upstream capi and also keep our openshift types for when we deploy.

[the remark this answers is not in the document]
n1r1: Thanks @elmiko 

We discussed this issue with the CAPI community, and CAPI is already using [RBAC aggregated label](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles) which allows external components to add their RBAC rules.

I believe this is an elegant way to solve this problem and it's always good to be aligned with upstream.

pkg/apis/machine/v1beta1/consts.go
Danil-Grigorev: I think these changes should at least belong to a separate commit or a different PR. Agree - we don't use these values, we don't have and planning to have cluster resource in MAPI, yet it makes hard to approve such change.

pkg/util/external/util.go
```diff
@@ -0,0 +1,116 @@
...
+	// Set the object APIVersion.
+	if to.GetAPIVersion() == "" {
+		to.SetAPIVersion(in.Template.GetAPIVersion())
+	}
+
+	// Set the object Kind and strip the word "Template" if it's a suffix.
+	if to.GetKind() == "" {
+		to.SetKind(strings.TrimSuffix(in.Template.GetKind(), TemplateSuffix))
+	}
+	return to, nil
```
elmiko, line 115: do the `Labels` from `GenerateTemplateInput` need to be applied in here somewhere?

elmiko (commented):
this generally makes sense to me, i have a few questions just to make sure i'm not missing anything.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -264,6 +261,124 @@ func (r *ReconcileMachineHealthCheck) Reconcile(ctx context.Context, request rec
...
+			// Issue a delete for remediation request.
+			if err := r.client.Delete(ctx, obj); err != nil && !apierrors.IsNotFound(err) {
+				klog.Errorf("failed to delete %v %q for Machine %q: %v", obj.GroupVersionKind(), obj.GetName(), t.Machine.Name, err)
+			}
+		}
+	}
+}
+
+func (r *ReconcileMachineHealthCheck) externalRemediation(ctx context.Context, m *mapiv1.MachineHealthCheck, t target) error {
+	klog.Infof(" %s: start external remediation logic", t.string())
```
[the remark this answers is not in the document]
mshitrit: I think you are right - I'll change it :+1:

pkg/util/external/util.go
mshitrit: That's a very good point.
From what I can tell `Labels` are not relevant for either `GenerateTemplateInput` or for the actual `EMR` (which is created here), so I don't know if they should be copied or not.
IMO we should remove it altogether to avoid confusion.
@n1r1 what do you think,  Is there a purpose for the labels here that I'm missing  ?

n1r1: I don't see any immediate need for that but I can understand where it comes from and why someone might expect that.
I'm not sure what's the correct behavior here but for the sake of being consistent with upstream implementation (which doesn't add labels AFAIU), I would keep at as is.

If we think this should be changed we might want to do that upstream first.

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -477,50 +447,104 @@ func TestReconcile(t *testing.T) {
...
-			objects = append(objects, tc.mhc)
-			if tc.machine != nil {
-				objects = append(objects, tc.machine)
-			}
-			objects = append(objects, tc.node)
-			recorder := record.NewFakeRecorder(2)
-			r := newFakeReconcilerWithCustomRecorder(recorder, objects...)
+		recorder := record.NewFakeRecorder(2)
+		r := newFakeReconcilerWithCustomRecorder(recorder, buildRunTimeObjects(tc)...)
+		assertBaseReconcile(t, tc, ctx, r)
```
JoelSpeed, line 452: This (and the other loops in this file) still need to be wrapped in a `t.Run` otherwise there is only one test, not a parent with children. If we do it like this, the first test failure will fail the whole table, and it will be hard to work out exactly which one.

Please restore the `t.Run` here and in the external remediation test too

JoelSpeed (commented):
I've given this a review and am happy to merge, apart from the tests. Can you please take a look at my comment and fix up the subtest logic and then we can get this merged.

Thanks @mshitrit

beekhof:
Looks like the test failures are unrelated...

`Cluster operator authentication Degraded is True with IngressStateEndpoints_MissingSubsets::OAuthClientsController_SyncError::OAuthServerDeployment_PreconditionNotFulfilled::OAuthServerRouteEndpointAccessibleController_SyncError::OAuthServerServiceEndpointAccessibleController_SyncError::OAuthServerServiceEndpointsEndpointAccessibleController_SyncError::Route_InvalidCanonicalHost::WellKnownReadyController_SyncError: IngressStateEndpointsDegraded: No subsets found for the endpoints of oauth-server"`

/retest

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
mshitrit: Hi @JoelSpeed,
Thanks Great comment :+1: 
As far as I could see the other loop was already ok so I I've only changed the one you've commented about.
```for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			recorder := record.NewFakeRecorder(2)
			r := newFakeReconcilerWithCustomRecorder(recorder, buildRunTimeObjects(tc)...)
			assertBaseReconcile(t, tc, ctx, r)
			//assertExternalRemediation
			assertExternalRemediation(t, tc, ctx, r)

		})
	}```

File: pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go [hunk 16]
```diff
@@ -477,50 +447,105 @@ func TestReconcile(t *testing.T) {
...
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			recorder := record.NewFakeRecorder(2)
+			r := newFakeReconcilerWithCustomRecorder(recorder, buildRunTimeObjects(tc)...)
+			assertBaseReconcile(t, tc, ctx, r)
+			//assertExternalRemediation
```
JoelSpeed, line 546: What's this comment? Do we need this?

mshitrit: Nope. it's quite redundant.
I've removed it.

JoelSpeed (approved):
/approve

elmiko (approved):
thanks @mshitrit
/lgtm

JoelSpeed:
/approve cancel

The failures in CI are related to the RBAC changes in this PR, the rule aggregation isn't working properly and as such MAO is not working, check the logs for further details, eg https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_machine-api-operator/795/pull-ci-openshift-machine-api-operator-master-e2e-aws/1380127742913155072/artifacts/e2e-aws/gather-extra/artifacts/pods/openshift-machine-api_machine-api-operator-5468cb8c8d-2l8sh_machine-api-operator.log

JoelSpeed (commented):
/approve

JoelSpeed:

Test failures seem to be unrelated to the changes in this PR

elmiko:
i left a comment on the other PR, https://github.com/openshift/machine-api-operator/pull/846#pullrequestreview-632748735

sttts:
When this got merged, aws-serial tests started failing: https://testgrid.k8s.io/redhat-openshift-ocp-release-4.8-blocking#periodic-ci-openshift-release-master-nightly-4.8-e2e-aws-serial

PR #798, opened 2021-02-09
MachineWithNoRunningPhase: Doc node drain related cause

2uasimojo:
We recently debugged a `MachineWithNoRunningPhase` alert and discovered it was caused by slow draining. Add this scenario to the Possible Causes.


wking (approved):

2uasimojo:
/retest

@wking @michaelgugino is this still a reasonable change?

PR #803, opened 2021-02-15
fix logging message

relyt0925:


elmiko (approved):
thanks for the pr @relyt0925 !
/approve

PR #804, opened 2021-02-17
Bug 1929645: Remove "kubevirt-machine-controllers" image from release 4.7

nirarg:
This change is done because of the desicion not to include this image in the release
because it defined in "dev preview" stage


elmiko (approved):
thanks for the patch @nirarg
/lgtm

PR #814, opened 2021-03-02
Short circuiting backoff

mshitrit:
This PR is the implementation of [this](https://github.com/openshift/enhancements/pull/673) enhancement.
The main purpose is to allow create a large enough time frame for an admin to fix certain issues by delaying the remediation time for machines that are in a failed stated and has no node.


pkg/apis/machine/v1beta1/machinehealthcheck_types.go
```diff
@@ -80,6 +80,17 @@ type MachineHealthCheckSpec struct {
 	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$"
 	// +kubebuilder:validation:Type:=string
 	NodeStartupTimeout metav1.Duration `json:"nodeStartupTimeout,omitempty"`
+
+	//Failed Machines older than this duration and are without a nodeRef or a providerID
+	//will be considered to have passed the time period allocated for manual fix and will be remediated.
+	//Expects an unsigned duration string of decimal numbers each with optional
+	//fraction and a unit suffix, eg "300ms", "1.5h" or "2h45m".
+	//Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
```
JoelSpeed, line 88: Might be worth updating the node startup comment as well so that it's clear that won't apply to failed machines?
```suggestion
	// Failed Machines that are older than this value and are without a nodeRef or a providerID
	// will be considered to have passed the time period allocated for a manual fix and will be remediated.
	// Expects an unsigned duration string of decimal numbers each with optional
	// fraction and a unit suffix, eg "300ms", "1.5h" or "2h45m".
	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
```

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -68,6 +68,26 @@ const (
...
+	namespace string
+	recorder  record.EventRecorder
+}
+
+type target struct {
+	Machine mapiv1.Machine
+	Node    *corev1.Node
+	MHC     mapiv1.MachineHealthCheck
+}
+
```
JoelSpeed, line 90: This seems to have moved without changes, what happened there? Can we put it back to reduce the diff please?

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -613,12 +619,17 @@ func (t *target) nodeName() string {
...
 }
 
-func (t *target) needsRemediation(timeoutForMachineToHaveNode time.Duration) (bool, time.Duration, error) {
+func (t *target) needsRemediation(timeoutForMachineToHaveNode time.Duration, failedNodeStartupTimeout time.Duration) (bool, time.Duration, error) {
 	var nextCheckTimes []time.Duration
 	now := time.Now()
 
 	// machine has failed
 	if derefStringPointer(t.Machine.Status.Phase) == machinePhaseFailed {
+		//backoff to allow user to fix the error
```
JoelSpeed, line 628: Let's make this more descriptive of why and what happened here
```suggestion
		// A machine that fails without getting a NodeRef of ProviderID means that there was a fatal configuration error.
		// Typically remediating this machine will not actually resolve the issue.
		// Delay remediation for this machine to allow manual intervention to resolve the configuration issue.
```

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -238,11 +234,18 @@ func TestReconcile(t *testing.T) {
...
 
 	machineHealthCheckNegativeMaxUnhealthy := maotesting.NewMachineHealthCheck("machineHealthCheckNegativeMaxUnhealthy")
 	negativeOne := intstr.FromInt(-1)
 	machineHealthCheckNegativeMaxUnhealthy.Spec.MaxUnhealthy = &negativeOne
 
+	//Failed Machine
+	machineWithFailedStatus := maotesting.NewMachine("machineWithFailedStatus", "")
+	failedVar := machinePhaseFailed
+	machineWithFailedStatus.Status.Phase = &failedVar
+	machineWithFailedStatus.CreationTimestamp = metav1.Now()
```
JoelSpeed, line 247: Do we have a test case that covers a failed machine with and without noderef? Would be good to make sure we have both test cases (and also make the description of this more descriptive to show exactly what this is for, i.e. Failed Machine with no Node)

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -494,6 +516,9 @@ func TestReconcile(t *testing.T) {
 				},
 			}
 			result, err := r.Reconcile(ctx, request)
+			if &result == nil {
+				t.Errorf("Test case: %s. Expected: non nil result error, got: nil", tc.node.Name)
+			}
```
JoelSpeed, line 521: Since the return here is not a point, it's impossible to get a nil, this check isn't needed. What are you trying to guard against?

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -1827,14 +1855,14 @@ func TestNeedsRemediation(t *testing.T) {
...
-			expectedNextCheck:           time.Duration(1 * time.Minute), // 300-200 rounded
+			expectedNextCheck:           time.Minute, // 300-200 rounded
 			expectedError:               false,
 		},
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.testCase, func(t *testing.T) {
-			needsRemediation, nextCheck, err := tc.target.needsRemediation(tc.timeoutForMachineToHaveNode)
+			needsRemediation, nextCheck, err := tc.target.needsRemediation(tc.timeoutForMachineToHaveNode, 0)
```
JoelSpeed, line 1865: Should we not adjust this to cater for having some failed machines? This is currently just skipping over the logic that was added in this PR

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -2515,9 +2543,9 @@ func TestHealthCheckTargets(t *testing.T) {
 		recorder := record.NewFakeRecorder(2)
 		r := newFakeReconcilerWithCustomRecorder(recorder)
 		t.Run(tc.testCase, func(t *testing.T) {
-			currentHealhty, needRemediationTargets, nextCheckTimes, errList := r.healthCheckTargets(tc.targets, tc.timeoutForMachineToHaveNode)
-			if currentHealhty != tc.currentHealthy {
-				t.Errorf("Case: %v. Got: %v, expected: %v", tc.testCase, currentHealhty, tc.currentHealthy)
+			currentHealthy, needRemediationTargets, nextCheckTimes, errList := r.healthCheckTargets(tc.targets, tc.timeoutForMachineToHaveNode, 0)
```
JoelSpeed, line 2546: Same here, by setting this to 0 we are skipping over the current logic

JoelSpeed (commented):
Changes look ok, I've added a bunch of suggestions. I'd like to see some more thorough unit testing of this feature before we merge though, at the moment it's not obvious that we have test cases in place that cover the new behaviour

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
mshitrit: To keep the class more readable consts, vars and structs are aligned to the top instead of being scattered all over.
In order to reduce the diff I've placed all the non-functional changes into a single commit - so the functional changes can be reviewed without this static noise.

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
mshitrit: Basically I was fixing the lint warning.
You are right, it is not needed here :+1:

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
mshitrit: I don't think so.
Please note that- the code you are referring is part of `TestNeedsRemediation` method, where our tests are in `TestReconcile` method.
This change was required because of the change in the method signature.

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
mshitrit: Same idea here.
This is `TestHealthCheckTargets` method where our tests are in `TestReconcile` method.

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
mshitrit: Good catch.
Actually we only have a case without the noderef.
I'll add a case with a noderef (and a providerID)

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
beekhof: I would move them to a whole other PR

elmiko (commented):
this mostly makes sense to me, just a question about the defaults.

JoelSpeed (commented):
Because I went through this commit by commit, there may be some comments that are fixed by later commits. 

Could you do a rebase and fix up the commits and commit messages so that the commits are atomic and have good descriptions

Please also add a decent description to the PR

No major blockers from my side though, mostly nits/ additional testing required

pkg/apis/machine/v1beta1/machinehealthcheck_types.go
```diff
@@ -80,6 +80,17 @@ type MachineHealthCheckSpec struct {
...
+	// Failed Machines that are older than this value and are without a nodeRef or a providerID
+	// will be considered to have passed the time period allocated for a manual fix and will be remediated.
+	// Expects an unsigned duration string of decimal numbers each with optional
+	// fraction and a unit suffix, eg "300ms", "1.5h" or "2h45m".
+	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+	// +optional
+	// +kubebuilder:default:="48h"
+	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$"
+	// +kubebuilder:validation:Type:=string
+	FailedNodeStartupTimeout metav1.Duration `json:"failedNodeStartupTimeout,omitempty"`
```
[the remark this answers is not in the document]
mshitrit: That's an interesting point.
Definitely make sense to me as well that using ``default`` would make the ``omitempty`` redundant.
However I did a quick sweep on the code under this repo and saw that in most cases both are being used together - so I am not completely sure :thinking:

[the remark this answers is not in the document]
JoelSpeed: Having looked at the enhancement, I think we need to have some discussion about whether or not to even have a default. Can we continue that on the enhancement?

elmiko:
looks like this might need a rebase?

mshitrit:
> looks like this might need a rebase?

Hi @elmiko , any idea if you are planning to merge it soon ?
(I'd like to know whether to prioritize this)

elmiko:
> Hi @elmiko , any idea if you are planning to merge it soon ?

good question. given that we don't have a bugzilla associated with this, i'm guessing we are waiting for 4.8 master to open, if that's the case then you have a couple weeks before we could merge it. (cc @JoelSpeed )

JoelSpeed:
Yep, this will have to wait until the 4.9 master branch opens towards the end of May

JoelSpeed (commented):
I think this is pretty much ready to go apart from the default.

Would like to make sure we have support upstream for this too, let's make sure we are having that conversation

JoelSpeed:
/lgtm

> Would like to make sure we have support upstream for this too, let's make sure we are having that conversation

Would like to pursue this (at least start the conversation) before we merge this if possible. We have just under 4 weeks til feature freeze, I think we could probably get some consensus one way or another by then

mshitrit:
> /lgtm
> 
> > Would like to make sure we have support upstream for this too, let's make sure we are having that conversation
> 
> Would like to pursue this (at least start the conversation) before we merge this if possible. We have just under 4 weeks til feature freeze, I think we could probably get some consensus one way or another by then

Adding Marc, I assume he'll be involved in necessary upstream process.  
/cc @slintes

mshitrit:
@JoelSpeed , @elmiko 
Is this PR still relevant ?

JoelSpeed:
As far as I'm aware, this is still an issue that we wanted to solve. As far as I know, nothing else has solved this issue just yet.

CC @slintes @beekhof

JoelSpeed:
/approve

I think this is good and aligns with the enhancement document now. I'm keen to see if we can get agreement for this feature upstream as well. If we can hold off merging this for a little bit while we negotiate that, would be appreciated.

slintes:
Leaving a link to the upstream discussiom here, I always have to search for it 😉 
https://github.com/kubernetes-sigs/cluster-api/issues/3106#issuecomment-947466135

No activity since the last comment 2 weeks ago...

mshitrit:
Closing this PR since upstream is distinctly uninterested in this feature and the burden of maintaining the delta indefinitely was considered disproportionate to the benefits gained 
 [Jira Link](https://issues.redhat.com/browse/OCPCLOUD-800)

PR #826, opened 2021-03-11
install/0000_90_machine-api-operator_04_alertrules: Drop MachineAPIOperatorDown

wking:
The cluster-version operator is responsible for complaining if the machine-API operator's deployment is sad, so no need for the operator to handle this directly (we end up doubling up if there's an issue).  This drops the alert which was added in 9b666d6f20 (#388).

See also openshift/cloud-credential-operator#308.  Also, I just saw MachineAPIOperatorDown in the wild during the recent build01 explosion.


elmiko (approved):
this makes sense to me, thanks Trevor!
/lgtm

PR #838, opened 2021-04-05
Fetch credentials  with server name in lowercase

apoorvajagtap:
Searches for the credentialsSecretUser and credentialsSecretPassword with server name in lower case.
As per the secret's format, parameters in `data` section are expected to be in lower case, so a Workspace.server  set in another format leads to errors while scaling up the node.

Reference: [bz-1945823](https://bugzilla.redhat.com/show_bug.cgi?id=1945823)

### DIFF
--- a/pkg/controller/vsphere/machine_scope.go
+++ b/pkg/controller/vsphere/machine_scope.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


apoorvajagtap:
As per the discussion on [RHBZ #1945823](https://bugzilla.redhat.com/show_bug.cgi?id=1945823), closing this PR, to ensure that we enforce the DNS names correctly at our API level rather than mutating customer input.

PR #840, opened 2021-04-07
Bug 1918307: backported vmsizes and closing response body to 4.6 in cluster-api-provider-azure

Makdaam:
Updating to newest commit in release-4.6 branch of cluster-api-provider-azure.


Makdaam:
After running 'make vendor' it turns out that the update of go.mod is not required.
Thanks for the help @JoelSpeed

PR #846, opened 2021-04-09
Backport MHC External Remediation Feature

n1r1:
this is the same as https://github.com/openshift/machine-api-operator/pull/795 but hopefully with no conflicts.
I have no access to push commits to that PR so trying my luck as new PR since FF is coming.


elmiko (approved):
this looks fine to me, but i guess we will need to figure out if this or #795 should be the PR to take forward. also, seems like we will need to make a decision about how, or if, to bring this in given the release cycle.
/approve

elmiko:
i talked with @mshitrit and #795 is all fixed up. i am going to close this so we can keep the history on the other pr.
/close

PR #859, opened 2021-04-29
[release-4.6] Bug 1933584: Make sure phase is always set before creation

alexander-demicev:
[release-4.6] Bug 1933584: Make sure phase is always set before creation


elmiko (approved):
/lgtm

PR #863, opened 2021-05-07
Correct doc namespace err and misspellings

srcarrier:
I believe this Dev instruction in the readme : 

- Create a `openshift-machine-api-operator` namespace

aught to be this namespace instead:

- Create a `openshift-machine-api` namespace

small update


JoelSpeed (commented):
Thanks for the fixes!

/approve

PR #864, opened 2021-05-10
Fixing label from clutserapi to clusterapi

rutvik23:
Since the MachineHealthCheck object does not allow modifying default labels manually, hence I would like to fix this in it's resource yaml itself.


JoelSpeed (commented):
/approve

Thanks, good find!

lobziik:
/hold

Forgot about code freeze, sorry...

PR #875, opened 2021-06-17
Add kubevirt images

elmiko:
This change re-adds the kubevirt images and tests to the
machine-api-operator. They were removed to enable a backport to the 4.8
release branch, in which those images were not shipped. This changeset
sets things back to the status quo.

ref: #873

### DIFF
--- a/install/0000_30_machine-api-operator_01_images.configmap.yaml
+++ b/install/0000_30_machine-api-operator_01_images.configmap.yaml
@@ -19,5 +19,6 @@ data:
       "clusterAPIControllerAzure": "registry.svc.ci.openshift.org/openshift:azure-machine-controllers",
       "clusterAPIControllerGCP": "registry.svc.ci.openshift.org/openshift:gcp-machine-controllers",
       "clusterAPIControllerOvirt": "registry.svc.ci.openshift.org/openshift:ovirt-machine-controllers",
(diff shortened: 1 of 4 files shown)


elmiko:
this must go after #873
/hold

install/0000_30_machine-api-operator_01_images.configmap.yaml
```diff
@@ -19,5 +19,6 @@ data:
       "clusterAPIControllerAzure": "registry.svc.ci.openshift.org/openshift:azure-machine-controllers",
       "clusterAPIControllerGCP": "registry.svc.ci.openshift.org/openshift:gcp-machine-controllers",
       "clusterAPIControllerOvirt": "registry.svc.ci.openshift.org/openshift:ovirt-machine-controllers",
+      "clusterAPIControllerKubevirt": "registry.svc.ci.openshift.org/openshift:kubevirt-machine-controllers",
```
Fedosin, line 22: As I know `registry.svc.ci.openshift.org` is dead now. Shouldn't we use either `registry.ci.openshift.org` or `quay.io` instead?

elmiko: that's a good point, but iirc these all get replaced during the release image process. for this change i reversed the removal i had done, but your comment makes me wonder if we shouldn't update all of these?

cc @joelspeed, wdyt?

elmiko:
talk with @Fedosin in chat, we are ok to merge this but should revisit the image reference question.
/hold cancel

PR #881, opened 2021-06-28
Replace broken image references

Fedosin:
Now all image references in the operator use "registry.svc.ci.openshift.org", which has been disabled.

This commit replaces broken image references with working ones from quay.io

### DIFF
--- a/install/0000_30_machine-api-operator_01_images.configmap.yaml
+++ b/install/0000_30_machine-api-operator_01_images.configmap.yaml
@@ -10,18 +10,18 @@ metadata:
 data:
   images.json: >
     {
-      "machineAPIOperator": "registry.svc.ci.openshift.org/openshift:machine-api-operator",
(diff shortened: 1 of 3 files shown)


elmiko (commented):
shouldn't we be using the `registry.ci.openshift.org` versions of these?

i'm not sure, just wondering what effects using the quay.io refs will have.

Fedosin:
registry.ci.openshift.org is an internal image registry for storing testing images, it shouldn't be exposed publicly.
Other components always use either quay.io or docker.io: 
https://github.com/openshift/cluster-storage-operator/blob/master/manifests/image-references 
https://github.com/openshift/cluster-network-operator/blob/master/manifests/image-references

elmiko (approved):
thanks for the explanation @Fedosin !
/approve

alexander-demichev (approved):
/lgtm

PR #882, opened 2021-06-30
Bug 1977369: Prevent machine from stucking in Deleting phase on vSphere if related node object not found

lobziik:
Treat node as unreachable and proceed with vm deletion in case on vSphere if node object referenced in machine not found.


elmiko (approved):
this makes sense to me
/lgtm

PR #891, opened 2021-07-21
Bug 1975296: Remove external remediation annotation for healthy machines

slintes:
There are cases when the machine is healthy again before external remediation happened, e.g. when remediation failed, but the machine rebooted for another reason. Prevent remediation of these now healthy machines by removing the external remediation annotation.

This is part of the fix for bz 1975296, the other one is in https://github.com/openshift/cluster-api-provider-baremetal/pull/157

### DIFF
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -724,6 +724,9 @@ func (t *target) remediationStrategyExternal(r *ReconcileMachineHealthCheck) err
 
 func (t *target) removeRemediationStrategyExternal(r *ReconcileMachineHealthCheck) error {
 	// external annotation does not exist on the machine, nothing to do
+	if t.Machine.Annotations == nil {
+		return nil
+	}
 	if _, ok := t.Machine.Annotations[machineExternalAnnotationKey]; !ok {
 		return nil
 	}
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -69,6 +69,12 @@ const (
 	// EventExternalAnnotationAdded is emitted when external annotation was
 	// successfully added to a Node object
 	EventExternalAnnotationAdded string = "ExternalAnnotationAdded"
+	// EventExternalAnnotationRemovalFailed is emitted in case removing external annotation
+	// from a Machine object failed
+	EventExternalAnnotationRemovalFailed string = "ExternalAnnotationRemovalFailed"
+	// EventExternalAnnotationRemoved is emitted when external annotation was
+	// successfully removed from a Machine object
+	EventExternalAnnotationRemoved string = "ExternalAnnotationRemoved"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -706,6 +722,35 @@ func (t *target) remediationStrategyExternal(r *ReconcileMachineHealthCheck) err
 	return nil
 }
 
+func (t *target) removeRemediationStrategyExternal(r *ReconcileMachineHealthCheck) error {
+	// external annotation does not exist on the machine, nothing to do
+	if _, ok := t.Machine.Annotations[machineExternalAnnotationKey]; !ok {
```
mshitrit, line 727: (Nice to have) - Might be worth checking `t.Machine.Annotations` isn't nil as a safeguard

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -241,6 +241,10 @@ func TestReconcile(t *testing.T) {
 	machineAlreadyDeleted := maotesting.NewMachine("machineAlreadyDeleted", nodeAlreadyDeleted.Name)
 	machineAlreadyDeleted.SetDeletionTimestamp(&metav1.Time{Time: time.Now()})
 
+	// machine with external remediation annotation, but with now healthy node
+	machineHealthyWithExtRemAnnotation := maotesting.NewMachine("healthyWithExtRemAnn", nodeHealthy.Name)
+	machineHealthyWithExtRemAnnotation.Annotations[machineExternalAnnotationKey] = ""
```
mshitrit, line 246: (NTH) I realize that the value of the annotation isn't important (only it's existence), however I still think it'll be better to use in at least one of the cases none-empty string.
That is in the off chance that latter on the production code will actually regard the value.

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
slintes: will do 👍🏼

pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
slintes: we are in full control of how this annotation is used, and it will be replaced by the external remediation feature soonish anyway... so I think this is fine

pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -253,8 +259,18 @@ func (r *ReconcileMachineHealthCheck) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, err
 	}
 	errList = append(errList, r.remediate(ctx, needRemediationTargets, mhc)...)
+
+	// remove external remediation annotation for healthy machines in case they get healthy before remediation happened
+	for _, h := range currentHealthy {
```
zaneb, line 264: I don't believe there's any safe way to treat a Machine that is potentially in the middle of being rebooted as 'healthy'. We need to treat it as unhealthy and apply the MaxUnhealthy limit as in #898, otherwise there will always be a possibility of remediating too many Machines at once.

It's possible to return a _separate_ list of "reprieved" Machines and try to cancel remediation as a backup. However, per https://github.com/openshift/cluster-api-provider-baremetal/pull/157#discussion_r683473906 I'm not sure there is ever a safe way to implement this.

zaneb (commented):
Note that this must not merge before openshift/cluster-api-provider-baremetal#157, since without that change removing the annotation will simply prevent CAPBM from completing the reboot process.
/hold

slintes:
closing in favor of https://github.com/openshift/machine-api-operator/pull/898

/close

PR #878, opened 2021-06-22
install/0000_90_machine-api-operator_04_alertrules: Use '!~' for MachineWithNoRunningPhase

wking:
b13dbe9949 (#807) added `|Deleting`.  But to move from "not equal" to "regexp does not match" we need to match that with `!~`.

/assign @elmiko

PR #907, opened 2021-08-31
[release-4.7] add alert for machine with long deletion phase

vrutkovs:
Cherry-pick of #807 and #878, as CCX has noticed an uptick in MachineWithoutValidNode alerts on vsphere clusters recently


sdodson:
I've added approved label because this has merged upstream and shipped there already.

PR #914, opened 2021-09-08
[OCPCLOUD-855] add a distinct status message for NoOp mode

elmiko:
When the MAO is running on platforms it does not recognize it will be in
NoOp mode. This change adds a message to the operator status that
indicates this mode of operation, it also adds tests and documentation.


JoelSpeed (commented):
/approve

PR #920, opened 2021-09-20
Add logging for possible error during vsphere session creation

lobziik:
Log possible error, which might be returned from `SessionIsActive` check.


alexander-demichev (approved):
/approve

JoelSpeed (commented):
/lgtm

PR #924, opened 2021-10-01
podsecurity: enforce privileged for  openshift-machine-api namespace

s-urbaniak:
Starting with OpenShift 4.10 we are introducing PodSecurity admission (https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement).

Currently, all pods are marked as privileged, however, over time we want to enforce at least baseline, admirably restricted as default. In order not to break control plane workloads this allows workloads in `openshift-machine-api` namespace to run privileged pods.

See https://github.com/openshift/enhancements/pull/899 for more details (and excuse the eventual consistency of updates).

/cc @stlaz


elmiko (commented):
this makes sense to me, i'd like to get another set of eyes on it though
/approve

Fedosin (approved):
/lgtm

PR #931, opened 2021-10-11
Fix flaky AWS provider unit test

slintes:
The map used in `getDuplicatedTags` does not guarantee any order when items are read,
so create the `duplicatedTags` slice on the fly in the first loop.

Fixes flakiness of the `TestValidateAWSProviderSpec/with_double_tag_names,_lists_duplicated_tags` test.


elmiko (commented):
this makes sense to me, thanks for finding the fix!
/approve

PR #933, opened 2021-10-12
machine-api-operator: release leader lock when context is cancelled

vrutkovs:
This should speed up machine-api operator updates.

Testing 4.9 upgrades:
* [test upgrade 4.9 openshift/machine-api-operator#933 aws](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1447830580975636480)
* [test upgrade 4.9 openshift/machine-api-operator#933 gcp](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1447830601599029248)
* [test upgrade 4.9 openshift/machine-api-operator#933 azure](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-azure/1447830619546456064)

### DIFF
--- a/cmd/machine-api-operator/start.go
+++ b/cmd/machine-api-operator/start.go
@@ -73,10 +73,11 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	stopCh := make(chan struct{})
 
 	leaderelection.RunOrDie(context.TODO(), leaderelection.LeaderElectionConfig{
-		Lock:          CreateResourceLock(cb, componentNamespace, componentName),
-		LeaseDuration: util.LeaseDuration,
-		RenewDeadline: util.RenewDeadline,
-		RetryPeriod:   util.RetryPeriod,
+		Lock:            CreateResourceLock(cb, componentNamespace, componentName),
+		ReleaseOnCancel: true,
+		LeaseDuration:   util.LeaseDuration,
+		RenewDeadline:   util.RenewDeadline,
+		RetryPeriod:     util.RetryPeriod,
 		Callbacks: leaderelection.LeaderCallbacks{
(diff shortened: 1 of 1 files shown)


cmd/machine-api-operator/start.go
```diff
@@ -73,10 +73,11 @@ func runStartCmd(cmd *cobra.Command, args []string) {
 	stopCh := make(chan struct{})
 
 	leaderelection.RunOrDie(context.TODO(), leaderelection.LeaderElectionConfig{
-		Lock:          CreateResourceLock(cb, componentNamespace, componentName),
-		LeaseDuration: util.LeaseDuration,
-		RenewDeadline: util.RenewDeadline,
-		RetryPeriod:   util.RetryPeriod,
+		Lock:            CreateResourceLock(cb, componentNamespace, componentName),
+		ReleaseOnCancel: true,
```
wking, line 77: I don't expect this to do anything, because the `context.TODO()` getting passed in above will never be canceled.  For comparison, the cluster-version operator [sets up a few separate contexts and a TERM catcher][1] so we can:

1. Catch the TERM and cancel `runContext`.
2. Collect all the goroutines that had been using `runContext`, to ensure we no longer have any goroutines managing in-cluster operands.
3. Cancel `postMainContext`, which was used for informers and the leader lease.
4. `shutdownContext`, which lets callers know when they can give up on a graceful child-goroutine reap and stop blocking on it.

[1]: https://github.com/openshift/cluster-version-operator/blob/0ec39d9b2ab1feee8815d7b6b4bbe2db23daf847/pkg/start/start.go#L176-L195

PR #936, opened 2021-10-17
Remove "kubevirt-machine-controllers" image from permanently

nirarg:
It is done as part of the decision to stop support kubevirt platform in Openshift Installer


elmiko (commented):
thanks @nirarg 
/approve

PR #949, opened 2021-11-10
GCP CI runs are complaining about APIs not being enabled

joelddiaz:
disable the API service check

### DIFF
--- a/install/0000_30_machine-api-operator_00_credentials-request.yaml
+++ b/install/0000_30_machine-api-operator_00_credentials-request.yaml
@@ -116,6 +116,7 @@ spec:
   providerSpec:
     apiVersion: cloudcredential.openshift.io/v1
     kind: GCPProviderSpec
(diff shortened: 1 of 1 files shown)


elmiko (commented):
talked with @deads2k , i am adding the approval labels to this to help us solve the CI issue we are seeing.
/approve
/lgtm

deads2k:
We're merging this PR to get payloads moving again in the face of https://bugzilla.redhat.com/show_bug.cgi?id=2021731 .  We believe the  GCP API we're using is having issues (reproduced in a script), so we are going to merge this PR to unstick our payloads, but we want to ensure we get the safety check back before shipping 4.10

deads2k:
gcp passed, this is neeed to unblock our installs.  merging.

wking:
/retitle Bug 2021731: GCP credentials reporting networksecurity.googleapis.com API disabled

PR #955, opened 2021-11-12
Bug 2022840: GCP CI runs are complaining about APIs not being enabled

elmiko:
disable the API service check


wking:
unrelated platform, but aws-operator is required, so retest to see if we avoid the `...100 maximum total nodes...` autoscaler failure this time:

/test e2e-aws-operator

PR #958, opened 2021-11-23
Move to setup-envtest

alexander-demichev:
This PR removes kubebuilder envtest setup script and introduces the usage of setup-envtest binary.


lobziik:
did you try `go install`?

https://pkg.go.dev/sigs.k8s.io/controller-runtime/tools/setup-envtest#section-readme

elmiko:
i tried testing this locally by running `make unit`, but i got a bunch of errors about being unable to start the control plane. is the expectation that i should be able to run this?

PR #961, opened 2021-11-23
Bug 1918005: Use known vSphere cluster to uniquely identify networks.

bostrt:
vSphere Network names are not necessarily unique, especially in NSX-T environments which our CI uses. The vSphere Cluster is known from within the reconciler and can be used to isolate the unique network needed to clone machines. 

Tracking in bz1918005 and [SPLAT-345](https://issues.redhat.com/browse/SPLAT-345)

### DIFF
--- a/pkg/controller/vsphere/reconciler.go
+++ b/pkg/controller/vsphere/reconciler.go
@@ -638,7 +638,7 @@ func clone(s *machineScope) (string, error) {
 	}
 
 	klog.V(3).Infof("Getting network devices")
-	networkDevices, err := getNetworkDevices(s, devices)
+	networkDevices, err := getNetworkDevices(s, resourcepool, devices)
 	if err != nil {
 		return "", fmt.Errorf("error getting network specs: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


bostrt:
Just found a problem installing across multiple vSphere clusters. 

/hold

pkg/controller/vsphere/reconciler.go
```diff
@@ -719,20 +720,47 @@ func getNetworkDevices(s *machineScope, devices object.VirtualDeviceList) ([]typ
...
+				backing, err = networkObject.EthernetCardBackingInfo(s.Context)
+				if err != nil {
+					return nil, fmt.Errorf("unable to create new ethernet card backing info for network %q: %w", netSpec.NetworkName, err)
+				}
+				break
+			}
+		}
+
+		if backing == nil {
+			return nil, fmt.Errorf("unable to get network for %q: %w", netSpec.NetworkName, err)
```
bostrt, line 763: `err` may be nil here resulting in an ugly message like this:
```
error getting network specs: unable to get network for "ocp-ci-seg-13": %!w(<nil>)
```

elmiko (commented):
this generally looks good to me, just a quick question @bostrt , is it possible to test some of the failure states for the new change in the unit tests?

i totally understand if this is beyond the scope the unit test capabilities, i just want to make sure i understand completely.

elmiko:
> I've added one additional test for when the wrong network is configured. Unfortunately, testing [the code inside the loop](https://github.com/openshift/machine-api-operator/pull/961/files#diff-21a982202fdcdf4beffc4d00659798b54717940bf07f685006dbcb99d9dbf0ecR739-R760) is difficult without adding a mock framework around govmomi calls.

i figured it might be difficult to test some parts of it, thanks for adding that additional test though.
/lgtm

PR #985, opened 2022-02-01
Updating infrastructure CR and upgrading latest version of api

Connect2naga:
Intake of latest openshift api : to support userTags in infrastructure spec


Connect2naga:
Closing

PR #991, opened 2022-02-10
Fix broken link to CVO run levels

bshephar:
The current hyperlink for run level points to the CVO
repo version of the document, which no longer exists. This change
is adding the new, correct link to the operator.md file
in the enhancements repo.

Current link:
```
❯ curl -I https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/operators.md#how-do-i-get-added-as-a-special-run-level
HTTP/2 404
```

New Link in this change:
```
❯ curl -I https://github.com/openshift/enhancements/blob/master/dev-guide/cluster-version-operator/dev/operators.md#how-do-i-get-added-as-a-special-run-level
HTTP/2 200
```

### DIFF
--- a/docs/user/machine-api-operator-overview.md
+++ b/docs/user/machine-api-operator-overview.md
@@ -18,7 +18,7 @@ Namely:
 
 MAO installation/upgrade procedure relies on CVO. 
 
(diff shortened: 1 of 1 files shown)


elmiko (commented):
/lgtm

PR #999, opened 2022-03-11
Remove 30s wait before creating a machine

mdbooth:
In the machine create workflow we were setting the phase to Provisioning
and then requeuing with the default 30s delay before starting the actual
machine creation. This delay is unnecessary in this case. We let the
default rate limiter handle it instead, which will requeue after 5ms.


mdbooth:
How many of these e2e tests do we want to pass? Looks like a bit of a bloodbath, but this change isn't really closely tied to the behaviour of any cloud provider.

PR #1027, opened 2022-06-15
Bug 2097153: change ListTags call to ListTagsForCategory

briantward:
Improves performance in vCenter clusters with thousands of tags and heavy API loads. There should be no reason we need to query for all tags unrelated to a particular install.

### DIFF
--- a/pkg/controller/vsphere/actuator_test.go
+++ b/pkg/controller/vsphere/actuator_test.go
@@ -166,7 +166,7 @@ func TestMachineEvents(t *testing.T) {
 		g.Expect(k8sClient.Delete(context.Background(), userDataSecret)).To(Succeed())
 	}()
 
-	createTagAndCategory(session, "CLUSTERID", "CLUSTERID")
+	createTagAndCategory(session, "openshift-CLUSTERID", "CLUSTERID")
 
 	ctx := context.Background()
 
--- a/pkg/controller/vsphere/actuator_test.go
+++ b/pkg/controller/vsphere/actuator_test.go
@@ -166,7 +166,7 @@ func TestMachineEvents(t *testing.T) {
 		g.Expect(k8sClient.Delete(context.Background(), userDataSecret)).To(Succeed())
 	}()
 
-	createTagAndCategory(session, "openshift-CLUSTERID", "CLUSTERID")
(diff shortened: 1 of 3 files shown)


briantward:
Based the category format on installer code:

https://github.com/openshift/installer/blob/master/data/data/vsphere/pre-bootstrap/main.tf#L57

elmiko (commented):
this makes sense to me, thanks for proposing the fix!
/lgtm

briantward:
related BZ I opened on this... https://bugzilla.redhat.com/show_bug.cgi?id=2097153

elmiko:
i just noticed this is failing the unit tests, i think we need some investigation on that.
/cancel lgtm

briantward:
I'll check it out.

elmiko:
@briantward thanks for the update. we talked about this pr and the related bug in our team bug triaging. we think this patch looks good, but we want to confirm the source of the `openshift-*` tags just to make sure we don't introduce a regression. do you have any more source info this?

elmiko:
@JoelSpeed @lobziik ptal, curious for your thoughts

JoelSpeed:
Given this additional context about the tag category naming being the tag name with the `openshift-` prefix, perhaps we can add a function something like this that captures the context in the code? Might be better to use permalinks rather than `master` links though in case things move later

```
// tagToCategoryName converts the tag name to the category name based upon the format set up by the installer.
// Note this is only valid in IPI clusters as typically a UPI cluster won't have the cluster ID tag, in which case the
// controller skips tag creation.
// Ref: https://github.com/openshift/installer/blob/master/data/data/vsphere/pre-bootstrap/main.tf#L57
// Ref: https://github.com/openshift/installer/blob/master/pkg/destroy/vsphere/vsphere.go#L231
func tagToCategoryName(tagName string) string {
  return fmt.Sprintf("openshift-%s", tagName)
}
```

briantward:
@JoelSpeed good recommendation.  ptal at added function, let me know if it's ok there or you want it elsewhere.

JoelSpeed (commented):
/lgtm

elmiko (commented):
happy to approve this once we get the final fixups in place

PR #1038, opened 2022-07-12
Bug 2104642: fix the Machine validation webhook for nutanix providerSpec

yanhua121:
Bug 2105382: fix the Machine validation webhook for nutanix providerSpec
Add validation for fields: CredentialsSecret and UserDataSecret


elmiko:
given the timing, we will have to target this bug for 4.12 and then backport it to 4.11. it shouldn't be a problem though.

elmiko:
if this is in addition to @thunderboltsid patch, then we need to make this the bug https://bugzilla.redhat.com/show_bug.cgi?id=2104642

elmiko:
doing a little more research, i think we just need to put this on the same bug as the previous change.

/retitle Bug 2104642: fix the Machine validation webhook for nutanix providerSpec

miyadav:
Details of test added to : https://bugzilla.redhat.com/show_bug.cgi?id=2104642#c3

/lgtm 
/label qe-approved

elmiko:
looks like we need https://github.com/openshift/machine-api-operator/pull/1036 to merge before the cherry-pick would work.

elmiko:
i have created #1039 to address the backport of both PRs

PR #1042, opened 2022-07-20
Bug 2109258: add more support for old delete annotation

elmiko:
this change adds detection to the newest and random delete priority
functions to ensure that we have coverage for the older annotation until
it is removed. also updates the tests to cover a situation where the old
annotation was not detected properly and adds more tests for the new
interactions.


wking:
Grepping around to confirm that this doesn't need a backport, the pivot away from the old annotation began in 10a5e776 (#1024), and that merged after 4.11 forked off, so we don't need a 4.11 backport here.

PR #1044, opened 2022-07-25
Bug 2110505: revert: Bug 2101880: operator NS manifest: Set empty openshift.io/run-level

bparees:
revert of https://github.com/openshift/machine-api-operator/pull/1031


LalatenduMohanty (approved):
/lgtm

bparees:
/hold

not merging this until at a minimum we have some verification from @sdodson 
we also need the master version of the PR

PR #1049, opened 2022-07-28
update windows helper functions

elmiko:
this change updates the tag identifiers that are used by the WMCO when
creating the secret for the windows boot script. we need to have a
slightly tighter coupling with the WMCO until such time as they drop the
tags from their implementation. once the WMCO removes all the tags, we
should evualte the usefulness of the helper functions and remove or
modify them as necessary. for context, only some of the cloud provider
will need these tags, currently AWS and Azure.

### DIFF
--- a/pkg/util/windows/helpers.go
+++ b/pkg/util/windows/helpers.go
@@ -22,8 +22,11 @@ import (
 )
 
 const (
+	// these powershell tag constants are based on the secret creation in WMCO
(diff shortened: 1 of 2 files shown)


elmiko:
this will be needed by https://github.com/openshift/machine-api-provider-gcp/pull/14

JoelSpeed (commented):
/lgtm
/approve

PR #1052, opened 2022-08-08
Remove openstack-machine-controllers from image-references

mdbooth:
In support of https://github.com/openshift/ocp-build-data/pull/1817


mdbooth:
@JoelSpeed Nothing failed here that I care about, and nothing failed by not building anything. I think this is good?

mdbooth:
We forgot about this and I submitted it again as #1111, which has now merged.

PR #1069, opened 2022-09-16
Fix lint issues reported by golangci-lint

Fedosin:
This PR adds golangci-lint support and enables additional linters for the operator.


pkg/controller/machineset/controller_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	machinev1 "github.com/openshift/api/machine/v1beta1"
+	"github.com/stretchr/testify/assert"
```
JoelSpeed, line 27: Can we use gomega here instead? It's already being used within the tests in this file

pkg/controller/vsphere/actuator_test.go
```diff
@@ -24,8 +24,15 @@ import (
 
 func init() {
 	// Add types to scheme
-	machinev1.AddToScheme(scheme.Scheme)
-	configv1.AddToScheme(scheme.Scheme)
+	err := machinev1.AddToScheme(scheme.Scheme)
+	if err != nil {
```
JoelSpeed, line 28: Can you inline this
```suggestion
	if err := machinev1.AddToScheme(scheme.Scheme); err != nil {
```

pkg/controller/vsphere/actuator_test.go
```diff
@@ -24,8 +24,15 @@ import (
...
+		panic(err)
+	}
+
+	err = configv1.AddToScheme(scheme.Scheme)
+	if err != nil {
```
JoelSpeed, line 33: There's a linter that checks for inlining these, lets enable it
```suggestion
	
	if err := configv1.AddToScheme(scheme.Scheme); err != nil {
```

pkg/controller/vsphere/machineset/controller_suite_test.go
```diff
@@ -55,9 +55,9 @@ var _ = BeforeSuite(func() {
 	testEnv = &envtest.Environment{
 		CRDDirectoryPaths: []string{filepath.Join("..", "..", "..", "..", "install")},
 	}
-	machinev1.AddToScheme(scheme.Scheme)
+	err := machinev1.AddToScheme(scheme.Scheme)
+	Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 59: Could merge these lines

tools.go
```diff
@@ -1,6 +1,5 @@
 package tools
 
 import (
-	// Import the v1 API so that we can sync the CRD into the install folder
-	_ "github.com/openshift/api/machine/v1"
```
JoelSpeed, line 5: Do we not need this anymore?

JoelSpeed (commented):
Should we add a golangci config file like we did in the CPMS repo?

tools.go
Fedosin: yes, it's imported directly in the code in several places, like [here](https://github.com/openshift/machine-api-operator/blob/7dd1f5294b1ff9730f51ebad442035938240d94d/pkg/webhooks/machine_webhook.go#L14)

go.mod
```diff
@@ -6,20 +6,19 @@ require (
...
 	github.com/go-logr/logr v1.2.3
 	github.com/google/gofuzz v1.2.0 // indirect
-	github.com/google/uuid v1.2.0
+	github.com/google/uuid v1.3.0
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.20.1
-	github.com/openshift/api v0.0.0-20220921125526-1866ef90edbf
-	github.com/openshift/client-go v0.0.0-20220915152853-9dfefb19db2e
-	github.com/openshift/library-go v0.0.0-20220920133651-093893cf326b
+	github.com/openshift/api v0.0.0-20220628012702-9368b92de0ab
```
JoelSpeed, line 12: Why is this going back to a version from june? When it is up to date at present?

JoelSpeed:
The goling job is failing here, is that to be expected given the changes? Looks like it's picking up some files that are not gofmt'd

lobziik (commented):
Mostly lgtm, just couple things

pkg/controller/machineset/util.go
```diff
@@ -60,7 +60,7 @@ func GetReadyCondition(status *corev1.NodeStatus) *corev1.NodeCondition {
 
 // IsNodeReady returns true if a node is ready; false otherwise.
 func IsNodeReady(node *corev1.Node) bool {
-	if node == nil || &node.Status == nil {
```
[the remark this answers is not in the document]
lobziik: Ok, I checked type definition, `Status` can not be nil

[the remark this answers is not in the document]
Fedosin: yes, it's correct... `Status` is not a pointer, so it can't be `nil`

pkg/controller/vsphere/session/test_ids_caching_client_test.go
```diff
@@ -105,13 +105,13 @@ func TestGetTag(t *testing.T) {
 
 			cleanupTagsAndCategories(ctx, m, g)
 
-			tag, err = m.GetTag(ctx, "fizz")
+			_, err = m.GetTag(ctx, "fizz")
```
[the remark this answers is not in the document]
Fedosin: The issue was that we assign a value to the variable and then don't use it. In this particular example, we expect the `err` to occur, so we don't care about `tag`.

go.mod
```diff
@@ -35,79 +34,224 @@ require (
 )
 
 require (
-	cloud.google.com/go v0.97.0 // indirect
+	github.com/golangci/golangci-lint v1.49.0
```
lobziik, line 37: Is that intentional to have this in separate block?

pkg/controller/vsphere/reconciler_test.go
```diff
@@ -2220,8 +2240,13 @@ func TestReconcileMachineWithCloudState(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	createTagAndCategory(session, zoneKey, testZone)
-	createTagAndCategory(session, regionKey, testRegion)
+	if err := createTagAndCategory(session, zoneKey, testZone); err != nil {
+		t.Fatalf("cannot create tag and category: %v", err)
+	}
+
+	if err := createTagAndCategory(session, regionKey, testRegion); err != nil {
```
lobziik, line 2247: I'm thinking about introducing gomega here.

P.S. Just thought, lets address this in another pr

lobziik (commented):
LGTM overall. Just one question about go.mod structuring

go.mod
Fedosin: The file has been generated by `go mod tidy`. Probably there is a limit on how many items you can have in one section.

elmiko (commented):
lgtm

i'll leave it for @lobziik  to add the label once his question is answered

lobziik (approved):
I'm ok with the answers :) Dont see the reason to block it

/lgtm

PR #1090, opened 2022-11-28
Use StringDeref from k8s.io/utils instead of custrom implementation

Fedosin:
This is a cleanup commit that replaces custrom implementations of StringDeref function with one from k8s.io/utils


lobziik (approved):

elmiko (commented):
/lgtm

PR #1091, opened 2022-11-29
Increase timeout for machineset tests

Fedosin:
Currently, when running in CI, machineset tests sometimes fail with timeout error. 
```txt
• Failure in Spec Teardown (AfterEach) [12.059 seconds]
MachineSet Reconciler
/go/src/github.com/openshift/machine-api-operator/pkg/controller/machineset/machineset_controller_test.go:31
  Should reconcile a MachineSet [AfterEach]
  /go/src/github.com/openshift/machine-api-operator/pkg/controller/machineset/machineset_controller_test.go:77
  Timed out after 10.000s.
```

To prevent this we increase the timeout to 20 seconds.

### DIFF
--- a/pkg/controller/machineset/machineset_controller_test.go
+++ b/pkg/controller/machineset/machineset_controller_test.go
@@ -18,6 +18,7 @@ package machineset
 
 import (
 	"context"
+	"time"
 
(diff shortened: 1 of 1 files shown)


lobziik (approved):
/lgtm

PR #1094, opened 2022-12-08
Add permissions for regionBackendServices

odvarkadaniel:
This PR adds permission for `regionBackendServices`, which are needed by this bug fix for GCP.
The reference PR can be seen here: https://github.com/openshift/machine-api-provider-gcp/pull/22
There we used get, create and update method for the `regionBackendServices`.


odvarkadaniel:
I wasn't in the office at all today, will take a look tomorrow. Nevertheless, this permission should be valid as per: https://cloud.google.com/compute/docs/reference/rest/v1/regionBackendServices/insert

PR #1097, opened 2022-12-09
Revert custom role

patrickdillon:
Revert #1094 & #1087 due to hitting iam custom role limit in gcp ci.


sdodson:
Changes here are local to GCP config and none of the failing jobs were GCP, merging to unbreak CI.

PR #1106, opened 2023-01-08
Add warning for unknown fields when unmarshaling the provider spec

eranco74:
Prior to this change unknown fields in the porvider spec were ignored silently.
With this change, if the provider spec contains unknown fields it will output a warning e.g.
```
providerSpec.value: Unsupported value: \"randomField-1\": Unknown field (randomField-1) will be ignored
```

### DIFF
--- a/pkg/webhooks/machine_webhook_test.go
+++ b/pkg/webhooks/machine_webhook_test.go
@@ -1932,6 +1932,7 @@ func TestValidateAWSProviderSpec(t *testing.T) {
 	testCases := []struct {
 		testCase         string
 		modifySpec       func(*machinev1beta1.AWSMachineProviderConfig)
+		overrideRawBytes []byte
 		expectedError    string
 		expectedOk       bool
 		expectedWarnings []string
@@ -2159,6 +2160,12 @@ func TestValidateAWSProviderSpec(t *testing.T) {
 			},
 			expectedOk: true,
 		},
+		{
+			testCase:         "with unknown fields in the providerSpec",
+			overrideRawBytes: []byte(`{"kind":"AWSMachineProviderConfig","apiVersion":"machine.openshift.io/v1beta1","metadata":{"creationTimestamp":null},"ami":{"id":"ami"},"instanceType":"m5.large","iamInstanceProfile":{"id":"profileID"},"userDataSecret":{"name":"secret"},"credentialsSecret":{"name":"secret"},"deviceIndex":0,"securityGroups":[{"id":"sg"}],"subnet":{"id":"subnet"},"placement":{"region":"region"},"metadataServiceOptions":{},"randomField-1": "something"}`),
+			expectedOk:       true,
+			expectedWarnings: []string{"providerSpec.value: Unsupported value: \"randomField-1\": Unknown field (randomField-1) will be ignored"},
+		},
 	}
 
 	secret := &corev1.Secret{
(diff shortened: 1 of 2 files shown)


pkg/webhooks/machine_webhook_test.go
```diff
@@ -3209,6 +3210,13 @@ func TestValidateGCPProviderSpec(t *testing.T) {
 			},
 			expectedOk: true,
 		},
+		{
+			testCase:         "with unknown fields in the providerSpec",
+			overrideRawBytes: []byte(`{"kind":"GCPMachineProviderSpec","apiVersion":"gcpprovider.openshift.io/v1beta1","metadata":{"creationTimestamp":null},"userDataSecret":{"name":"name"},"credentialsSecret":{"name":"name"},"canIPForward":false,"deletionProtection":false,"disks":[{"autoDelete":false,"boot":false,"sizeGb":16,"type":"","image":"","labels":null}],"networkInterfaces":[{"network":"network","subnetwork":"subnetwork"}],"serviceAccounts":[{"email":"email","scopes":["scope"]}],"machineType":"machineType","region":"region","zone":"region-zone","projectID":"projectID","gpus":[{"count":0,"type":"type"}],"onHostMaintenance":"Terminate","randomField-1": "something","randomField-2": "something"}`),
+			expectedOk:       true,
+			expectedError:    "",
+			expectedWarnings: []string{"providerSpec.value: Unsupported value: \"randomField-1\": Unknown field (randomField-1) will be ignored"},
```
damdo, line 3218: So in the warning we will return the first unknown field, correct? If there are multiple of those we won't alert on the others, is this by design or is it because `yaml.DisallowUnknownFields` doesn't return them all?

pkg/webhooks/machine_webhook_test.go
```diff
@@ -3209,6 +3210,13 @@ func TestValidateGCPProviderSpec(t *testing.T) {
 			},
 			expectedOk: true,
 		},
+		{
+			testCase:         "with unknown fields in the providerSpec",
+			overrideRawBytes: []byte(`{"kind":"GCPMachineProviderSpec","apiVersion":"gcpprovider.openshift.io/v1beta1","metadata":{"creationTimestamp":null},"userDataSecret":{"name":"name"},"credentialsSecret":{"name":"name"},"canIPForward":false,"deletionProtection":false,"disks":[{"autoDelete":false,"boot":false,"sizeGb":16,"type":"","image":"","labels":null}],"networkInterfaces":[{"network":"network","subnetwork":"subnetwork"}],"serviceAccounts":[{"email":"email","scopes":["scope"]}],"machineType":"machineType","region":"region","zone":"region-zone","projectID":"projectID","gpus":[{"count":0,"type":"type"}],"onHostMaintenance":"Terminate","randomField-1": "something","randomField-2": "something"}`),
```
damdo, line 3215: It would be great if you could add a test, say for aws and azure, just to cover the main providerSpec providers.

damdo (commented):
Thanks a lot! This is a really nice addition.
Left two minor comments.

pkg/webhooks/machine_webhook_test.go
eranco74: The latter, I think returning all the fields is better but the error we get just return the first unknown field
see [here](https://go.dev/play/p/rl7VAMwBRho)

damdo: Gotcha. Thanks for letting me know.

damdo (commented):
/lgtm

PR #1110, opened 2023-01-23
MGMT-13425: Don't fail when workers not needed

flaper87:
Don't fail if expected workers is less than 2 since masters are expected to be schedulable in this architecture, which means there will be enough replicas of the operator scheduled.

Signed-off-by: Flavio Percoco <flavio@redhat.com>

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -317,10 +317,11 @@ func (optr *Operator) checkDaemonSetRolloutStatus(resource *appsv1.DaemonSet) (r
 	return reconcile.Result{}, nil
 }
 
-// checkMinimumWorkerMachines looks at the worker Machines in the cluster and checks if they are running.
-// If fewer than 2 worker Machines are Running, it will return an error.
-// This is used during initialization of the cluster to prevent the operator from being Available
-// until the minimum required number of worker Machines have started working correctly.
+// checkMinimumWorkerMachines looks at the worker Machines in the cluster and checks if they are
+// running. If fewer than 2 worker Machines are Running and the expected number is higher than 1, it
+// will return an error. This is used during initialization of the cluster to prevent the operator
+// from being Available until the minimum required number of worker Machines have started working
+// correctly.
 func (optr *Operator) checkMinimumWorkerMachines() error {
 	machineSets, err := optr.machineClient.MachineV1beta1().MachineSets(optr.namespace).List(context.Background(), metav1.ListOptions{})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


zaneb:
So IIUC the goal here is for installation to succeed on a compact cluster even if there should be workers but none of the workers came up. Since the masters are schedulable the user presumably won't be faced with cryptic complaints about the Ingress operator not coming up that we would want to short-circuit.

We came up with the minimum-of-2-workers criterion to split the difference between doing nothing and blocking on _any_ failures, because we recognised that some users would want all workers to come up before considering the installation successful, while others would not want to block on a small percentage of failures. But at the time the 1-worker case was not considered. You could make an argument either way for which bucket it should fall into (on one hand, Ingress will still come up; on the other... 100% of workers failed). The ticket doesn't appear to do that though - it's just that somebody expected it to work one way and was surprise that it's the other? You could easily imagine a future ticket with the exact opposite complaint.

flaper87:
> We came up with the minimum-of-2-workers criterion to split the difference between doing nothing and blocking on any failures, because we recognised that some users would want all workers to come up before considering the installation successful, while others would not want to block on a small percentage of failures.

Right, but this was not with compact in mind but multinode. In fact, for compact + 1 worker it used to work as this PR proposes. The installation would succeed even if there were no workers available. This PR is trying to bring back the way it used to work. We could refine this PR and make it more explicit by adding a check to verify whether the masters are actually schedulable rather than just basing everything on the counter.

zaneb:
I wouldn't want to see this applied to e.g. IPI, because in IPI the workers are provisioned by a completely separate method to the control plane, so if we can't provision a worker at all that suggests a problem that should cause installation to fail. In practice that doesn't matter because we don't (yet?) support the 3+1 topology in IPI. So I'm not taking a side, but I would like us to consider what we're trying to achieve, other than make it work how it used to (prior to 4.12 it used to never block no matter how many workers failed, and changing that was intentional).

flaper87:
> I wouldn't want to see this applied to e.g. IPI, because in IPI the workers are provisioned by a completely separate method to the control plane, so if we can't provision a worker at all that suggests a problem that should cause installation to fail. In practice that doesn't matter because we don't (yet?) support the 3+1 topology in IPI. So I'm not taking a side, but I would like us to consider what we're trying to achieve, other than make it work how it used to (prior to 4.12 it used to never block no matter how many workers failed, and changing that was intentional).

Agreed! To be clear, my goal is not to "just make it work". I am actually making a point in favor of not blocking the installation in the compact+1 case. My main motivation being:

- There are 3 masters that are schedulable
- I don't see the difference between not blocking the installation in the case of 6 nodes deployment with 1 failing worker vs a 3+1 with a failing worker. In both cases the cluster has enough resources to complete the installation and allow for post-deployment debugging of the worker.

@zaneb maybe I am missing something here (or on the IPI side), help me out :)

tsorya:
@zaneb what blocks us from pushing this pr?

zaneb:
I have no strong objections to this. In assisted installs specifically, the failure of workers is unlikely to be due to the Machine API. In practice I think only assisted (and UPI?) supports 3+1 cluster installation.
I do think we could end up revisiting this in the future if we support other ways of installing 3+1 clusters, because in other installation methods a 100% failure rate of workers could be indicative of a systemic problem with the Machine API.

tsorya:
@zaneb So lets push it

zaneb:
@tsorya I have no approval rights in this repo.

PR #1117, opened 2023-02-07
Add Machine webhook to validate the new fields of NutanixMachineProviderConfig

yanhua121:
1. Revendor the openshift/api for the newly added NutanixMachineProviderConfig fields: https://github.com/openshift/api/pull/1390
2. Add Machine webhook to validate the new fields of NutanixMachineProviderConfig


pkg/webhooks/machine_webhook.go
```diff
@@ -1507,6 +1533,22 @@ func validateNutanixResourceIdentifier(resource string, identifier machinev1.Nut
...
 
+func validateNutanixBootType(bootType machinev1.NutanixBootType) error {
+	parentPath := field.NewPath("providerSpec")
+	// verify the bootType configurations
+	// Type bootType field is optional, and valid values include: "", Legacy, UEFI, SecureBoot
+	switch bootType {
+	case "", machinev1.NutanixLegacyBoot, machinev1.NutanixUEFIBoot, machinev1.NutanixSecureBoot:
+		// valid bootType
+	default:
+		errMsg := fmt.Sprintf("valid bootType: \"\", %q, %q, %q.",
```
elmiko, line 1544: this might be slightly confusing for a user, could we change this to say something like "valid bootType values are: ...`

elmiko (commented):
code looks good to me, just a comment about one of the error messages

pkg/webhooks/machine_webhook.go
yanhua121: Changed.

elmiko (commented):
/lgtm

@JoelSpeed ptal

PR #1121, opened 2023-02-27
OCPBUGS-6063: Add missed permission for a pod deletion on vsphere platform

lobziik:
Follow up for https://github.com/openshift/machine-api-operator/pull/1118 with missing permission.

Machine deletion stucks due to vpshere machine controller trying to delete unevicted pods after machine was powered off and drained.


elmiko (commented):
/lgtm

PR #1126, opened 2023-03-08
OCPBUGS-7249: Fix zone tag value reconciliation for vSphere machines

lobziik:
VCenter does not return 'Cluster' object as a VM ancestor, use VM's host system instead.

Also, this commit introduces using of CachingTagsManager for region/zone tags info obtaining, instead of constructing a non-caching one.


elmiko (commented):
this generally makes sense to me, it didn't look like the failing vsphere test was related to this, but i'm not 100% sure.

/lgtm

PR #1128, opened 2023-03-15
OCPBUGS-10591: Fix empty component version

vfreex:
Currently `git describe` is used to determine the component during OCP build. Since git tags may not be always applied, use environment variables instead.


vfreex:
More details: https://redhat-internal.slack.com/archives/CB95J6R4N/p1678862484292639?thread_ts=1678815070.084429&cid=CB95J6R4N

elmiko (commented):
thanks @vfreex !
/approve

PR #1145, opened 2023-05-23
OCPBUGS-13956: Bump controller-runtime to v0.15.0

damdo:
Bump controller-runtime to v0.15.0 as currently we are using an alpha version.


damdo:
Thanks for the pointers.
It turned out it was a change in behavior of the controller-runtime's fake client:

From: https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0
```
Update fake client deletionTimestamp behavior (#2316)
The fake client will panic if initialized with an object that has a DeletionTimestamp and no finalizer
The fake client will silently ignore a DeletionTimestamp on an object in a Create request, matching the behavior of the kube-apiserver
```
Pushed an update and now it is fixed.

PR #1153, opened 2023-07-13
MULTIARCH-3693: Add policies to operate on VPC loadbalacers

Karthik-K-N:
As a part of adding CPMSO support for Power VS. We need to add required policies to enable controller to operator on VPC services.


mkumatag (approved):
/lgtm

PR #1166, opened 2023-09-12
SPLAT-1172: Enhance machine status relating to IPAddressClaimed

vr4manta:
The following enhancments are being made for static IP support of vSphere IPI
- Improve the provider status field of machines relating to static IP claiming
- Fix orphaned IPAddressClaims

### DIFF
--- a/vendor/github.com/openshift/api/config/v1/types_feature.go
+++ b/vendor/github.com/openshift/api/config/v1/types_feature.go
@@ -163,7 +163,7 @@ var FeatureSets = map[FeatureSet]*FeatureGateEnabledDisabled{
 		Disabled: []FeatureGateDescription{},
 	},
 	TechPreviewNoUpgrade: newDefaultFeatures().
-		with(validatingAdmissionPolicy).
+		without(validatingAdmissionPolicy).
 		with(externalCloudProvider).
 		with(externalCloudProviderGCP).
 		with(csiDriverSharedResource).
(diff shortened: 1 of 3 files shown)


pkg/controller/vsphere/reconciler.go
```diff
@@ -97,13 +97,15 @@ func (r *Reconciler) create() error {
 		}
 		condition := metav1.Condition{
 			Type:    string(machinev1.IPAddressClaimedCondition),
-			Reason:  machinev1.WaitingForIPAddressReason,
+			Reason:  machinev1.IPAddressClaimedReason,
 			Message: "All IP address claims are bound",
-			Status:  metav1.ConditionFalse,
+			Status:  metav1.ConditionTrue,
```
elmiko, line 102: i'm a little confused here, it seems like we have inverted the status but we have not changed the logic around how we end up here. could you explain a little bit about this?

elmiko (commented):
this generally makes sense to me, but i have a question

pkg/controller/vsphere/reconciler.go
vr4manta: Looking at the if block after this, if "outstandingClaims > 0", that means the IP Address Claims (IPAddressClaimed) should have been false.  This was a bug.  So the default block has true since message is "All IP address claims are bound" and type is IPAddressesClaimed.  I also fixed the if block below to have a meaningful reason since it never actually applied one so it would always say "All IP address claims are bound" even though that was untrue. It now says "WaitingForIPAddress".

vr4manta: Here is example of the outputs now:

Default status:
`    conditions:
    - lastTransitionTime: "2023-09-06T13:52:51Z"
      message: All IP address claims are bound
      reason: IPAddressesClaimed
      status: "True"
      type: IPAddressClaimed`

Status from outstanding > 0:
`    conditions:
    - lastTransitionTime: "2023-09-08T15:26:23Z"
      message: Waiting on 1 IP address claims to be bound
      reason: WaitingForIPAddress
      status: "False"
      type: IPAddressClaimed`

PR #1188, opened 2023-12-11
OCPBUGS-24277: Update reference URL

nrb:
Update URL after docs refactor.

Manual cherry-pick to avoid a duplicated Jira issue.


elmiko:
seems like it thinks there is a race in the unit tests, but i can't imagine how this change would cause that

/lgtm
/label backport-risk-assessed

/test unit

happy to approve just wanted to see if the unit test error would go away

nrb:
/retest

I suspect the vsphere e2es are timing out, not really failing on this change.

PR #1213, opened 2024-01-31
OCPBUGS-26119: add tools.go to snyk configuration file excludes

elmiko:
This file is only used for bringing in test dependencies it is not used in the final build release artifacts.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -5,3 +5,4 @@ exclude:
   global:
     - "vendor/**"
     - "**/vendor/**"
+    - tools.go


racheljpg:
Hello! Mike and I have been looking at running snyk tests manually - this seems like a reasonable change to me, but unfortunately, after testing this change manually, snyk seems to still be trying to scan this file. Not sure if there's something we are missing in the formatting?

elmiko:
we found the underlying issue, basically snyk is running the go tooling to determine the dependencies and this `tools.go` file is throwing it off. since we don't use this file in our release builds, we have added build tags to the tools file so that it will not be included in the go commands that snyk is running. this happens before snyk excludes files which is why we aren't seeing snyk exclude it.

PR #1219, opened 2024-02-27
OCPBUGS-29956: Azure MAO CredentialsRequest Contains Unnecessary virtualMachines/extensions Permissions

bennerv:
ARO team believes these permissions are unnecessary for MAO.  Testing permission removal with e2e.  

[OCPBUGS-29956](https://issues.redhat.com/browse/OCPBUGS-29956) is tracking this.

### DIFF
--- a/install/0000_30_machine-api-operator_00_credentials-request.yaml
+++ b/install/0000_30_machine-api-operator_00_credentials-request.yaml
@@ -91,9 +91,6 @@ spec:
     - Microsoft.Compute/galleries/images/versions/read
     - Microsoft.Compute/skus/read
     - Microsoft.Compute/virtualMachines/delete
-    - Microsoft.Compute/virtualMachines/extensions/delete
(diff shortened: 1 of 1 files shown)


RadekManak:
You are correct that these permissions are not in use. I believe the read permission is only reachable from an old bootstrap method that was never suported in openshift. I have opened https://github.com/openshift/machine-api-provider-azure/pull/102 to remove that code. If that merges we can remove the read permission as well.

RadekManak:
My PR merged. Can you please update this to remove the read permission.

bennerv:
> Should we be backporting these two PRs?
@JoelSpeed I'd like to if possible.  What's the process, do we need labels on the jira?  Otherwise I can open the /cherrypick PRs.

bennerv:
Cherry-pick after https://github.com/openshift/machine-api-provider-azure/pull/103 is merged in.

PR #1226, opened 2024-04-11
OCPCLOUD-2501: Update to set GpuType annotation to remove hardcoded value

racheljpg:
Hello! This is a PR to remove the hardcoded value for the GpuType annotation, so that we can use the one passed in from the CAO instead. Thanks!

### DIFF
--- a/pkg/util/machineset/util.go
+++ b/pkg/util/machineset/util.go
@@ -77,9 +77,8 @@ func SetGpuCountAnnotation(annotations map[string]string, value string) map[stri
 
 // SetGpuTypeAnnotation sets a value for gpu type in the annotations of a MachineSet.
 // Currently, we only support nvidia as a gpu type.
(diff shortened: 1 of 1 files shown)


elmiko (commented):
this is exactly what we want.  we chatted on slack about adding  some unit tests too, i think it would be nice to just add a test file that checks the functions in this util.go.

racheljpg:
Thanks, Mike! A link to the unit test PR: https://github.com/openshift/machine-api-operator/pull/1227

elmiko (commented):
/approve
/lgtm

elmiko:
/hold

you need to update the unit tests once they merge, or add this change to the other PR

elmiko:
@racheljpg for future reference, i would just fix the test error here instead of making a new pr.

elmiko:
feel free to remove the hold here once the other PR merges

PR #1235, opened 2024-04-30
OCPCLOUD-2493: Add case in unit test to check for empty GpuType Value if GpuCount is 0

racheljpg:
Hello! This is a PR to add a case to the unit test to make sure that if the GpuCount annotation is set to 0, that the GpuType annotation should return an empty string. Thanks!


elmiko:
/hold

i think we want the test in the CAO

racheljpg:
Closing this as this case is a bit redundant (with another case already testing this specific setter) and to move this logic to the CAO (ref Mike's comment above).

PR #1237, opened 2024-05-07
handle nodes without UID

freedge:
On a large deployment on OpenStack, with high latency, we observe the mhc controller remediating nodes immediately after their creation.  The remediation happens without any log for its cause, and the only path in the code where that is the case is when the node.UID is empty.  We do see a node Name in the target string() that is logged, so we know that a node.Name is present at that time.

Change the code so that a node is considered to be existing when it has a name and no UID, even if the circumstances by which this happens are blurry.

Also add a log so that all code paths leading to remediation are clearly seen.

### DIFF
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
@@ -788,7 +788,8 @@ func (t *target) needsRemediation(timeoutForMachineToHaveNode time.Duration) (bo
 	}
 
 	// the node does not exist
-	if t.Node != nil && t.Node.UID == "" {
+	if t.Node != nil && t.Node.UID == "" && t.Node.Name == "" {
+		klog.V(3).Infof("%s: unhealthy: node does not exist", t.string())
 		return true, time.Duration(0), nil
 	}
 
--- a/pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
+++ b/pkg/controller/machinehealthcheck/machinehealthcheck_controller_test.go
@@ -1964,6 +1964,54 @@ func TestNeedsRemediation(t *testing.T) {
 			expectedNextCheck:           1 * time.Minute, // 300-200 rounded
 			expectedError:               false,
 		},
+		{
+			testCase: "unhealthy: node does not fully exist",
+			target: &target{
+				Machine: *maotesting.NewMachine("test", "node"),
+				Node: &corev1.Node{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "node",
+						UID:  "",
+					},
+					TypeMeta: metav1.TypeMeta{
+						Kind: "Node",
+					},
+				},
+				MHC: machinev1.MachineHealthCheck{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "test",
+						Namespace: namespace,
+					},
+					TypeMeta: metav1.TypeMeta{
+						Kind: "MachineHealthCheck",
+					},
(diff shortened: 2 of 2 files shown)


pkg/controller/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -788,7 +788,8 @@ func (t *target) needsRemediation(timeoutForMachineToHaveNode time.Duration) (bo
 	}
 
 	// the node does not exist
-	if t.Node != nil && t.Node.UID == "" {
+	if t.Node != nil && t.Node.UID == "" && t.Node.Name == "" {
```
slintes, line 791: This will never be true:

- The node is nil when noderef isn't set on the Machine yet.
- When noderef is set, and the Node was found, it has a UID.
- When the Node wasn´t found, the name is set by MHC code based on the noderef, but UID stays empty.

So we can never have a non nil node without name.

See https://github.com/openshift/machine-api-operator/blob/45519639494da10af8dbb5526be9e2d2311abcb4/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go#L522

slintes (changes requested):
The intention of the code you identified is to remediate machines which should have a node (noderef on the Machine is set), but the node doesn't exist. And that's probably the correct code place for the node deletion you're observing.

However:

>  We do see a node Name in the target string() that is logged, so we know that a node.Name is present at that time.

It's a wrong assumption though, that we found the node when we have its name, see inline comment.

I think we have a kind of race condition here: when the noderef on the Machine set, a reconcile is triggered. However, the node might not be in the MHC controller's cache yet, probably because of " large deployment on OpenStack, with high latency".  We would need to better deal with that case, maybe with a "retry to get the node once" approach... 🤔

freedge:
thank you for the review!

> retry to get the node once

couldnt we just do

https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/machinehealthcheck/machinehealthcheck_controller.go#L769-L770

```if t.Node == nil || t.Node.UID == ""```

so that the case where Node is empty is equivalent to the case where the node is not found? (and the logic with nodeStartupTimeout applies)

I had another similar issue on https://github.com/openshift/machine-config-operator/pull/4357 then I think.

Also I did not report this I see our mhc are performing around 50 node healthcheck per second - I'm wondering whether every single machine would trigger a healthcheck to be triggered in the next 5 minutes (assuming that's what configured as unhealthy condition timeout) - in that case a mhc targetting 35 nodes would trigger 35 * 35 checks per 5 minutes.

(I appreciate the time you took to reply to my pr! I can open a OCPBUGS or a customer case if you feel this is more appropriate though. Thanks a ton!).

slintes:
Hi, sorry for the delay, there was some PTO last week.

> couldnt we just do `if t.Node == nil || t.Node.UID == ""` ...

That would make remediation slower for those who really have an issue with the node not being created 🤔 

Yes, please create an OCPBUGS issue (Cloud Compute / MachineHealthCheck), I'm afraid this needs some more consideration. Feel free to assign to me, I will discuss with the team then what's the best solution here.

About the nr of healthchecks: changes on MHCs, Machines or Nodes trigger a healthcheck. The unhealthy condition timeout only applies if a node condition matches an unhealthy condition, then another healthcheck is scheduled after the timeout. If this doesn't explain what you see, please create a new issue for it, ideally with MHC logs.

freedge:
> If this doesn't explain what you see
also opening https://issues.redhat.com/browse/OCPBUGS-33586
I don't know if I can attach data to it myself and might end up creating a case

PR #1259, opened 2024-06-11
OCPBUGS-34148: manifests: gcp add compute.instanceGroups.use permission

damdo:
For gcp, add compute.instanceGroups.use permission as that is required in some QE environment.


damdo:
cc. @miyadav @huali9 
Would you be able to test if with this change OCPBUGS-34148 improves?

damdo:
Not required anymore. Solved in a different way

PR #1262, opened 2024-06-12
OCPBUGS-1735: Vsphere: Handle cloned instance with lost taskID fix reproducing PR

RadekManak:
/hold

This PR is the same as https://github.com/openshift/machine-api-operator/pull/1223 with an additional commit that discards the machine creation taskID. Simulating the issue to allow QE to verify the fix.

Some unit tests will fail here, but that is expected. Machine creation should still work, leveraging the new logic.

"Powering on cloned machine without taskID: %v", r.machine.Name
%v: already exists, but was not powered on after clone"
Should be present in machine controller logs.


RadekManak:
This test PR is no longer needed.

PR #1275, opened 2024-08-12
OCPCLOUD-2565: Updates operator to pass feature gates flags to Machine controllers

theobarberbany:
This change:
- Updates the operator to pass the `--feature-gate` argument to the machine controller container.
- Updates the operator to pass the feature gates required by the operands
- Updates the MahineSet operator to use the new library-go helperfunc, `AddFlagsToGoFlagSet` to keep the way we handle feature gates consistent across all controllers. 

Until providers are updated we want a hold
/hold


theobarberbany:
Just waiting on powervs:  https://github.com/openshift/machine-api-provider-powervs/pull/80

theobarberbany:
> is it possible to disable the default enabled feature gates?

Yep! It reads from the features definitions in `o/api`. This is handled by the implementation in the operand.

e.g  the following would be defaulted to `True` 

```
enableIn(configv1.Default, configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade)`
```
and this would be `False`

```
enableIn( configv1.DevPreviewNoUpgrade, configv1.TechPreviewNoUpgrade)`
```

elmiko:
thanks Theo, changes make sense to me. would like to see the gcp tests pass.

/lgtm
/retest

PR #1294, opened 2024-09-27
OCPBUGS-42529: Updates paused condition messages

theobarberbany:
This change updates the Paused condition:

- Use the same reasons across MachineSet and Machine
- If `Status.AuthoritativeAPI` is not set, update the condition message to reflect this.


JoelSpeed (commented):
/lgtm
/approve

PR #1323, opened 2025-01-20
SPLAT-1800: vSphere host zonal missing operator feature

jcpowermac:
Follow up to #1285.
This PR adds `FeatureGateVSphereHostVMGroupZonal` to the map of features so the command line of the controller pods is set correctly.


jcpowermac:
tested locally working as expected

/assign @nrb

PR #1353, opened 2025-04-16
OCPBUGS-55083: Updates GCP CredentialsRequest

theobarberbany:
- Adds compute.images.useReadOnly, which is required for provisioning additional disks.


nrb:
The e2es on GCP passed, and this is a platform specific change.

CI (particularly AWS) is still recovering from quota issues, so I'm going to override the tests on non-GCP platforms.

/override ci/prow/e2e-aws-ovn ci/prow/e2e-aws-operator

PR #1357, opened 2025-04-28
OCPBUGS-55417: Resync CRDs

damdo:
After the openshift/api changes in https://github.com/openshift/api/pull/2279
we need to revendor and sync CRDs in MAO


damdo:
cc. @nrb @theobarberbany

PR #1360, opened 2025-04-29
OCPBUGS-55461: Fix updateMachineSetStatus nil pointer

theobarberbany:
This change updates the error log to rely on the machineSetCopy, rather than the machineSet passed to be updated.

This is because the machineSet may be nil if the update fails, due to the get at the end of updateMachineSetStatus.


damdo (commented):
/approve
/lgtm

PR #1370, opened 2025-05-13
OCPBUGS-56159: correct aws kms permissions

Phaow:
- In **[AWS official doc](https://docs.aws.amazon.com/ebs/latest/userguide/ebs-encryption-requirements.html)** the ebs volume kms encryption needs the follow permissions ->
```console
- kms:CreateGrant
- kms:Decrypt
- kms:DescribeKey
- kms:GenerateDataKeyWithoutPlainText
- kms:ReEncrypt
```
Using customer kms key create machine in the local zone failed caused by missed `kms:ReEncrypt` permission.
image

### DIFF
--- a/install/0000_30_machine-api-operator_00_credentials-request.yaml
+++ b/install/0000_30_machine-api-operator_00_credentials-request.yaml
@@ -45,6 +45,7 @@ spec:
       resource: "*"
     - effect: Allow
       action:
+      - 'kms:ReEncrypt*'
       - kms:Decrypt
       - kms:Encrypt
       - kms:GenerateDataKey


Phaow:
/hold
wait for testing

install/0000_30_machine-api-operator_00_credentials-request.yaml
```diff
@@ -45,6 +45,7 @@ spec:
       resource: "*"
     - effect: Allow
       action:
+      - 'kms:ReEncrypt*'
```
mtulio, line 48: Hi @Phaow @JoelSpeed , would the `kms:ReEncrypt` (without star) resolve the problem?

Phaow: Hi @mtulio , actually there's no standlone `kms:ReEncrypt`  I think it need both `ReEncryptFrom` and `ReEncryptTo` in local zone, and this permission keep consistent with storage operator -> https://github.com/openshift/cluster-storage-operator/blob/main/manifests/03_credentials_request_aws.yaml#L42 and CI customer kms configuration.

> Unlike other AWS KMS API operations, ReEncrypt callers must have two permissions:
> kms:ReEncryptFrom permission on the source CMK
> kms:ReEncryptTo permission on the destination CMK

https://docs.aws.amazon.com/cli/latest/reference/kms/re-encrypt.html

```console

Term | Meaning
-- | --
kms:ReEncrypt | ❌ Invalid — not an actual action
kms:ReEncrypt* | ✅ Wildcard — includes both ReEncryptFrom and ReEncryptTo
kms:ReEncryptFrom | ✅ Decrypt part of re-encryption
kms:ReEncryptTo | ✅ Encrypt part of re-encryption


```

mtulio: Apologies, I answered quickly. Sure, I wanted ask if we could use explicit policy definition (without star), expanding the required ones, like the ones you've mentioned: `kms:ReEncryptFrom` and `kms:ReEncryptTo`.

Phaow: I think it also works to declare all. In the AWS official doc also recommend with `kms:ReEncrypt*`, is there any concern we have to declare the whole without any wildcard?
```
To permit reencryption from or to a KMS key, include the "kms:ReEncrypt*" permission in your [key policy](https://docs.aws.amazon.com/kms/latest/developerguide/key-policies.html) . This permission is automatically included in the key policy when you use the console to create a KMS key. But you must include it manually when you create a KMS key programmatically or when you use the PutKeyPolicy operation to set a key policy.
```

PR #1375, opened 2025-06-03
OCPBUGS-57018: Fixed test logic for UPI scenario

vr4manta:
OCPBUGS-57018

### Changes
- Fixed checks relating to length of arrays
- Added skip for when no machines / UPI scenario


rvanderp3:
/title OCPBUGS-57018,OCPBUGS-57059: Fixed test logic for UPI scenario

PR #1380, opened 2025-06-18
fix: controllers: guard on empty .status.authoritativeAPI

damdo:


pkg/controller/machine/controller.go
```diff
@@ -177,43 +177,46 @@ func (r *ReconcileMachine) Reconcile(ctx context.Context, request reconcile.Requ
...
-
-		// Set the paused condition to false, continue reconciliation
+		// Set the paused condition to false, continue reconciliation.
 		conditions.Set(m, conditions.FalseCondition(
 			PausedCondition,
 			NotPausedConditionReason,
 			machinev1.ConditionSeverityInfo,
 			"%s",
-			pausedFalseReason,
+			fmt.Sprintf("The AuthoritativeAPI is set to '%s'", string(m.Status.AuthoritativeAPI)),
```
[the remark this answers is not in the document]
theobarberbany: Right, I see.  

I think from a UX perspective having both migrating and empty may be confusing. 

How is a cluster operator meant to interpret empty, without knowing how the controllers work?

pkg/controller/machine/controller.go
```diff
@@ -177,43 +177,46 @@ func (r *ReconcileMachine) Reconcile(ctx context.Context, request reconcile.Requ
...
 				PausedConditionReason,
-				"The AuthoritativeAPI is set to %s", string(m.Status.AuthoritativeAPI),
+				"The AuthoritativeAPI is set to '%s'", string(m.Status.AuthoritativeAPI),
 			))
-			if patchErr := r.updateStatus(ctx, m, ptr.Deref(m.Status.Phase, ""), nil, originalConditions); patchErr != nil {
+			if patchErr := r.updateStatus(ctx, m, ptr.Deref(m.Status.Phase, ""), nil, originalConditions); patchErr != nil && !apierrors.IsNotFound(patchErr) {
 				klog.Errorf("%v: error patching status: %v", machineName, patchErr)
+				return reconcile.Result{}, fmt.Errorf("error patching status: %w", patchErr)
+			} else if apierrors.IsNotFound(patchErr) {
+				return reconcile.Result{}, nil
```
mdbooth, line 191: Why are we special casing this? Does it happen a lot? This would suggest that something else has deleted the machine before we got here, which is possible but:
* it seems unlikely
* just throwing the error would be safe because we'd re-reconcile and get the 'normal' object deleted behaviour

If this does happen a lot for some reason, could we have a comment here explaining why? Otherwise I'm in danger of getting stuck down this rabbit hole every time I see this code.

pkg/controller/machine/controller.go
```diff
@@ -177,43 +177,46 @@ func (r *ReconcileMachine) Reconcile(ctx context.Context, request reconcile.Requ
...
 			"%s",
-			pausedFalseReason,
+			fmt.Sprintf("The AuthoritativeAPI is set to '%s'", string(m.Status.AuthoritativeAPI)),
 		))
-		if patchErr := r.updateStatus(ctx, m, ptr.Deref(m.Status.Phase, ""), nil, originalConditions); patchErr != nil {
+		if patchErr := r.updateStatus(ctx, m, ptr.Deref(m.Status.Phase, ""), nil, originalConditions); patchErr != nil && !apierrors.IsNotFound(patchErr) {
 			klog.Errorf("%v: error patching status: %v", machineName, patchErr)
+			return reconcile.Result{}, fmt.Errorf("error patching status: %w", patchErr)
+		} else if apierrors.IsNotFound(patchErr) {
+			return reconcile.Result{}, nil
```
mdbooth, line 210: Same as above. A comment would be nice if this is really necessary.

pkg/controller/machine/controller.go
```diff
@@ -177,43 +177,46 @@ func (r *ReconcileMachine) Reconcile(ctx context.Context, request reconcile.Requ
...
+		}
+
+		klog.Infof("%v: setting machine paused condition to false", machineName)
+
+		// Here we want to check again the .status.AuthoritativeAPI is set to MachineAPI
+		// as the above updateStatus does a get of the latest machine and the value could be anything again at that point.
```
mdbooth, line 216: Is there any reason we don't just let c-r reconcile the object again after patching the status?

pkg/controller/machine/controller.go
```diff
@@ -177,43 +177,46 @@ func (r *ReconcileMachine) Reconcile(ctx context.Context, request reconcile.Requ
...
+		if m.Status.AuthoritativeAPI != machinev1.MachineAuthorityMachineAPI {
```
mdbooth, line 217: Combined with the above check this feels very tightly-coupled. Could we write this as:

```
switch status.authAPI {
case "":
  paused = true
  reason = "Waiting for migration controller"

  return ctrl.Result{} so we're rescheduled

case "ClusterAPI":
  paused = true
  reason = "Not our problem"

  return ctrl.Result{} so we're rescheduled

case "MachineAPI":
  if paused == true {
    paused = false
    reason = "It's our problem"

    return ctrl.Result{} so we're rescheduled
  }

  fallthrough: It's our object and we're already unpaused, so do the reconcile.
}
```

I think this would be much easier to reason about.

damdo: I tend to agree this could be clearer and thought about this too, but didn't want to necessarily refactor everything.

pkg/controller/machine/controller.go
damdo: Same as above

pkg/controller/machine/controller.go
damdo: The status patching utility does this for archeological reasons, I don't necessarily want to dig into that and go down that rabbit hole, I think it is fine as is, this is just a targeted bugfix and I do not intend to rewrite status updates which already work.

pkg/controller/machine/controller.go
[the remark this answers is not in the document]
damdo: I reworked this and so given we have a specific switch/case branch for `""`, I added `is not set` :)

pkg/controller/machine/controller.go
damdo: Got rid of it

pkg/controller/machine/controller.go
damdo: Done, I've implemented logic changes along these lines, as I think it improves readability, understanding and logging.

pkg/controller/machineset/controller.go
```diff
@@ -176,57 +176,72 @@ func (r *ReconcileMachineSet) Reconcile(ctx context.Context, request reconcile.R
...
+				"The AuthoritativeAPI status is not yet set",
+			)
+
+			if !conditions.IsEquivalentTo(conditions.Get(machineSet, machine.PausedCondition), newCondition) {
+				klog.Infof("%v: machine set .status.authoritativeAPI is not yet set, ensuring machine set is paused", machineSetName)
+				conditions.Set(machineSetCopy, newCondition)
+				if _, err := updateMachineSetStatus(r.Client, machineSet, machineSetCopy.Status); err != nil {
+					klog.Errorf("%v: error updating status: %v", machineSetName, err)
+					return reconcile.Result{}, fmt.Errorf("error updating status for machine set %s: %w", machineSetName, err)
+				}
```
damdo, line 196: There is a slight semantic/behavioural change here.
Now in the `.status.AuthoritativeAPI: ""` case we are also setting Paused: True, rather than only just returning, I suppose this is semantically more aligned with our desire of only unpausing the controllers when they are authoritative. @JoelSpeed sounds reasonable?

JoelSpeed: SGTM

PR #1386, opened 2025-07-02
OCPCLOUD-2986,OCPCLOUD-2985,OCPBUGS-56849: Fix controllers guard on empty status authoritative api + defaulting

damdo:
This PR:
1. Reintroduces/Un-reverts https://github.com/openshift/machine-api-operator/pull/1380 (originally reverted by https://github.com/openshift/machine-api-operator/pull/1383) **(first commit)** 
1. It fixes the original issue which prevented non-AWS TechPreview clusters to not be able to reconcile MAPI resources when .status.authoritativeAPI was empty/not set by **implementing defaulting of the Authoritative API from spec to status (when not set) for MAPI Machines/MachineSets** (which was something we were ought to do anyway https://issues.redhat.com/browse/OCPCLOUD-2985). For this we took a slight different approach than the agreed MutatingAdmissionPolicy, as we found out that it wouldn't work on CREATE for status subresource (context: https://redhat-internal.slack.com/archives/GE2HQ9QP4/p1751450680311659) **(second commit)**

### DIFF
--- a/pkg/controller/machine/controller.go
+++ b/pkg/controller/machine/controller.go
@@ -179,21 +179,21 @@ func (r *ReconcileMachine) Reconcile(ctx context.Context, request reconcile.Requ
 	if r.gate.Enabled(featuregate.Feature(openshiftfeatures.FeatureGateMachineAPIMigration)) {
 		switch m.Status.AuthoritativeAPI {
 		case "":
-			// An empty .status.authoritativeAPI normally means the resource has not yet been reconciled
-			// by the migration controller, which has the responsibility for propagating .spec.authoritativeAPI to the status.
-			// Pause the resource and take no further action but return until that field is populated.
-			desiredCondition := conditions.TrueConditionWithReason(
-				PausedCondition, PausedConditionReason,
-				"The AuthoritativeAPI status is not yet set",
-			)
+			// An empty .status.authoritativeAPI normally means the resource has not yet been reconciled.
+			// and that the value in .spec.authoritativeAPI has not been propagated to .status.authoritativeAPI yet.
+			// This value can be set by two separate controllers, depending which one of them is running at that time,
+			// or in case they are both running, which one gets to set it first (the operation is idempotent so there is no harm in racing).
+			// - the cluster-capi-operator machine-api-migration's migration controller
+			// - this controller
 
-			if _, err := r.ensureUpdatedPausedCondition(ctx, m, desiredCondition,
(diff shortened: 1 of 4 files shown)


damdo:
/assign  @JoelSpeed @mdbooth 
As you reviewed the previous one

damdo:
/assign @JoelSpeed @mdbooth

As you reviewed the previous one

damdo:
/hold

For pre-merge testing

pkg/controller/machine/controller.go
```diff
@@ -177,43 +177,60 @@ func (r *ReconcileMachine) Reconcile(ctx context.Context, request reconcile.Requ
...
+			// or in case they are both running, which one gets to set it first (the operation is idempotent so there is no harm in racing).
+			// - the cluster-capi-operator machine-api-migration's migration controller
+			// - this controller
+
+			klog.Infof("%v: machine .status.authoritativeAPI is not yet set, setting it to .spec.authoritativeAPI", m.Name)
+
+			if err := r.patchStatusAuthoritativeAPI(ctx, m, m.Spec.AuthoritativeAPI); err != nil {
+				klog.Errorf("%v: error patching status to set .status.authoritativeAPI for machine: %v", m.Name, err)
+				return reconcile.Result{}, fmt.Errorf("error patching status to set .status.authoritativeAPI for machine %s: %w", m.Name, err)
+			}
```
mdbooth, line 194: I still worry slightly about duplication of this logic between 2 controllers, especially in separate repos.

* What would happen if we decided to change the 'empty status -> default from spec' logic in the migration controller?
* Can we think of any reason we would ever do that?

I think this is probably ok, just putting on my pessimistic hat to think of problems it could cause.

An alternative to this would be to always allow the migration controller to run, even when the platform is not supported. We could then implement the behaviour for unsupported platforms in one place, which could include defaulting status.authAPI. That feels cleaner to me and it's probably not too much effort, but more than this. Would it have any other benefits to make it worthwhile?

mdbooth: [Discussed in slack.](https://redhat-internal.slack.com/archives/GE2HQ9QP4/p1751536658080839)

I'm happy that we've considered the points here.

damdo:
Thanks both! @sunzhaohua2 please do let us know when the testing is done and if we can unhold it.

damdo:
More context on the above issue reported here, the further testing done and the measures that were taken:
https://redhat-internal.slack.com/archives/GE2HQ9QP4/p1751450680311659

We can proceed with merging this.

damdo:
Thanks a lot @sunzhaohua2 

/unhold

PR #1389, opened 2025-07-09
Changes to vm-host zonal e2e

jcpowermac:


jcpowermac:
the test needs to fail to see if the issue is fixed, try again...

/test e2e-vsphere-host-groups-ovn

jcpowermac:
ugh the test passed :) let's try again


/test e2e-vsphere-host-groups-ovn

PR #1407, opened 2025-09-01
OCPBUGS-60901: bump openshift/api

chrischdi:
Pulls in https://github.com/openshift/api/pull/2462


chrischdi:
Waiting for lifting the `acknowledge-critical-fixes-only` requirement :-)

chrischdi:
Yes, Imho we should backport to 4.20, because the API change is already in there.

@damdo, Question: Should we also backport to 4.19 together with https://github.com/openshift/api/pull/2464 ?

PR #1414, opened 2025-09-09
NO-JIRA: Update OWNERS_ALIASES

RadekManak:
https://xkcd.com/1205/


damdo:
/verified bypass

This is an OWNERS change, no need to verify

PR #1419, opened 2025-09-18
SPLAT-2167: Added dedicated hosts support for AWS

vr4manta:
Subsequently reverted by: https://github.com/openshift/machine-api-operator/pull/1434

--

[SPLAT-2167](https://issues.redhat.com//browse/SPLAT-2167)

### Changes
- Added logic to webhook to validate fields are correct

### Dependencies
- https://github.com/openshift/api/pull/2484

### DIFF
--- a/cmd/machine-api-tests-ext/provider.go
+++ b/cmd/machine-api-tests-ext/provider.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -10,10 +9,7 @@ import (
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
-	corev1 "k8s.io/api/core/v1"
-	kclientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/kubernetes/openshift-hack/e2e"
 	conformancetestdata "k8s.io/kubernetes/test/conformance/testdata"
 	"k8s.io/kubernetes/test/e2e/framework"
 	"k8s.io/kubernetes/test/e2e/framework/testfiles"
@@ -89,11 +85,6 @@ func initializeTestFramework(provider string) error {
 		return err
 	}
 	testContext.Host = cfg.Host
-
-	// Ensure that Kube tests run privileged (like they do upstream)
-	testContext.CreateTestingNS = func(ctx context.Context, baseName string, c kclientset.Interface, labels map[string]string) (*corev1.Namespace, error) {
-		return e2e.CreateTestingNS(ctx, baseName, c, labels, true)
-	}
 
 	gomega.RegisterFailHandler(ginkgo.Fail)
 


vr4manta:
/payload-job periodic-ci-openshift-release-master-nightly-4.21-e2e-vsphere-ovn-serial

pkg/webhooks/machine_webhook.go
```diff
@@ -897,6 +897,32 @@ func validateAWS(m *machinev1beta1.Machine, config *admissionConfig) (bool, []st
...
+			switch *placement.Affinity {
+			case machinev1beta1.HostAffinityAnyAvailable:
+				// Cannot have DedicatedHost set
+				if placement.DedicatedHost != nil {
+					errs = append(errs, field.Forbidden(field.NewPath("spec.hostPlacement.dedicatedHost"), "dedicatedHost is not allowed when affinity is AnyAvailable"))
+				}
+			case machinev1beta1.HostAffinityDedicatedHost:
+				// We need to make sure DedicatedHost is set with a HostID
+				if placement.DedicatedHost == nil {
+					errs = append(errs, field.Required(field.NewPath("spec.hostPlacement.dedicatedHost"), "dedicatedHost is required when affinity is DedicatedHost"))
```
everettraven: Same terminology as is present in the CEL expression for o/api:

```suggestion
					errs = append(errs, field.Required(field.NewPath("spec.hostPlacement.dedicatedHost"), "dedicatedHost is required when affinity is DedicatedHost, and forbidden otherwise"))
```

pkg/webhooks/machine_webhook.go
```diff
@@ -897,6 +902,33 @@ func validateAWS(m *machinev1beta1.Machine, config *admissionConfig) (bool, []st
...
+			case machinev1beta1.HostAffinityAnyAvailable:
+				// Cannot have DedicatedHost set
+				if placement.DedicatedHost != nil {
+					errs = append(errs, field.Forbidden(field.NewPath("spec.hostPlacement.dedicatedHost"), "dedicatedHost is required when affinity is DedicatedHost, and forbidden otherwise"))
+				}
+			case machinev1beta1.HostAffinityDedicatedHost:
+				// We need to make sure DedicatedHost is set with a HostID
+				if placement.DedicatedHost == nil {
+					errs = append(errs, field.Required(field.NewPath("spec.hostPlacement.dedicatedHost"), "dedicatedHost is required when affinity is DedicatedHost, and forbidden otherwise"))
+				} else if awsDedicatedHostNamePattern.FindStringSubmatch(placement.DedicatedHost.ID) == nil {
```
everettraven: Would it make sense to split this into two separate checks so you can return a `required` and an `invalid` type of field error based on provided value?

I.e if the value is `""`, return an error stating that the field is required along with the prose for the regex.

If the value is not `""`, if it doesn't match the regex return an error stating invalid with the prose for the regex.

Having the separate checks may also read a bit easier for people looking at this in the future.

everettraven (commented):
Validations on the webhook LGTM

damdo (commented):
/approve


/assign @everettraven 
For LGTM

PR #1424, opened 2025-10-27
Add mapi azure datadisk e2e

jcpowermac:
This commit adds e2e tests for the azure
data disk feature gate.

Assisted with Claude

### DIFF
--- a/cmd/machine-api-tests-ext/main.go
+++ b/cmd/machine-api-tests-ext/main.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/component-base/logs"
 
 	// If using ginkgo, import your tests here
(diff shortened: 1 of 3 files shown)


damdo:
cc. @shellyyang1989 @sunzhaohua2 @huali9 @miyadav

sunzhaohua2:
@shellyyang1989 is the MAPI E2E need to write to the [cluster-api-actuator-pkg](https://github.com/openshift/cluster-api-actuator-pkg) repo?

@jcpowermac thank you to help add machine api e2e, I don't quite understand why this feature requires a techpreview job, and it would be better if add testing of different deletion policies (delete and detach).

jcpowermac:
@sunzhaohua2 because the feature is currently in tech preview and we need to have at least 5 tests to promote to ga.

PR #1430, opened 2025-11-07
CORS-4280: AWS: Add throughput validation for gp3 volumes

jhixson74:
Throughput can only be configured on gp3 volumes and must be between 125 and 2000 MiB's

This pull request depends on https://github.com/openshift/api/pull/2480


damdo (commented):
/approve

/hold 

For other reviews and in case you want to address my personal preference

damdo:
@jhixson74 Needs rebasing

jhixson74:
> @jhixson74 looks like something might still be off

I am not seeing anything other than the expected ThroughputMib error. I tested this locally with a patched openshift/api and build and tests were fine. What is off?

PR #1433, opened 2025-11-14
Fix VSphereMultiNetworks tests failing with 'cannot clone suite after…

jcpowermac:
… tree has been built'

After bumping to k8s 1.34 (commit dd24836d9), which updated ginkgo/v2 from v2.23.4 to v2.27.2, all VSphereMultiNetworks tests began failing immediately with the error "cannot clone suite after tree has been built".

Root cause: The newer Ginkgo version enforces stricter checks preventing suite cloning after the tree has been built. When tests run in parallel, child processes are spawned with the "run-test" command, which re-enters main() and calls BuildExtensionTestSpecsFromOpenShiftGinkgoSuite() again, rebuilding the Ginkgo suite tree. When the child process attempts to run the test, Ginkgo's RunSpecs() calls PushClone() which fails because the tree was already built.

Solution: Only build the Ginkgo test specs when NOT in "run-test" mode. This prevents child processes from rebuilding the suite tree, allowing the clone operation to succeed during parallel test execution.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/machine-api-tests-ext/main.go
+++ b/cmd/machine-api-tests-ext/main.go
@@ -45,36 +45,42 @@ func main() {
 		Qualifiers: []string{`labels.exists(l, l == "Serial") && labels.exists(l, l == "Conformance")`},
 	})
 
-	// Build our specs from ginkgo
-	specs, err := g.BuildExtensionTestSpecsFromOpenShiftGinkgoSuite()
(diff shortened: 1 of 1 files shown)


jcpowermac:
doesn't work
/close

PR #1440, opened 2025-12-02
OCPBUGS-66244: azure: remove image defaulting

patrickdillon:
Note: For now we went for an alternative approach here: https://github.com/openshift/machine-api-operator/pull/1441

--

Removes the default image value populated by the webhook. Uses existing validation to ensure the image id is populated.

The installer switched to defaulting to marketplace images, and no longer creates a cluster-specific image gallery (or managed image). So the default value populated by the webhook is no longer correct. The behavior now is similar to AWS, which requires an image be specified. 

We will also need to fix the tests to provide the image value.

/hold

I ran out of time, and still need to fix up a few unit tests, but wanted to get this open for feedback.

I realize this is not ideal to remove the defaulting behavior, but the existing defaulting behavior is quite buggy as well.

### DIFF
--- a/pkg/webhooks/machine_webhook.go
+++ b/pkg/webhooks/machine_webhook.go
@@ -65,18 +65,6 @@ var (
 	defaultAzureNetworkResourceGroup = func(clusterID string) string {
 		return fmt.Sprintf("%s-rg", clusterID)
 	}
-	defaultAzureImageResourceID = func(clusterID string) string {
-		// image gallery names cannot have dashes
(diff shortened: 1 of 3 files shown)


patrickdillon:
> > When did the installer switch from gen1 to gen2 VMs by default?
> 
> Gen2-image support was added in 4.11. We default to gen2 images for compatible instance types.

But it looks like MAO defaults to a Gen1-only instance type for compute nodes, so the default makes sense.

PR #1441, opened 2025-12-03
OCPBUGS-66244: default Azure to marketplace image

patrickdillon:
Rather than use the AWS approach proposed in #1440, this proposes to use a similar approach to GCP of hardcoding a 
marketplace image default value. 

In 4.21, the installer stopped creating a cluster-specific image gallery and moved to marketplace images. This PR updates the default for a minimal provider spec to use a hard-coded marketplace image, similar to the approach used for GCP.

### DIFF
--- a/pkg/webhooks/machine_webhook.go
+++ b/pkg/webhooks/machine_webhook.go
@@ -65,23 +65,26 @@ var (
 	defaultAzureNetworkResourceGroup = func(clusterID string) string {
 		return fmt.Sprintf("%s-rg", clusterID)
 	}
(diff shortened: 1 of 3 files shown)


damdo (commented):
/approve
/lgtm

Thanks @patrickdillon 
Looks like a better approach than #1440 in the short term as it avoids an immediate user facing breaking change. We can iterate over the idea at #1440 though and see if we can find a way to do that in a non-breaking way.

Thanks!

damdo:

This fixes a "critical" bug that impairs Machine API functionality on Azure

PR #1443, opened 2025-12-04
OCPBUGS-66940: fix default image for confidential VMs

patrickdillon:
The installer still creates an image gallery for confidfential VMs, and we can tell its a confidential VM from the machine provider spec, so this updates the default to point to an installer-created gallery rather than the marketplace image.


nrb:
/title OCPBUGS-66240: fix default image for confidential VMs

perdasilva:
@tthvo @patrickdillon this has been hanging around for a bit - is it something we want to get merged?

patrickdillon:
@damdo @perdasilva Yes you can merge it. There are no customer cases and I don't believe confidential VMs are a common use case, so I'm also fine just to close it. But AFAIK it's a valid bug and fix.

PR #1448, opened 2025-12-16
OKD-259: Vendor API and add OKD featureset Specific CRDs

jatinsu:
Vendor changes made [here](https://github.com/openshift/api/pull/2451) to add support for the OKD featureset


Prashanth684:
@elmiko could we get an approval here please?

PR #1451, opened 2026-01-08
SPLAT-2603: Updated AWS dedicated host id validation pattern

vr4manta:
[SPLAT-2603](https://issues.redhat.com//browse/SPLAT-2603)

### Changes
- Modified the AWS dedicated host validation pattern to include older deprecated host id pattern of 8 alphanumeric

### Notes
CodeRabbit in a separate PR had found that AWS has two patterns for this.  It seems this pattern for 8 alphanumeric is an older deprecated ID format that was changed in 2018.  We are updating our logic just in case some customer still owns one of these older instances and wants to try to use it with OCP.


vr4manta:
/verified by @vr4manta via manual and unit tests
This is being added to our tests as well.

PR #1465, opened 2026-02-25
vsphere - Clear task ID cache on PatchMachine failure

jcpowermac:
When PatchMachine fails during Create, the stale task ID cache entry would permanently block the machine from progressing past the stale check on every future reconcile. Delete the cache entry on PatchMachine failure so the machine can recover.


perdasilva:
@jcpowermac this has been open for a little while now, is it still relevant?

PR #1476, opened 2026-03-13
fix(machine): propagate MaxConcurrentReconciles to drain controller

loick111:
The drain controller was always created with default options, ignoring the MaxConcurrentReconciles value passed via caller opts. This meant platforms like Azure that set concurrent reconciles to 10 would still have the drain controller running with a single worker, creating a bottleneck during large-scale drain operations.

Propagate MaxConcurrentReconciles from the incoming opts to the drain controller options so both controllers scale equally.

Case: #04399828

### DIFF
--- a/pkg/controller/machine/drain_controller.go
+++ b/pkg/controller/machine/drain_controller.go
@@ -3,6 +3,7 @@ package machine
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"golang.org/x/time/rate"
@@ -38,6 +39,13 @@ type machineDrainController struct {
 	scheme *runtime.Scheme
 
 	eventRecorder record.EventRecorder
+
+	// controlPlaneDrainLock ensures only one control plane node can pass the
+	// isDrainAllowed check and be cordoned at a time. Without this, concurrent
+	// reconciles of different control plane machines can both pass
+	// isDrainAllowed before either cordons, leading to simultaneous control
+	// plane node drains (TOCTOU race).
+	controlPlaneDrainLock sync.Mutex
 }
 
 // newDrainController returns a new reconcile.Reconciler for machine-drain-controller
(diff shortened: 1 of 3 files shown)


JoelSpeed:
/hold

The drain controller contains logic to prevent the concurrent draining of control plane machines. This logic presently isn't thread safe and assumes only a single concurrent reconcile. We cannot adjust the reconcile count for the drain controller until that guard is made thread safe

pkg/controller/machine/drain_controller.go
```diff
@@ -193,6 +195,30 @@ func (d *machineDrainController) drainNode(ctx context.Context, machine *machine
...
+// For uncordoned control plane nodes, the check and cordon are held under
+// controlPlaneDrainLock so that only one CP node at a time can pass the
+// isDrainAllowed check and get cordoned, preventing a TOCTOU race where
+// concurrent reconciles would each see no other CP node cordoned and then
+// all proceed to drain simultaneously.
+func (d *machineDrainController) cordonNode(ctx context.Context, drainer *drain.Helper, node *corev1.Node) error {
+	if isControlPlaneNode(*node) && !node.Spec.Unschedulable {
+		d.controlPlaneDrainLock.Lock()
+		defer d.controlPlaneDrainLock.Unlock()
+	}
```
JoelSpeed, line 208: Maybe move this into `isDrainAllowed`?

I'd expect the flow to be:
* Without mutex, check if another CP machine is draining - if yes, return
* If no - check again, this time under lock - if another CP machine is draining, return
* If no - complete drain (lock should be released ideally)

JoelSpeed: If we want to move forward, I think we need to make this change. The current design will block a thread if there's another thread already holding the lock. We want in this instance to put the request back on the queue and try another node.

perdasilva:
@loick111 this has been open in WIP for a little while now. Is it still relevant?

PR #1488, opened 2026-04-15
OCPBUGS-80829: Bump google.golang.org/grpc to v1.79.3

nindsimonv:
Go version used: go1.25.5

The commands used to generate this commit were:

go get google.golang.org/grpc@v1.79.3
go mod tidy
go mod vendor

Made-with: Cursor


nrb:
Looks like we need to ensure that the context library used is correct.

PR #1489, opened 2026-04-16
Fixes https://redhat.atlassian.net/browse/OCPBUGS-82584

rgordill:
It adds a new status field called selector, with the same logic as other scale subresources (an joined string of spec.selector fields comma separated), and fix the scale subresource definition to point to that field. With that, the HPA and other scaling mechanisms would not raise exceptions.

## Release Notes

* **New Features**
  * Added a new `status.selector` field to MachineSet status that exposes the label selector in string format, enabling proper autoscaler and HPA integration through the scale subresource.

* **Tests**
  * Added test coverage to validate the status selector field correctly matches the spec selector configuration.


RadekManak:
/close
Thanks, I'll merge this on #1490

PR #1494, opened 2026-05-01
NO-JIRA: machine-api-termination-handler: add scc annotation, terminationMessage: FallbackToLogsOnError

damdo:
- [set terminationMessagePolicy on termination-handler container](https://github.com/openshift/machine-api-operator/commit/6af8c5ae4ced936e9609e7fa5de6d12a056d7771) 
The termination-handler container was missing terminationMessagePolicy=FallbackToLogsOnError

- [require machine-api-termination-handler SCC via annotation](https://github.com/openshift/machine-api-operator/commit/e527fdd99cb8ecdf53b8eb437c96d9262bbd35e8) 
pin the termination-handler DaemonSet pods to the dedicated
machine-api-termination-handler SCC using the openshift.io/required-scc annotation.
This makes the SCC binding explicit and prevents silent fallback to a different
SCC if admission ordering changes

## Summary by CodeRabbit

* **Chores**
  * Applied required OpenShift security constraint to the termination-handler pod so it meets platform security requirements.
  * Updated termination behavior to prefer logs on errors for the termination-handler container, improving post-termination diagnostics and error visibility.

### DIFF
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -1067,11 +1067,12 @@ func newTerminationContainers(config *OperatorConfig) []corev1.Container {
 
 	return []corev1.Container{
 		{
-			Name:      "termination-handler",
-			Image:     config.Controllers.TerminationHandler,
-			Command:   []string{"/termination-handler"},
-			Args:      terminationArgs,
-			Resources: resources,
+			Name:                    "termination-handler",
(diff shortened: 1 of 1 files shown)


nrb:
/lgtm
/approve
/verified by checking CI artifacts and observing the SCC is retained

PR #1499, opened 2026-05-13
OCPBUGS-85416: Enable actuators to set terminal failure on Exists and Update

mdbooth:
An actuator can already mutate its Machine argument to specificy an explicit Failed phase. However, because we don't also pass the error, updateStatus() cannot correctly set ErrorMessage and ErrorReason.

## Summary by CodeRabbit

* **Bug Fixes**
  * Machine controller now properly captures and reports error details in machine status when actuator operations fail during reconciliation.

* **Tests**
  * Expanded test coverage to validate error messaging in actuator failure scenarios and ensure status reporting is accurate.

### DIFF
--- a/pkg/controller/machine/controller.go
+++ b/pkg/controller/machine/controller.go
@@ -346,7 +346,7 @@ func (r *ReconcileMachine) Reconcile(ctx context.Context, request reconcile.Requ
 			"Failed to check if machine exists: %v", err,
 		))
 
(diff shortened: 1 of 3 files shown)


damdo (commented):
/approve
/lgtm

mdbooth:
/verified by https://github.com/openshift/machine-api-provider-azure/pull/193

PR #1510, opened 2026-06-15
SPLAT-2790: Added new VAP for vSphere infra validation against current cpms and machines

vr4manta:
[SPLAT-2790](https://redhat.atlassian.net/browse/SPLAT-2790)

### Changes
- Added logic to add new VAP for validating infrastructure (FD) removals against current CPMS and machines

## Summary by CodeRabbit

* **New Features**
  * Added vSphere failure-domain protection using ValidatingAdmissionPolicies and bindings for Machines, ControlPlaneMachineSets, and MachineSets. Infrastructure updates that remove still-referenced failure domains are blocked for the vSphere platform, gated by Multi–vCenter Day 2.
* **Security / RBAC**
  * Extended operator RBAC to manage vSphere ValidatingAdmissionPolicies/Bindings and to read Machines and related machine set resources.
* **Tests**
  * Added synchronization tests and expanded VAP/binding unit coverage, including MachineSet behavior and name/link consistency checks.

### DIFF
--- a/install/0000_30_machine-api-operator_09_rbac.yaml
+++ b/install/0000_30_machine-api-operator_09_rbac.yaml
@@ -472,12 +472,28 @@ rules:
     resources:
       - validatingwebhookconfigurations
       - mutatingwebhookconfigurations
+      - validatingadmissionpolicies
+      - validatingadmissionpolicybindings
     verbs:
       - get
       - list
       - watch
       - create
       - update
+
+  # The machine-api-operator service account must be able to read Machines,
+  # ControlPlaneMachineSets, and MachineSets so the kube-apiserver can resolve
+  # them as VAP params when validating Infrastructure/cluster updates.
+  - apiGroups:
+      - machine.openshift.io
+    resources:
+      - machines
+      - controlplanemachinesets
+      - machinesets
+    verbs:
+      - get
+      - list
+      - watch
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/pkg/operator/sync.go
+++ b/pkg/operator/sync.go
@@ -23,6 +23,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	configv1 "github.com/openshift/api/config/v1"
+	apifeatures "github.com/openshift/api/features"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	utiltls "github.com/openshift/controller-runtime-common/pkg/tls"
 	libgocrypto "github.com/openshift/library-go/pkg/crypto"
@@ -268,6 +269,57 @@ func (optr *Operator) syncWebhookConfiguration(config *OperatorConfig) error {
 			return err
 		}
 	}
+	if config.PlatformType == configv1.VSpherePlatformType {
+		if err := optr.syncVSphereFailureDomainVAPs(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// syncVSphereFailureDomainVAPs ensures that the ValidatingAdmissionPolicies and their
+// bindings for protecting vSphere failure domains are present and up to date.
+// These policies prevent an administrator from removing a failure domain from the
+// Infrastructure/cluster CR while it is still referenced by Machines or
+// ControlPlaneMachineSets managed by the Machine API Operator.
+func (optr *Operator) syncVSphereFailureDomainVAPs() error {
+	if !optr.featureGates.Enabled(apifeatures.FeatureGateVSphereMultiVCenterDay2) {
+		return nil
+	}
+
+	recorder := events.NewLoggingEventRecorder(optr.name, clock.RealClock{})
+
+	if _, _, err := resourceapply.ApplyValidatingAdmissionPolicyV1(context.TODO(),
+		optr.kubeClient.AdmissionregistrationV1(), recorder,
+		mapiwebhooks.NewVSphereFailureDomainMachineVAP(), optr.cache); err != nil {
+		return fmt.Errorf("failed to apply vSphere failure domain Machine ValidatingAdmissionPolicy: %w", err)
+	}
+	if _, _, err := resourceapply.ApplyValidatingAdmissionPolicyBindingV1(context.TODO(),
+		optr.kubeClient.AdmissionregistrationV1(), recorder,
+		mapiwebhooks.NewVSphereFailureDomainMachineVAPBinding(), optr.cache); err != nil {
+		return fmt.Errorf("failed to apply vSphere failure domain Machine ValidatingAdmissionPolicyBinding: %w", err)
+	}
+	if _, _, err := resourceapply.ApplyValidatingAdmissionPolicyV1(context.TODO(),
+		optr.kubeClient.AdmissionregistrationV1(), recorder,
+		mapiwebhooks.NewVSphereFailureDomainCPMSVAP(), optr.cache); err != nil {
+		return fmt.Errorf("failed to apply vSphere failure domain ControlPlaneMachineSet ValidatingAdmissionPolicy: %w", err)
+	}
+	if _, _, err := resourceapply.ApplyValidatingAdmissionPolicyBindingV1(context.TODO(),
+		optr.kubeClient.AdmissionregistrationV1(), recorder,
+		mapiwebhooks.NewVSphereFailureDomainCPMSVAPBinding(), optr.cache); err != nil {
+		return fmt.Errorf("failed to apply vSphere failure domain ControlPlaneMachineSet ValidatingAdmissionPolicyBinding: %w", err)
+	}
+	if _, _, err := resourceapply.ApplyValidatingAdmissionPolicyV1(context.TODO(),
+		optr.kubeClient.AdmissionregistrationV1(), recorder,
+		mapiwebhooks.NewVSphereFailureDomainMachineSetVAP(), optr.cache); err != nil {
+		return fmt.Errorf("failed to apply vSphere failure domain MachineSet ValidatingAdmissionPolicy: %w", err)
+	}
+	if _, _, err := resourceapply.ApplyValidatingAdmissionPolicyBindingV1(context.TODO(),
+		optr.kubeClient.AdmissionregistrationV1(), recorder,
+		mapiwebhooks.NewVSphereFailureDomainMachineSetVAPBinding(), optr.cache); err != nil {
+		return fmt.Errorf("failed to apply vSphere failure domain MachineSet ValidatingAdmissionPolicyBinding: %w", err)
+	}
+
 	return nil
 }
 
--- a/pkg/webhooks/vap.go
+++ b/pkg/webhooks/vap.go
@@ -1,0 +1,357 @@
+package webhooks
+
+import (
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+)
+
+const (
+	// VAPMachineFailureDomainName is the name of the ValidatingAdmissionPolicy that guards
+	// against removing a vSphere failure domain that is still referenced by a Machine.
+	VAPMachineFailureDomainName = "vsphere-failure-domain-in-use-by-machine"
+
+	// VAPCPMSFailureDomainName is the name of the ValidatingAdmissionPolicy that guards
+	// against removing a vSphere failure domain that is still referenced by a ControlPlaneMachineSet.
+	VAPCPMSFailureDomainName = "vsphere-failure-domain-in-use-by-cpms"
+
+	// VAPMachineSetFailureDomainName is the name of the ValidatingAdmissionPolicy that guards
+	// against removing a vSphere failure domain that is still referenced by a MachineSet (including
+	// MachineSets with zero replicas that would have no running Machines to catch the check).
+	VAPMachineSetFailureDomainName = "vsphere-failure-domain-in-use-by-machineset"
+
+	// machineRegionLabel is the label used to identify the region a Machine belongs to.
+	machineRegionLabel = "machine.openshift.io/region"
+
+	// machineZoneLabel is the label used to identify the availability zone a Machine belongs to.
+	machineZoneLabel = "machine.openshift.io/zone"
+
+	// vspherePlatformType is the platform type string as used in infrastructure.status.platformStatus.type.
+	vspherePlatformType = "VSphere"
+
+	// openMachineAPINamespace is the namespace where Machines and CPMS live.
+	openMachineAPINamespace = "openshift-machine-api"
+)
+
+var (
+	// vapDenyAction is the enforcement action that denies the admission request.
+	vapDenyAction = admissionregistrationv1.Deny
+
+	// vapParamNotFoundAllow means: if no param object exists (e.g. no Machines yet), allow the infra update.
+	vapParamNotFoundAllow = admissionregistrationv1.AllowAction
+)
+
+// NewVSphereFailureDomainMachineVAP returns a ValidatingAdmissionPolicy that prevents
+// an infrastructure/cluster UPDATE from removing a vSphere failure domain that is still
+// referenced by at least one Machine (identified via machine.openshift.io/region and
+// machine.openshift.io/zone labels).
+//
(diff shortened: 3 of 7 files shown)


install/0000_30_machine-api-operator_09_rbac.yaml
```diff
@@ -472,13 +472,28 @@ rules:
...
+  # when validating Infrastructure/cluster updates.
+  - apiGroups:
+      - machine.openshift.io
+    resources:
+      - machines
+      - controlplanemachinesets
+    verbs:
+      - get
+      - list
+      - watch
```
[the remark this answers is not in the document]
vr4manta: With the suggested change,  this results in the following error when installing the cluster:
```
I0615 14:17:26.642012       1 sync.go:116] Suppressing Degraded status during upgrade: error syncing machine API webhook configurations: ValidatingAdmissionPolicy.admissionregistration.k8s.io "vsphere-failure-domain-in-use-by-machine" is invalid: spec.paramKind: Forbidden: user &{system:serviceaccount:openshift-machine-api:machine-api-operator ef90e6a3-30aa-4540-af49-17f09fb20602 [system:serviceaccounts system:serviceaccounts:openshift-machine-api system:authenticated] map[authentication.kubernetes.io/credential-id:[JTI=0231afaf-2c2b-4557-a229-638c9ab0c183] authentication.kubernetes.io/node-name:[ngirard-multi-tlsw6-master-1] authentication.kubernetes.io/node-uid:[c6cc1934-2db5-495c-bf8a-38fd4f8a1ae4] authentication.kubernetes.io/pod-name:[machine-api-operator-7d6fbc48dc-txtk8] authentication.kubernetes.io/pod-uid:[aad9e358-c073-4ed8-8650-6700c92bdac5]]} must have "get" permission on all objects of the referenced paramKind (kind=Machine, apiVersion=machine.openshift.io/v1beta1)
```

pkg/operator/sync.go
```diff
@@ -268,6 +268,42 @@ func (optr *Operator) syncWebhookConfiguration(config *OperatorConfig) error {
...
+		return err
+	}
+	if _, _, err := resourceapply.ApplyValidatingAdmissionPolicyV1(context.TODO(),
+		optr.kubeClient.AdmissionregistrationV1(), recorder,
+		mapiwebhooks.NewVSphereFailureDomainCPMSVAP(), optr.cache); err != nil {
+		return err
+	}
+	if _, _, err := resourceapply.ApplyValidatingAdmissionPolicyBindingV1(context.TODO(),
+		optr.kubeClient.AdmissionregistrationV1(), recorder,
+		mapiwebhooks.NewVSphereFailureDomainCPMSVAPBinding(), optr.cache); err != nil {
```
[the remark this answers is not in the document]
vr4manta: HI coderabbit, here is Claude's response:

Every single other `resourceapply` call in the file — `ApplyDeployment`, `ApplyDaemonSet`, all four webhook Apply calls — uses `context.TODO()`. The finding asks to single out the VAP calls for special treatment that no equivalent call in the file receives. That is **inconsistent** and therefore wrong to apply in isolation. If this is a real concern, it belongs as a follow-up that touches all `context.TODO()` sites uniformly, not just these four.

**Skipping** — the finding is not valid as scoped. The VAP calls are identical in pattern to every other `resourceapply` call in the same file; applying a timeout only here would introduce an arbitrary inconsistency with no safety benefit. No change made.

vr4manta:
@coderabbit help

vr4manta:
@coderabbitai summary

vr4manta:
/assign @JoelSpeed 
Hi  Joel, can you take a look at this PR and let me know if this approach is OK.

install/0000_30_machine-api-operator_09_rbac.yaml
```diff
@@ -472,13 +472,29 @@ rules:
...
     verbs:
       - get
       - list
       - watch
       - create
       - update
 
+  # The machine-api-operator service account must be able to read Machines,
+  # ControlPlaneMachineSets, and MachineSets so the kube-apiserver can resolve
+  # them as VAP params when validating Infrastructure/cluster updates.
```
JoelSpeed, line 486: Really? Is that documented somewhere in the upstream docs you can reference?

I can see why it would make sense, but I've never thought about it like that before

pkg/operator/sync.go
```diff
@@ -268,6 +269,62 @@ func (optr *Operator) syncWebhookConfiguration(config *OperatorConfig) error {
...
+// syncVSphereFailureDomainVAPs ensures that the ValidatingAdmissionPolicies and their
+// bindings for protecting vSphere failure domains are present and up to date.
+// These policies prevent an administrator from removing a failure domain from the
+// Infrastructure/cluster CR while it is still referenced by Machines or
+// ControlPlaneMachineSets managed by the Machine API Operator.
+func (optr *Operator) syncVSphereFailureDomainVAPs() error {
+	featureGates, err := optr.featureGateAccessor.CurrentFeatureGates()
+	if err != nil {
+		return fmt.Errorf("failed to get current feature gates: %w", err)
+	}
```
JoelSpeed: Do this earlier. The gates extracted here don't change, or, if they do, the whole bin exits. These can be extracted and held as a global in the optr struct earlier in the construction

pkg/operator/sync.go
```diff
@@ -268,6 +269,62 @@ func (optr *Operator) syncWebhookConfiguration(config *OperatorConfig) error {
...
+	if _, _, err := resourceapply.ApplyValidatingAdmissionPolicyV1(context.TODO(),
+		optr.kubeClient.AdmissionregistrationV1(), recorder,
+		mapiwebhooks.NewVSphereFailureDomainMachineSetVAP(), optr.cache); err != nil {
+		return err
+	}
+	if _, _, err := resourceapply.ApplyValidatingAdmissionPolicyBindingV1(context.TODO(),
+		optr.kubeClient.AdmissionregistrationV1(), recorder,
+		mapiwebhooks.NewVSphereFailureDomainMachineSetVAPBinding(), optr.cache); err != nil {
+		return err
+	}
```
JoelSpeed, line 321: Please wrap each err to give us a unique error message per return

pkg/webhooks/vap.go
```diff
@@ -0,0 +1,374 @@
...
+	VAPMachineSetFailureDomainName = "vsphere-failure-domain-in-use-by-machineset"
+
+	// vapMachineBindingName is the name of the binding for the Machine VAP.
+	vapMachineBindingName = "vsphere-failure-domain-in-use-by-machine"
+
+	// vapCPMSBindingName is the name of the binding for the CPMS VAP.
+	vapCPMSBindingName = "vsphere-failure-domain-in-use-by-cpms"
+
+	// vapMachineSetBindingName is the name of the binding for the MachineSet VAP.
+	vapMachineSetBindingName = "vsphere-failure-domain-in-use-by-machineset"
```
JoelSpeed: Inconsistent casing on VAP here and duplication. Can we consolidate on one set of vars for the names

pkg/webhooks/vap.go
```diff
@@ -0,0 +1,374 @@
...
+				},
+			},
+			// Only evaluate when the cluster is running on vSphere.
+			MatchConditions: []admissionregistrationv1.MatchCondition{
+				{
+					Name: "is-vsphere-platform",
+					Expression: `has(object.spec.platformSpec) &&
+object.spec.platformSpec.type == "` + vspherePlatformType + `"`,
+				},
+			},
```
JoelSpeed, line 88: Using status is more reliable for a platform check normally for the infrastructure object

pkg/webhooks/vap.go
```diff
@@ -0,0 +1,374 @@
...
+			// Reusable sub-expressions.
+			Variables: []admissionregistrationv1.Variable{
+				{
+					// fds: the failure domains list from the incoming (updated) Infrastructure spec.
+					Name: "fds",
+					Expression: `has(object.spec.platformSpec.vsphere) &&
+has(object.spec.platformSpec.vsphere.failureDomains)
+  ? object.spec.platformSpec.vsphere.failureDomains
+  : []`,
```
JoelSpeed: This is equivalent
```suggestion
					Expression: `object.?spec.platformSpec.vsphere.failureDomains.orValue('[]')`,
```

pkg/webhooks/vap.go
```diff
@@ -0,0 +1,374 @@
...
+						},
+					},
+				},
+			},
+			// Only evaluate when the cluster is running on vSphere.
+			MatchConditions: []admissionregistrationv1.MatchCondition{
+				{
+					Name: "is-vsphere-platform",
+					Expression: `has(object.spec.platformSpec) &&
+object.spec.platformSpec.type == "` + vspherePlatformType + `"`,
```
JoelSpeed: This is equivalent
```suggestion
					Expression: `object.?spec.platformSpec.type.orValue("") == "` + vspherePlatformType + `"`,
```

pkg/webhooks/vap.go
```diff
@@ -0,0 +1,374 @@
...
+					Name: "fds",
+					Expression: `has(object.spec.platformSpec.vsphere) &&
+has(object.spec.platformSpec.vsphere.failureDomains)
+  ? object.spec.platformSpec.vsphere.failureDomains
+  : []`,
+				},
+				{
+					// machineRegion: the region label of the Machine param (empty string if absent).
+					Name:       "machineRegion",
+					Expression: `has(params.metadata.labels) && "` + machineRegionLabel + `" in params.metadata.labels ? params.metadata.labels["` + machineRegionLabel + `"] : ""`,
```
JoelSpeed: Use the optional pattern as I've shown in the previous examples to tidy this up too

pkg/webhooks/vap.go
```diff
@@ -0,0 +1,374 @@
...
+				{
+					// Pass when:
+					//   - Machine has no region/zone label (not a failure-domain-managed Machine), OR
+					//   - The failure domain is still present in the incoming spec.
+					Expression: `variables.machineRegion == "" || variables.machineZone == "" ||
+variables.fds.exists(fd,
+  fd.region == variables.machineRegion && fd.zone == variables.machineZone
+)`,
+					MessageExpression: `"Infrastructure update would remove vSphere failure domain (region=" + variables.machineRegion + ", zone=" + variables.machineZone + ") that is still in use by Machine '" + params.metadata.name + "'"`,
+					Reason:            ptr.To(metav1.StatusReasonForbidden),
```
JoelSpeed: Forbidden implies that another user could action this (it's a permissions issue), so use invalid instead

JoelSpeed (commented):
I've left a couple of examples to make the CEL more concise, they apply to pretty much all of the CEL so PTAL at how you can use the optional orValue pattern across the rules you've got

pkg/operator/sync.go
vr4manta: sure, sounds good.

pkg/operator/sync.go
vr4manta: makes sense.  I'll update this.

install/0000_30_machine-api-operator_09_rbac.yaml
vr4manta: ah, so during work w/ coderabbit and claude, i tried to make namespace specific but that always seemed to fail.  my only guess is because the infra spec is cluster scope.  Claude made this comment based on what we learned and this note would help future me for sure remember this.  I'll see if I can find any doc on this.

vr4manta: Claude's response:

The requirement is documented in the official Kubernetes docs on **Validating Admission Policies**, in the **"Parameter resources"** section:

> **https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#parameter-resources**

The relevant excerpt (stable since Kubernetes 1.30 when VAP went GA) states:

> The user creating the `ValidatingAdmissionPolicy` object must have **`get`** permission on **all objects** of the referenced `paramKind`.

This is also visible in the error message text itself — `"must have 'get' permission on all objects of the referenced paramKind"` — which is emitted directly by the kube-apiserver's VAP admission handler.

For the KEP (Kubernetes Enhancement Proposal) that introduced this design decision, the background is in:

> **https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/3488-cel-admission-control**

Specifically the **Authorization** section of KEP-3488, which describes why the creator must be authorized to read param objects (to prevent privilege escalation — a user shouldn't be able to create a policy that reads objects they don't have access to).

---

**TL;DR for the team:** The Kubernetes docs page at `kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#parameter-resources` is the canonical reference. The rule is: whoever creates (or updates) the `ValidatingAdmissionPolicy` must have `get` on the `paramKind` resource, enforced by the kube-apiserver at policy creation time — which is why the `machine-api-operator` service account needs the permission before it can apply the VAP.

pkg/operator/sync.go
vr4manta: It seems we do the check here: https://github.com/openshift/machine-api-operator/blob/0edbe3eab584437bb255713e8fe0e2ac3ce9006e/pkg/operator/operator.go#L460

I'll update to store features in optr.

pkg/webhooks/vap.go
vr4manta: Sure, i'll fix the inconsistent naming here.  For consolidating, I'm assuming you are OK with 1 name to be used by both VAP and VAPB, correct?

pkg/webhooks/vap.go
vr4manta: Makes sense.  I"ll update that.

pkg/webhooks/vap.go
vr4manta: Makes sense.  Is there a reason to do full CEL vs what I'm guessing is a hybrid approach?  Just learning so I can remember for future VAP work.

pkg/webhooks/vap.go
vr4manta: ah ok.  i'll switch to that.

pkg/webhooks/vap.go
JoelSpeed: I think historically some people have had issues with cost estimates for these approaches but I think the issue has since been resolved

pkg/webhooks/vap.go
JoelSpeed: Not updated

pkg/webhooks/vap.go
```diff
@@ -0,0 +1,357 @@
...
+					Name:       "is-vsphere-platform",
+					Expression: `object.?spec.platformSpec.type.orValue("") == "` + vspherePlatformType + `"`,
+				},
+			},
+			// Reusable sub-expressions.
+			Variables: []admissionregistrationv1.Variable{
+				{
+					// fds: failure domains from the incoming (updated) Infrastructure spec.
+					Name:       "fds",
+					Expression: `object.?spec.platformSpec.vsphere.failureDomains.orValue([])`,
```
JoelSpeed, line 202: Switch to status

pkg/webhooks/vap.go
```diff
@@ -0,0 +1,357 @@
...
+							},
+						},
+					},
+				},
+			},
+			// Only evaluate when the cluster is running on vSphere.
+			MatchConditions: []admissionregistrationv1.MatchCondition{
+				{
+					Name:       "is-vsphere-platform",
+					Expression: `object.?spec.platformSpec.type.orValue("") == "` + vspherePlatformType + `"`,
```
JoelSpeed: Switch to status

pkg/webhooks/vap.go
vr4manta: sorry, hadn't completely finished before i left thursday.  I've updated this morning.

pkg/webhooks/vap.go
vr4manta: I do not see failure domains in status

vr4manta:
/verified by @vr4manta 
Manual testing shows VAPs are preventing changes to infrastructure CR when machines / CPMS still use the FDs.

JoelSpeed:
/override ci/prow/e2e-metal-ipi

Confident this change is not related to the single test failure on haproxy there

JoelSpeed:
/override ci/prow/e2e-metal-ipi-ovn-ipv6

Confident that this change is not related to the single failure here

PR #1514, opened 2026-07-03
NO-JIRA: Bump golang.org/x/net to 0.56.0 to fix CVE

twoGiants:
## Summary 

Fixes [CVE-2026-25681](https://www.cve.org/CVERecord?id=CVE-2026-25681) which can cause XSS in golang.org/x/net/html. The html package is used indirectly by dependencies.

## Additional Info

As for the RIT process documentation the Jira ticket is not attached to this PR, I quote:

> In some cases, a symbol level analysis will show that the component is not actually exposed to the vulnerability. In these cases the package should still be updated, but not attached to the bug.

## References

Here is the closed issue https://github.com/golang/go/issues/79574 in the go repo.

## Summary by CodeRabbit

* **Chores**
  * Updated several Go indirect dependencies to newer versions.
  * Brought the tooling module up to a newer Go release and removed an explicit toolchain pin.
  * Added one additional indirect tooling package to keep the development environment aligned and current.


damdo (commented):
/approve
/lgtm

matejvasek:
also fixes OCPBUGS-95517

PR #1519, opened 2026-07-15
OCPBUGS-16118: fix vsphere machine stuck in infinite requeue when PatchMachine fails after clone

jcpowermac:
…fails after clone

When a VM clone is initiated, the task ID is saved to an in-memory TaskIDCache. If the subsequent PatchMachine() call fails (e.g. due to a transient ValidatingAdmissionPolicy rejection), the task ID is never persisted to the machine object. On every subsequent reconcile, the cache guard sees a mismatch between the cached task ID and the empty TaskRef in the machine's providerStatus, causing an infinite RequeueAfterError loop. The recovery code in reconciler.create() that handles "cloned but lost taskRef" is never reached.

Fix by clearing the TaskIDCache entry when PatchMachine() fails, so the next Create() call can proceed past the cache guard and reach the existing recovery path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved machine creation handling by clearing stale task information before updates.
  * Ensured failed creation attempts do not retain outdated task data.
  * Added reliable requeue behavior when task status does not match the machine’s provider status.

### DIFF
--- a/pkg/controller/vsphere/actuator.go
+++ b/pkg/controller/vsphere/actuator.go
@@ -109,6 +109,7 @@ func (a *Actuator) Create(ctx context.Context, machine *machinev1.Machine) error
 	}
 
 	if err := scope.PatchMachine(); err != nil {
+		delete(a.TaskIDCache, machine.Name)
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


mdbooth:
@jcpowermac Do you have logs (preferrably a must-gather) from a cluster where you saw:
> (e.g. due to a transient ValidatingAdmissionPolicy rejection)

Independent of this fix, I'm very interested in that.

PR #1527, opened 2026-07-30
OCPBUGS-100056: fix: skip Multisubnet test for single network infra

abhay-nutanix:
Guard in the BeforeEach that reuses the existing hasMultiSubnetConfiguration() helper (which returns true only when some failure domain has more than one subnet). When no multi-subnet infrastructure is present, the suite now skips gracefully:

## Summary by CodeRabbit

* **Tests**
  * Multi-subnet conformance checks now skip automatically when the cluster does not use a multi-subnet configuration.
  * Prevents irrelevant validation steps from running in unsupported environments.


damdo (commented):
/approve

PR #1528, opened 2026-08-03
OCPBUGS-101813: only block vSphere machine deletion for VMDK-backed v…

jcpowermac:
…olumes

The nodeHasVolumesAttached() check was volume-type-agnostic, blocking machine deletion for any attached volume. This caused indefinite blocking when non-VMDK volumes (NFS, iSCSI, etc.) were attached, since DaemonSet pods with these volumes remain Running and never get evicted.

The VMDK data loss risk only applies to vSphere-backed volumes. This change filters by VolumeAttachment.Spec.Attacher to only block for:
- csi.vsphere.vmware.com (vSphere CSI)
- kubernetes.io/vsphere-volume (vSphere in-tree)

Non-vSphere volumes no longer block deletion. Conservative behavior: if VolumeAttachment lookup fails, treat as vSphere-backed and block.

Includes unit tests for volume type filtering and e2e test for NFS volume deletion scenario.

## Summary by CodeRabbit

* **Bug Fixes**
  * Machine deletion now proceeds when only non-vSphere volumes, such as NFS-backed storage, are attached.
  * Deletion remains safely blocked while vSphere-managed, unidentified, or unresolved volumes may still be attached.
  * Mixed volume attachments are handled correctly, preventing unnecessary deletion delays while protecting active vSphere storage.

* **Tests**
  * Added coverage for NFS, vSphere, mixed, unknown, and other CSI volume scenarios.
  * Added end-to-end validation for deleting machines with non-vSphere volumes.

### DIFF
--- a/pkg/controller/vsphere/reconciler.go
+++ b/pkg/controller/vsphere/reconciler.go
@@ -27,8 +27,10 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	apimachinerytypes "k8s.io/apimachinery/pkg/types"
 	apimachineryutilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/component-base/featuregate"
@@ -41,6 +43,8 @@ import (
 	machinecontroller "github.com/openshift/machine-api-operator/pkg/controller/machine"
 	"github.com/openshift/machine-api-operator/pkg/controller/vsphere/session"
 	"github.com/openshift/machine-api-operator/pkg/metrics"
+
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/controller/vsphere/reconciler.go
```diff
@@ -549,7 +557,91 @@ func (r *Reconciler) nodeHasVolumesAttached(ctx context.Context, nodeName string
...
+	if err := r.apiReader.Get(ctx, apimachinerytypes.NamespacedName{Name: volName}, va); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return nil, err
+		}
+		klog.V(4).Infof("Machine %s: VolumeAttachment %s not found by name, attempting correlation via list", machineName, volName)
+	} else {
+		return va, nil
+	}
+
+	fieldSelector, err := fields.ParseSelector("spec.nodeName=" + nodeName)
```
gnufied: The node selector specified this way will not work for `VolumeAttachment` objects.  For this to work, the k8s API server has to explicitly register `spec.nodeName` as a selectable.  You will have to list all VA objects and filter.

pkg/controller/vsphere/reconciler.go
```diff
@@ -549,7 +557,91 @@ func (r *Reconciler) nodeHasVolumesAttached(ctx context.Context, nodeName string
...
+	return len(vsphereVolumes) > 0 || len(unknownVolumes) > 0, nil
+}
+
+// getVolumeAttachmentForAttachedVolume retrieves the VolumeAttachment corresponding to an attached volume
+// on a node. It first tries a direct lookup by the volume name, then falls back to listing and
+// correlating VolumeAttachments for the node (needed for CSI volumes with hashed names like
+// csi-<sha256(volumeHandle+driver+nodeName)>).
+func (r *Reconciler) getVolumeAttachmentForAttachedVolume(ctx context.Context, volName, nodeName, machineName string) (*storagev1.VolumeAttachment, error) {
+	va := &storagev1.VolumeAttachment{}
+	if err := r.apiReader.Get(ctx, apimachinerytypes.NamespacedName{Name: volName}, va); err != nil {
```
gnufied: This will never work. VA names are hashes, not the name in `node.VolumesAttached`

PR #1535, opened 2026-08-20
OCPBUGS-100316: vsphere: recover machines when the clone TaskRef is lost

mkowalski:
## What this does

Makes the Machine API vSphere provider resilient to a transient failure of the status patch that persists a machine's task reference. Previously a single denied patch (for example, a transient admission-webhook denial during install) permanently wedged the machine in `Provisioning`.

## Why

Root cause of the permanent wedge:

1. `Actuator.Create` tracks the submitted task reference in an in-memory `TaskIDCache`. A staleness guard compared the cache to the Machine's persisted `TaskRef` and, on any mismatch, requeued indefinitely. If the status patch that would have persisted the `TaskRef` was denied, the object never received it, so the guard requeued forever. The cache is only cleared by `Update()`/`Delete()`, which never run because `exists()` keeps returning false for a powered-off VM in `Provisioning`.
2. `reconciler.create()` only recovered a lost `TaskRef` when `InstanceState` was already `PoweredOff`; otherwise it re-cloned, risking duplicate VMs.

## The fix

- **`actuator.go`** — the staleness guard no longer requeues forever. The cache always holds the most recently submitted task (it is updated on every reconcile, before the patch), so when the Machine's persisted `TaskRef` differs from the cache, the actuator recovers the cached reference and reconciles that task. This covers both an empty persisted `TaskRef` (denied patch / informer lag) and a stale nonempty one (object still on the previous task while the cache advanced to the next), avoiding both the permanent wedge and a duplicate clone/power-on.
- **`reconciler.go` `create()`** — when `TaskRef == ""`, look the VM up directly in vCenter before cloning. If it already exists (cloned but the ref was lost), adopt it: restore VM-group membership if `Workspace.VMGroup` is configured (mirroring the completed-clone path so a recovered VM is not left outside its DRS host-affinity group), then power it on. Only clone when the VM is genuinely absent. This makes `create()` idempotent and prevents duplicate VMs.

Net effect: any transient patch failure during VM creation self-heals on the next reconcile instead of leaving workers stuck in `Provisioning`.

## Testing

- `TestActuatorCreateTaskRefLifecycle`:
  - a denied status patch **retains** the task reference in the cache (identity is not lost);
  - a successful patch caches the reference;
  - a lost task reference is reconciled on retry (with a freshly read Machine) without submitting a second clone, reproducing the in-flight window with `simulator.TaskDelay`;
  - a stale **nonempty** task reference (`C` persisted, cache advanced to power-on `P1`) is recovered from the cache on retry without submitting a second power-on.
- `TestCreateRecoversLostTaskRef`: `TaskRef==""` with an existing powered-off VM → no duplicate clone, VM powered on.
- `TestCreateRecoveryRestoresVMGroup`: recovery restores VM-group membership before power-on, no duplicate clone.
- All pass; `go build`, `go vet`, and `gofmt` are clean. Existing `TestCreate/TestExists/TestUpdate/TestDelete/TestClone/TestPowerOn` still pass. Each new test fails on the pre-fix code and passes with the fix.

## Background

This is the provider-side defense described in the OCPBUGS-100316 investigation: the VAP `paramKind` informer warm-up window is transient and inherent, but the vSphere actuator turned a single failed status patch into a permanent failure. No change to `cluster-capi-operator` VAP `failurePolicy` is required.

## Known residual

If the controller process crashes after vCenter accepts a task but before the task reference is recorded in either the Machine status or any durable store, an in-memory cache cannot close that gap. Since the persisting patch is precisely what failed in this scenario, this window is fundamentally bounded rather than something this change can eliminate.

---
This PR was generated using AI (Claude Opus 4.6). Please verify before acting on it.

### DIFF
--- a/pkg/controller/vsphere/actuator.go
+++ b/pkg/controller/vsphere/actuator.go
@@ -97,10 +97,6 @@ func (a *Actuator) Create(ctx context.Context, machine *machinev1.Machine) error
 
 	var retErr error
 	err = newReconciler(scope).create()
-	// save the taskRef in our cache in case of any error with patch.
-	if scope.providerStatus.TaskRef != "" {
-		a.TaskIDCache[machine.Name] = scope.providerStatus.TaskRef
-	}
 	if err != nil {
 		fmtErr := fmt.Errorf(reconcilerFailFmt, machine.GetName(), createEventAction, err)
 		retErr = a.handleMachineError(machine, fmtErr, createEventAction)
@@ -110,6 +106,17 @@ func (a *Actuator) Create(ctx context.Context, machine *machinev1.Machine) error
 
 	if err := scope.PatchMachine(); err != nil {
 		return err
+	}
+
+	// Only cache the taskRef once it has been durably persisted on the Machine
+	// object. Caching it before a successful patch can permanently wedge
+	// creation: if the patch fails (for example, a transient admission-webhook
+	// denial) the object never receives the taskRef and the staleness guard
+	// above would requeue forever. create() is idempotent (it looks the VM up
+	// in vCenter before cloning), so a lost taskRef is recovered on the next
+	// reconcile rather than re-cloning.
+	if scope.providerStatus.TaskRef != "" {
+		a.TaskIDCache[machine.Name] = scope.providerStatus.TaskRef
 	}
 
 	return retErr
--- a/pkg/controller/vsphere/actuator.go
+++ b/pkg/controller/vsphere/actuator.go
@@ -83,15 +83,16 @@ func (a *Actuator) Create(ctx context.Context, machine *machinev1.Machine) error
 		return a.handleMachineError(machine, fmtErr, createEventAction)
 	}
 
-	// If we already submitted a task for this machine (tracked in the in-memory
-	// cache) but the Machine object does not yet reflect it, the status patch
-	// that would have persisted the TaskRef may have failed, or the client cache
-	// may be stale. Recover the TaskRef from the cache so we reconcile the task
-	// we already submitted instead of requeueing forever (which permanently
-	// wedged creation) or dropping the task identity (which could submit a
-	// duplicate clone).
-	if cachedTaskRef, ok := a.TaskIDCache[machine.Name]; ok && scope.providerStatus.TaskRef == "" {
-		klog.Infof("%s: recovering task reference %q from cache; Machine status does not reflect it yet", machine.GetName(), cachedTaskRef)
+	// If the task we last submitted for this machine (tracked in the in-memory
+	// cache) differs from what the Machine object reflects, the status patch
+	// that would have persisted it may have failed, or the client cache may be
+	// stale. The cache always holds the most recently submitted task (it is
+	// updated on every reconcile, before the patch), so recover it and reconcile
+	// that task instead of requeueing forever (which permanently wedged
+	// creation) or reprocessing a stale reference (which could submit a
+	// duplicate clone or power-on).
+	if cachedTaskRef, ok := a.TaskIDCache[machine.Name]; ok && cachedTaskRef != scope.providerStatus.TaskRef {
+		klog.Infof("%s: recovering task reference %q from cache; Machine status reflects %q", machine.GetName(), cachedTaskRef, scope.providerStatus.TaskRef)
 		scope.providerStatus.TaskRef = cachedTaskRef
 	}
 
--- a/pkg/controller/vsphere/actuator.go
+++ b/pkg/controller/vsphere/actuator.go
@@ -5,12 +5,10 @@ package vsphere
 import (
 	"context"
 	"fmt"
-	"time"
 
 	"k8s.io/component-base/featuregate"
 
 	machinev1 "github.com/openshift/api/machine/v1beta1"
-	machinecontroller "github.com/openshift/machine-api-operator/pkg/controller/machine"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/events"
 	"k8s.io/klog/v2"
@@ -18,13 +16,12 @@ import (
 )
 
 const (
-	scopeFailFmt        = "%s: failed to create scope for machine: %v"
-	reconcilerFailFmt   = "%s: reconciler failed to %s machine: %w"
-	createEventAction   = "Create"
-	updateEventAction   = "Update"
-	deleteEventAction   = "Delete"
-	noEventAction       = ""
-	requeueAfterSeconds = 20
+	scopeFailFmt      = "%s: failed to create scope for machine: %v"
+	reconcilerFailFmt = "%s: reconciler failed to %s machine: %w"
+	createEventAction = "Create"
+	updateEventAction = "Update"
+	deleteEventAction = "Delete"
+	noEventAction     = ""
 )
 
 // Actuator is responsible for performing machine reconciliation.
@@ -86,17 +83,25 @@ func (a *Actuator) Create(ctx context.Context, machine *machinev1.Machine) error
 		return a.handleMachineError(machine, fmtErr, createEventAction)
 	}
 
-	// Ensure we're not reconciling a stale machine by checking our task-id.
-	// This is a workaround for a cache race condition.
-	if val, ok := a.TaskIDCache[machine.Name]; ok {
-		if val != scope.providerStatus.TaskRef {
-			klog.Errorf("%s: machine object missing expected provider task ID, requeue", machine.GetName())
-			return &machinecontroller.RequeueAfterError{RequeueAfter: requeueAfterSeconds * time.Second}
-		}
+	// If we already submitted a task for this machine (tracked in the in-memory
+	// cache) but the Machine object does not yet reflect it, the status patch
+	// that would have persisted the TaskRef may have failed, or the client cache
+	// may be stale. Recover the TaskRef from the cache so we reconcile the task
+	// we already submitted instead of requeueing forever (which permanently
+	// wedged creation) or dropping the task identity (which could submit a
+	// duplicate clone).
+	if cachedTaskRef, ok := a.TaskIDCache[machine.Name]; ok && scope.providerStatus.TaskRef == "" {
+		klog.Infof("%s: recovering task reference %q from cache; Machine status does not reflect it yet", machine.GetName(), cachedTaskRef)
+		scope.providerStatus.TaskRef = cachedTaskRef
 	}
 
 	var retErr error
 	err = newReconciler(scope).create()
+	// Remember the submitted task reference even if the patch below fails, so a
+	// retry reconciles the in-flight task instead of submitting a second clone.
+	if scope.providerStatus.TaskRef != "" {
+		a.TaskIDCache[machine.Name] = scope.providerStatus.TaskRef
+	}
 	if err != nil {
 		fmtErr := fmt.Errorf(reconcilerFailFmt, machine.GetName(), createEventAction, err)
 		retErr = a.handleMachineError(machine, fmtErr, createEventAction)
@@ -106,17 +111,6 @@ func (a *Actuator) Create(ctx context.Context, machine *machinev1.Machine) error
 
 	if err := scope.PatchMachine(); err != nil {
 		return err
-	}
-
-	// Only cache the taskRef once it has been durably persisted on the Machine
-	// object. Caching it before a successful patch can permanently wedge
-	// creation: if the patch fails (for example, a transient admission-webhook
-	// denial) the object never receives the taskRef and the staleness guard
-	// above would requeue forever. create() is idempotent (it looks the VM up
-	// in vCenter before cloning), so a lost taskRef is recovered on the next
-	// reconcile rather than re-cloning.
-	if scope.providerStatus.TaskRef != "" {
-		a.TaskIDCache[machine.Name] = scope.providerStatus.TaskRef
 	}
 
 	return retErr
--- a/pkg/controller/vsphere/reconciler.go
+++ b/pkg/controller/vsphere/reconciler.go
@@ -167,10 +167,26 @@ func (r *Reconciler) create() error {
 		}
 
 		// The VM already exists but we have no TaskRef for it: we cloned it
-		// previously and lost the TaskRef. Recover by powering it on and
-		// recording the power-on task so subsequent reconciles can track it,
-		// instead of requeueing forever.
-		klog.Infof("%v: VM already exists without a persisted taskRef, powering on to recover", r.machine.GetName())
+		// previously and lost the TaskRef. Complete the post-clone sequence to
+		// recover — restore VM group membership (if configured) and power the VM
+		// on, recording the power-on task so subsequent reconciles can track it,
+		// instead of requeueing forever. This mirrors the completed-clone path
+		// below so a recovered VM is not left outside its configured VM group.
+		klog.Infof("%v: VM already exists without a persisted taskRef, recovering", r.machine.GetName())
+		if r.machineScope.providerSpec.Workspace.VMGroup != "" {
+			klog.Infof("Adding recovered machine: %s to vm group: %s", r.machine.Name, r.machineScope.providerSpec.Workspace.VMGroup)
+
+			if err := modifyVMGroup(r.machineScope, false); err != nil {
+				var taskError task.Error
+				if errors.As(err, &taskError) {
+					return fmt.Errorf("could not update VM Group membership: %w", taskError)
+				}
+
+				return fmt.Errorf("could not update VM Group membership: %w", err)
+			}
+		}
+
+		klog.Infof("%v: powering on recovered machine", r.machine.GetName())
 		task, err := powerOn(r.machineScope)
 		if err != nil {
 			metrics.RegisterFailedInstanceCreate(&metrics.MachineLabels{
--- a/pkg/controller/vsphere/reconciler.go
+++ b/pkg/controller/vsphere/reconciler.go
@@ -130,39 +130,53 @@ func (r *Reconciler) create() error {
 			return fmt.Errorf("%v: not connected to a vCenter", r.machine.GetName())
 		}
 
-		// Attempt to power on instance in situation where we alredy cloned the instance and lost taskRef.
-		klog.V(4).Infof("%v: InstanceState is: %q", r.machine.GetName(), ptr.Deref(r.machineScope.providerStatus.InstanceState, ""))
-		if types.VirtualMachinePowerState(ptr.Deref(r.machineScope.providerStatus.InstanceState, "")) == types.VirtualMachinePowerStatePoweredOff {
-			klog.Infof("Powering on cloned machine without taskID: %v", r.machine.Name)
+		// A missing TaskRef usually means the VM has not been cloned yet. It can
+		// also mean we cloned the VM successfully but lost the TaskRef because
+		// the status patch that would have persisted it failed (for example, a
+		// transient admission-webhook denial during install). Look the VM up
+		// directly in vCenter before cloning so that a lost TaskRef never
+		// results in a duplicate VM: if the VM already exists we adopt it and
+		// power it on, otherwise we clone the template.
+		if _, err := findVM(r.machineScope); err != nil {
+			if !isNotFound(err) {
+				metrics.RegisterFailedInstanceCreate(&metrics.MachineLabels{
+					Name:      r.machine.Name,
+					Namespace: r.machine.Namespace,
+					Reason:    "FindVM finished with error",
+				})
+				return err
+			}
 
-			task, err := powerOn(r.machineScope)
+			klog.Infof("%v: cloning", r.machine.GetName())
+			task, err := clone(r.machineScope)
 			if err != nil {
 				metrics.RegisterFailedInstanceCreate(&metrics.MachineLabels{
 					Name:      r.machine.Name,
 					Namespace: r.machine.Namespace,
-					Reason:    "PowerOn task finished with error",
+					Reason:    "Clone task finished with error",
 				})
-
 				conditionFailed := conditionFailed()
 				conditionFailed.Message = err.Error()
 				statusError := setProviderStatus(task, conditionFailed, r.machineScope, nil)
 				if statusError != nil {
 					return fmt.Errorf("failed to set provider status: %w", err)
 				}
-
-				return fmt.Errorf("%v: failed to power on machine: %w", r.machine.GetName(), err)
+				return err
 			}
-
 			return setProviderStatus(task, conditionSuccess(), r.machineScope, nil)
 		}
 
-		klog.Infof("%v: cloning", r.machine.GetName())
-		task, err := clone(r.machineScope)
+		// The VM already exists but we have no TaskRef for it: we cloned it
+		// previously and lost the TaskRef. Recover by powering it on and
+		// recording the power-on task so subsequent reconciles can track it,
+		// instead of requeueing forever.
+		klog.Infof("%v: VM already exists without a persisted taskRef, powering on to recover", r.machine.GetName())
+		task, err := powerOn(r.machineScope)
 		if err != nil {
 			metrics.RegisterFailedInstanceCreate(&metrics.MachineLabels{
 				Name:      r.machine.Name,
 				Namespace: r.machine.Namespace,
-				Reason:    "Clone task finished with error",
+				Reason:    "PowerOn task finished with error",
 			})
 			conditionFailed := conditionFailed()
 			conditionFailed.Message = err.Error()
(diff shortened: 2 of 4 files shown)


mkowalski:
## Disposition: REQUEST_CHANGES

Two independent functional defects are **BLOCKING**. Both were confirmed with executable, package-local tests against the reviewed head `19bf23605e6b7f901a3800d5fdf222325d6eda10`.

## Specialist Findings

<details>
<summary><strong>Bugs</strong> — 2 BLOCKING</summary>

1. **Preserve in-flight clone identity across a failed patch** — `pkg/controller/vsphere/actuator.go:119`

   `clone()` returns when vCenter accepts an asynchronous task, before the new VM must be discoverable. If `PatchMachine` then fails, this change leaves neither the Machine nor `TaskIDCache` with that task reference. A retry during the visibility window submits a second clone. The tracked second task can fail with `DuplicateName` after the original succeeds, leaving the Machine tracking the failed task while its VM remains powered off.

2. **Restore VM-group membership during lost-task recovery** — `pkg/controller/vsphere/reconciler.go:174`

   The normal completed-clone path runs `modifyVMGroup(..., false)` before power-on. The new existing-VM recovery path powers on directly and records a power-on task, so it never returns to the only create-side VM-group update. The recovered VM permanently remains outside its requested DRS host-affinity group.

</details>

<details>
<summary><strong>Adversarial</strong> — 2 BLOCKING</summary>

- Independently derived the same asynchronous clone visibility gap. Specific failure sequence: T1 accepted → status patch rejected → retry sees no VM → T2 submitted → T1 succeeds → T2 fails `DuplicateName` → Machine persists T2 and cannot progress.
- Independently confirmed that a configured `Workspace.VMGroup` is bypassed specifically on lost-TaskRef recovery; update does not repair membership.

</details>

<details>
<summary><strong>Security</strong> — no findings</summary>

Reviewed credential loading, status/task inputs, vSphere lookup and power operations, logging, admission-denial handling, and dependency/build scope. No new injection, authorization bypass, credential disclosure, or supply-chain risk was found.

</details>

<details>
<summary><strong>Architecture</strong> — 2 suggestions, subsumed by BLOCKING findings</summary>

- Resource lookup cannot make asynchronous clone initiation idempotent while no inventory object exists; preserve and actively reconcile the submitted task identity.
- Factor the post-clone sequence so both normal completion and lost-task recovery apply VM-group membership before power-on.

</details>

<details>
<summary><strong>Consistency</strong> — no findings</summary>

Compared task-cache lifecycle, status serialization and patch ordering, VM lookup/error conventions, simulator fixtures, feature-gate setup, and analogous tests. No new duplicate helper or convention drift was found.

</details>

<details>
<summary><strong>QA</strong> — 3 suggestions</summary>

1. Make the denied-patch regression prove the interceptor ran and that the returned error is the injected denial; assert persisted provider status and cache both lack a TaskRef, then wait for the asynchronous clone task.
2. Cover a non-not-found `findVM` failure and assert no clone is submitted.
3. Cover power-on failure after adopting an existing VM and assert the failed condition, wrapped error, empty TaskRef, and unchanged inventory.

</details>

<details>
<summary><strong>Technical Writer</strong> — no findings</summary>

The change adds no public API, configuration, status meaning, or user-facing contract. Existing documentation does not describe this internal transient recovery sequence, so no documentation update is required.

</details>

## Panel Synthesis

The cache move fixes the permanent stale-cache requeue, but discards the only identity for a clone that is accepted and not yet visible. `findVM` protects only after the inventory object exists; it does not protect the asynchronous interval that begins when `clone()` returns. The runtime reproducer deterministically produced two clone submissions and persisted the failed second task.

The recovery branch also omits a required post-clone operation. VM-group membership is not advisory: it carries the configured placement policy. This defect was independently identified by the bugs, adversarial, and architecture reviews and was already raised in [the existing CodeRabbit thread](https://github.com/openshift/machine-api-operator/pull/1535#discussion_r3820753513). The runtime reproducer confirmed the VM reached `poweredOn` with an empty group membership list.

<details>
<summary><strong>Confirmed reproducer: duplicate asynchronous clone submission</strong></summary>

**Command**

```text
go test ./pkg/controller/vsphere -run '^TestActuatorCreateCloneRetryWhileFirstTaskDelayed$' -count=1 -v
```

The test used local `govmomi/simulator.TaskDelay` (`CloneVm=2000ms`, lock handoff disabled), rejected only the first status patch, and retried immediately.

**Expected:** one clone submission; the retry retains/reconciles T1.

**Actual:**

```text
after first Create: error="temporary status rejection" cloneTasks=1 trackedTaskRef=task-55 state=running vmCount=2 cacheEntries=0
after second Create: error=<nil> cloneTasks=2 firstTaskRef=task-55 firstState=running secondTaskRef=task-56 secondState=running vmCount=2 cacheRef=task-56
completed: cloneTasks=2 firstTaskRef=task-55 firstState=success firstWaitError=<nil> secondTaskRef=task-56 secondState=error secondWaitError=*types.DuplicateName vmCount=3 (initial=2)
PASS
```

T1 succeeded. The Machine/cache tracked T2, which failed `DuplicateName`.

</details>

<details>
<summary><strong>Confirmed reproducer: VM-group membership skipped</strong></summary>

**Command**

```text
go test ./pkg/controller/vsphere -run '^TestReproLostTaskRecoverySkipsVMGroup$' -count=1 -v
```

The test configured `VSphereHostVMGroupZonal`, an empty `ClusterVmGroup`, a matching powered-off VM, and an empty provider TaskRef.

**Expected:** recovery adds the VM to `Workspace.VMGroup` before power-on.

**Actual:**

```text
VM already exists without a persisted taskRef, powering on to recover
BUG REPRODUCED: task=task-54 taskType=Datacenter.powerOnMultiVM powerState=poweredOn vmGroup=lost-task-recovery-group members=[] recoveredVM=VirtualMachine:vm-51 member=false
PASS
```

The VM powered on successfully while its configured VM group remained empty.

</details>

## Required Actions

1. Preserve the accepted clone task identity across a failed Machine status patch. Replace the stale-object infinite requeue with active task reconciliation/status persistence so a retry cannot submit T2 while T1 is in flight.
2. In the existing-VM lost-TaskRef branch, run the same feature-gated `modifyVMGroup(..., false)` sequence and error handling used after normal clone completion, before `powerOn`.
3. Add regression coverage for both confirmed sequences. The clone test must hold T1 before inventory visibility and prove only one clone is submitted; the recovery test must assert group membership as well as power state.

## Optional Follow-ups

- Strengthen the denied-patch test so unrelated setup/clone failures cannot satisfy it.
- Add focused tests for non-not-found lookup errors and adopted-VM power-on errors.

## Stats

- Specialists: 7/7 completed
- Raw findings: 9
- Deduplicated findings: 2 BLOCKING, 3 optional test improvements
- BLOCKING reproducers: 2/2 confirmed
- Reviewed range: `951196122fe4f1115782da6c14660116905029c4...19bf23605e6b7f901a3800d5fdf222325d6eda10`
- Existing changed-contract tests: passed
  - `go test ./pkg/controller/vsphere -run 'TestActuatorCreateCachesTaskRefOnlyAfterSuccessfulPatch|TestCreateRecoversLostTaskRef' -count=1`
- PR CI: skipped because the PR is draft

---
This message was generated using AI. Please verify before acting on it.

Assisted-By: github-copilot/gpt-5.6-sol

<sub>Generated by [/code-review:deep-review](https://github.com/openshift-eng/ai-helpers/tree/main/plugins/code-review/skills/deep-review)</sub>

pkg/controller/vsphere/actuator.go
```diff
@@ -112,6 +108,17 @@ func (a *Actuator) Create(ctx context.Context, machine *machinev1.Machine) error
...
 
+	// Only cache the taskRef once it has been durably persisted on the Machine
+	// object. Caching it before a successful patch can permanently wedge
+	// creation: if the patch fails (for example, a transient admission-webhook
+	// denial) the object never receives the taskRef and the staleness guard
+	// above would requeue forever. create() is idempotent (it looks the VM up
+	// in vCenter before cloning), so a lost taskRef is recovered on the next
+	// reconcile rather than re-cloning.
+	if scope.providerStatus.TaskRef != "" {
+		a.TaskIDCache[machine.Name] = scope.providerStatus.TaskRef
```
mkowalski: **Bug: Preserve the in-flight clone identity across a failed patch**

`clone()` returns before its asynchronous task must create a discoverable VM. If the status patch fails, leaving this cache empty permits the immediate retry to see no VM and submit a second clone. The reproducer delayed T1 before inventory visibility: the retry submitted T2, T1 succeeded, and the persisted/tracked T2 failed `*types.DuplicateName`.

Fix: retain the accepted task identity across patch failure and make the stale-object path actively reconcile/persist that task instead of either infinite-requeueing or discarding it. Add a delayed-task regression that proves only one clone submission.

<details>
<summary>Reproducer</summary>

**Command:** `go test ./pkg/controller/vsphere -run '^TestActuatorCreateCloneRetryWhileFirstTaskDelayed$' -count=1 -v`

**Expected:** one clone task.

**Actual:** `cloneTasks=2`; `task-55` succeeded; tracked `task-56` failed `*types.DuplicateName`; inventory increased by one.
</details>

_AI-generated; verify before acting._

Assisted-By: github-copilot/gpt-5.6-sol

pkg/controller/vsphere/reconciler.go
```diff
@@ -130,47 +130,61 @@ func (r *Reconciler) create() error {
...
 		}
 
-		klog.Infof("%v: cloning", r.machine.GetName())
-		task, err := clone(r.machineScope)
+		// The VM already exists but we have no TaskRef for it: we cloned it
+		// previously and lost the TaskRef. Recover by powering it on and
+		// recording the power-on task so subsequent reconciles can track it,
+		// instead of requeueing forever.
+		klog.Infof("%v: VM already exists without a persisted taskRef, powering on to recover", r.machine.GetName())
+		task, err := powerOn(r.machineScope)
```
mkowalski, line 190: **Bug: Restore VM-group membership during lost-task recovery**

This recovery path skips the `modifyVMGroup(..., false)` step used after normal clone completion. Once this power-on task is recorded, the Machine never returns to the clone-task branch, and update does not repair membership. The reproducer powered the recovered VM on while `ClusterVmGroup.Vm` remained empty. This independently confirms the existing CodeRabbit finding.

Fix: apply the same feature-gated VM-group update and error handling before `powerOn`, then assert membership in the lost-TaskRef regression.

<details>
<summary>Reproducer</summary>

**Command:** `go test ./pkg/controller/vsphere -run '^TestReproLostTaskRecoverySkipsVMGroup$' -count=1 -v`

**Expected:** VM is added to `Workspace.VMGroup` before power-on.

**Actual:** `powerState=poweredOn ... members=[] ... member=false`.
</details>

_AI-generated; verify before acting._

Assisted-By: github-copilot/gpt-5.6-sol

mkowalski (commented):
Deep-review inline findings. Full panel verdict and both confirmed reproducers: https://github.com/openshift/machine-api-operator/pull/1535#issuecomment-5354865144

This review was generated using AI. Please verify before acting on it.

Assisted-By: github-copilot/gpt-5.6-sol

mkowalski:
Thanks for the detailed review — both BLOCKING findings were valid and are now addressed in `302951f`.

## 1. Preserve clone task identity across a failed status patch

You were right: caching the `TaskRef` only *after* a successful patch discarded the clone identity on failure, so a retry during the in-flight window (before the VM is discoverable by `findVM`) could submit a second clone → `DuplicateName`.

Resolution:
- The actuator now caches the `TaskRef` **regardless** of whether `PatchMachine` succeeds, so a submitted clone is never forgotten.
- The staleness guard no longer requeues forever. When the cache holds a ref the Machine object does not yet reflect, it **seeds that ref and reconciles the in-flight task** instead. This fixes both the original permanent wedge and the duplicate-clone window.
- `findVM`-before-clone remains as the cold-cache / process-restart defense.

## 2. Restore VM-group membership during lost-task recovery

Also correct. The existing-VM recovery branch now runs the same feature-gated `modifyVMGroup(..., false)` sequence and error handling as the normal completed-clone path, before power-on, so a recovered VM is not left outside its configured DRS host-affinity group. (The `VSphereHostVMGroupZonal` gate is already enforced at the top of `create()` for all paths.)

## Regression coverage

- `TestActuatorCreateTaskRefLifecycle`
  - a denied status patch **retains** the clone task reference;
  - a successful patch caches it;
  - reproducing your in-flight window with `simulator.TaskDelay{CloneVm: 2000, LockHandoff: 0}`, a retry reconciles the cached task and submits **no second clone** (the test self-validates the window by asserting the cloned VM is not yet materialized).
- `TestCreateRecoveryRestoresVMGroup`: the recovered VM is a member of its configured VM group before power-on, with no duplicate clone.

All four new/updated tests fail on the previous head and pass with `302951f`; `go build`, `go vet`, and `gofmt` are clean. Existing `TestCreate/TestExists/TestUpdate/TestDelete/TestClone/TestPowerOn` still pass.

## One residual corner (by design, not fixable here)

A duplicate clone is still theoretically possible only if the controller process **crashes** after `clone()` is accepted but before the cloned VM is inventory-visible **and** the status patch had failed. Closing that fully would require durable state that a failed patch cannot provide, so it is fundamentally bounded rather than something this change can eliminate. Happy to discuss if you'd like a follow-up (e.g. persisting the task ref via a separate, VAP-exempt path).

I did not adopt the optional QA suggestions (assert the interceptor ran / non-not-found `findVM` error / adopted-VM power-on failure) — let me know if you'd like them added.

---
This message was generated using AI. Please verify before acting on it.

Assisted-By: Claude Opus 4.6

mkowalski:
## Follow-up review of `302951f`

The replies are mostly correct:

- **Original duplicate-clone blocker:** fixed for the intended empty-status case. I reran the delayed-clone scenario with a **fresh Machine fetched from the fake API after the denied patch**; `Create` recovered `task-55` from `TaskIDCache` and did not submit a second clone.
- **VMGroup recovery blocker:** fixed. The new recovery path restores membership before power-on, and `TestCreateRecoveryRestoresVMGroup` passes.
- **Process-crash residual:** the explanation makes sense. If the controller dies after vCenter accepts the clone but before either API status or any other durable store records its task ID, an in-memory cache cannot close that gap.

One in-scope mismatch remains **BLOCKING**, however.

### Cached task recovery ignores a stale nonempty Machine TaskRef

`Actuator.Create` recovers the cached task only when `scope.providerStatus.TaskRef == ""`. Status patch failure or informer lag can also leave the Machine on the **previous nonempty task** while `TaskIDCache` already contains the next task.

Concrete sequence:

1. The Machine and cache contain completed clone task `C`.
2. `Create` processes `C` and submits power-on task `P1`.
3. The status patch for `P1` is denied; the cache advances to `P1`, while the API object still contains `C`.
4. A retry reads that fresh API object. Because `C` is nonempty, line 93 does not recover `P1`.
5. `Create` processes `C` again and submits `P2`.

A local simulator reproducer delayed `PowerOnMultiVM` and fetched a fresh Machine after the denied patch:

```text
Powering on cloned machine: stale-nonempty
Failed to patch machine status "stale-nonempty": admission webhook denied the request
Powering on cloned machine: stale-nonempty
Failed to patch machine status "stale-nonempty": admission webhook denied the request
BUG REPRODUCED: clone=task-55 cachedP1=task-58 cachedP2=task-59 powerOnTasks=2
PASS
```

The recovery condition needs to handle **any mismatch**, not only empty status—for example, recover when `cachedTaskRef != scope.providerStatus.TaskRef`. That preserves the original staleness invariant while reconciling instead of indefinitely requeueing.

The new delayed-clone test also retries with the same `machine` pointer. `PatchMachine` assigns the new provider status to that pointer before the intercepted status patch fails, so the test's claim that the TaskRef “lives only in the cache” is not true on its second call. Fetch a fresh Machine from `base` before retrying; then add the nonempty `C → P1` mismatch case above.

## Verification

Passed on current head:

```text
go test ./pkg/controller/vsphere -run 'TestActuatorCreateTaskRefLifecycle|TestCreateRecoversLostTaskRef|TestCreateRecoveryRestoresVMGroup' -count=1
```

Also passed after changing the delayed-clone test to retry with a fresh API object. The separate stale-nonempty reproducer confirmed two power-on submissions.

---
This message was generated using AI. Please verify before acting on it.

Assisted-By: github-copilot/gpt-5.6-sol

<sub>Follow-up to [/code-review:deep-review](https://github.com/openshift-eng/ai-helpers/tree/main/plugins/code-review/skills/deep-review)</sub>

pkg/controller/vsphere/actuator.go
```diff
@@ -86,18 +83,22 @@ func (a *Actuator) Create(ctx context.Context, machine *machinev1.Machine) error
...
-			return &machinecontroller.RequeueAfterError{RequeueAfter: requeueAfterSeconds * time.Second}
-		}
+	// If we already submitted a task for this machine (tracked in the in-memory
+	// cache) but the Machine object does not yet reflect it, the status patch
+	// that would have persisted the TaskRef may have failed, or the client cache
+	// may be stale. Recover the TaskRef from the cache so we reconcile the task
+	// we already submitted instead of requeueing forever (which permanently
+	// wedged creation) or dropping the task identity (which could submit a
+	// duplicate clone).
+	if cachedTaskRef, ok := a.TaskIDCache[machine.Name]; ok && scope.providerStatus.TaskRef == "" {
```
mkowalski: **Bug: recover any stale TaskRef mismatch, not only an empty one**

A status patch can fail when advancing from one nonempty task to the next. Reproduced sequence: the API object retained completed clone task `C`; the cache advanced to delayed power-on `P1`; retrying with a fresh API object ignored `P1` because `C != ""`, processed `C` again, and submitted `P2`. Runtime evidence: `clone=task-55 cachedP1=task-58 cachedP2=task-59 powerOnTasks=2`. Informer lag after a successful patch creates the same mismatch.

Fix: recover when the cached ref differs from provider status, not only when provider status is empty (for example, `cachedTaskRef != scope.providerStatus.TaskRef`). Update the regression to fetch a fresh Machine before retrying; the current test reuses the pointer that `PatchMachine` already mutated before the intercepted patch failed, so it does not prove cache-only recovery. Add the nonempty `C → P1` case.

_AI-generated; verify before acting._

Assisted-By: github-copilot/gpt-5.6-sol

mkowalski (commented):
Follow-up to the deep review. The original empty-TaskRef clone race and VMGroup findings are fixed, but one stale nonempty TaskRef mismatch remains. Full evidence is in the follow-up PR comment.

This review was generated using AI. Please verify before acting on it.

Assisted-By: github-copilot/gpt-5.6-sol

mkowalski:
Good catch — both points are valid and are fixed in `a1263f8`.

## Stale nonempty TaskRef not recovered (BLOCKING)

Confirmed. The guard only recovered the cached ref when the Machine's `TaskRef` was empty, so the `C → P1` sequence you describe (object stuck on the finished clone `C`, cache advanced to power-on `P1`, patch denied) skipped recovery and `create()` reprocessed `C`, resubmitting power-on.

Fix: recover on **any** mismatch, exactly as you suggested:

```go
if cachedTaskRef, ok := a.TaskIDCache[machine.Name]; ok && cachedTaskRef != scope.providerStatus.TaskRef {
    scope.providerStatus.TaskRef = cachedTaskRef
}
```

This is safe because the cache always holds the most recently submitted task (it is updated on every reconcile, before the patch), so it is always ≥ the persisted object — recovering it always reconciles the latest task and never resurrects an older one. It also preserves the original staleness invariant, just reconciling instead of requeueing.

## Misleading delayed-clone test (valid)

Also correct: `PatchMachine` assigns the new provider status to the in-memory Machine before the (denied) status patch, so reusing the pointer let the retry read the ref off the object rather than the cache. The test now retries with a **freshly read Machine** from the fake client (and asserts its `ProviderStatus` is nil, proving the denied patch persisted nothing), so it genuinely exercises cache recovery.

## New regression coverage

- `TestActuatorCreateTaskRefLifecycle/a stale nonempty task reference is reconciled from the cache on retry`: clones successfully (object+cache on `C`), then holds `PowerOnMultiVM` in-flight with denied status patches. The first reconcile submits `P1` and advances the cache; a retry with a freshly read Machine (still on `C`) must recover `P1` and submit **no** second power-on. Fails on the old `== ""` guard (submits a second power-on), passes with the mismatch guard.
- The delayed-clone subtest now retries with a fresh Machine as above.

## Verification

```
go test ./pkg/controller/vsphere -run 'TestActuatorCreateTaskRefLifecycle|TestCreateRecoversLostTaskRef|TestCreateRecoveryRestoresVMGroup|TestCreate$|TestExists|TestUpdate$|TestDelete$|TestClone$|TestPowerOn$' -count=1
```

All pass; `go build`, `go vet`, `gofmt` clean. Reverting only the guard change makes the new stale-nonempty subtest fail (and only that one), confirming it guards the fix.

---
This message was generated using AI. Please verify before acting on it.

Assisted-By: Claude Opus 4.6

mkowalski:
## Follow-up review of `a1263f8`

The response and code change make sense. The remaining task-cache blocker is fixed:

- `Actuator.Create` now recovers the cached ref on **any mismatch**, so a stale nonempty clone ref cannot replay clone completion and submit a second power-on.
- The delayed-clone test now retries with a freshly fetched Machine whose denied status patch persisted no provider status.
- The new stale-nonempty regression faithfully exercises `C → P1`: the API object remains on `C`, the cache advances to `P1`, and retry submits no `P2`.

Focused current-head verification passed:

```text
go test ./pkg/controller/vsphere -run 'TestActuatorCreateTaskRefLifecycle|TestCreateRecoversLostTaskRef|TestCreateRecoveryRestoresVMGroup|TestCreate$|TestExists|TestUpdate$|TestDelete$|TestClone$|TestPowerOn$' -count=1
```

I found no further code blocker in `a1263f8`.

One remaining documentation correction: the PR description still states the opposite of the current implementation:

- “cache the `TaskRef` only **after** `PatchMachine` succeeds” is no longer true;
- the Testing section still names `TestActuatorCreateCachesTaskRefOnlyAfterSuccessfulPatch` and says a denied patch leaves the cache empty.

Please update the description to say that the latest submitted task is retained before patching and recovered from the in-memory cache when persisted status differs, and rename the documented test to `TestActuatorCreateTaskRefLifecycle` with its empty and stale-nonempty mismatch cases.

I also checked CodeRabbit's instance-UUID-only concern. `Exists()` already called the same `findVM` UUID-then-name lookup before this PR, and the previous powered-off recovery path called `powerOn()`, which uses that same lookup. The name fallback may merit a separate ownership-policy discussion, but it is not introduced by these changes.

---
This message was generated using AI. Please verify before acting on it.

Assisted-By: github-copilot/gpt-5.6-sol

<sub>Follow-up to [/code-review:deep-review](https://github.com/openshift-eng/ai-helpers/tree/main/plugins/code-review/skills/deep-review)</sub>

mkowalski:
Thanks — no code changes needed this round; agreed on both points.

**Description updated.** It now reflects the current implementation:
- the most recently submitted task reference is retained in the cache **before** patching, and recovered from the cache whenever the persisted status differs (empty or stale-nonempty), rather than "cached only after a successful patch";
- the Testing section now names `TestActuatorCreateTaskRefLifecycle` with its denied-patch, successful-patch, lost-ref, and stale-nonempty mismatch cases, plus `TestCreateRecoversLostTaskRef` and `TestCreateRecoveryRestoresVMGroup`.

**Instance-UUID / name fallback:** agreed it's out of scope here. `findVM` already did the UUID-then-name lookup, and both `Exists()` and the pre-existing powered-off `powerOn()` recovery used it, so this PR doesn't introduce or change that behavior. Happy to take the ownership-policy discussion to a separate issue if it's worth pursuing.

---
This message was generated using AI. Please verify before acting on it.

Assisted-By: Claude Opus 4.6

mkowalski:
## Follow-up after description update

Confirmed. No new commit was added; the reviewed head remains `a1263f82fdd91f30d8b63261b41ffc7e0f5bf5e4`.

The updated PR description now matches the implementation and tests:

- the latest submitted task is cached before status patching;
- any persisted/cache TaskRef mismatch is reconciled from the cache;
- both empty and stale-nonempty mismatch regressions are documented under `TestActuatorCreateTaskRefLifecycle`;
- lost-TaskRef VM recovery and VMGroup restoration tests are named correctly;
- the unavoidable process-crash window is disclosed.

The instance-UUID/name-fallback assessment is also correct: that lookup behavior existed in `findVM`, `Exists`, and the prior powered-off recovery path, so it is not introduced by this PR.

Focused verification on the unchanged head still passes:

```text
go test ./pkg/controller/vsphere -run 'TestActuatorCreateTaskRefLifecycle|TestCreateRecoversLostTaskRef|TestCreateRecoveryRestoresVMGroup' -count=1
```

No further code or description pushback from this review. CodeRabbit's bounded test-context suggestion is reasonable hardening, but the same unbounded simulator wait pattern already exists throughout this package and it is not a merge blocker for this fix.

---
This message was generated using AI. Please verify before acting on it.

Assisted-By: github-copilot/gpt-5.6-sol

<sub>Follow-up to [/code-review:deep-review](https://github.com/openshift-eng/ai-helpers/tree/main/plugins/code-review/skills/deep-review)</sub>

PR #1557, opened 2026-09-20
OCPBUGS-126898: Don't default vSphere numCoresPerSocket to numCPUs

nikhilprajapati-world:
## Summary
- When `providerSpec.numCoresPerSocket` is omitted (0), clone no longer sets it equal to `numCPUs`.
- The clone ConfigSpec leaves `NumCoresPerSocket` unset so vSphere uses the template analogue, which is what the API documents.
- Unit test covers omit vs explicit cores-per-socket.

Fixes OCPBUGS-126898

## Test plan
- [x] `go test ./pkg/controller/vsphere/ -count=1 -run TestCoresPerSocketForClone`
- [x] Lab on OpenShift 5.0.0-rc.2 vSphere: custom MachineSet `numCPUs: 8`, omit `numCoresPerSocket`. Before this change, guest `lscpu` was 8 CPUs / 8 cores/socket / 1 socket (IPI workers 4/4/1). After this change, omitted field must follow the template (4 cores/socket → 2 sockets for 8 vCPU), not 1×8.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vSphere virtual machine cloning when the cores-per-socket setting is unset.
  * Clone operations now preserve the template’s compatible CPU topology instead of forcing all vCPUs onto one socket, reducing clone failures on incompatible hosts.


jcpowermac:
@nikhilprajapati-world @coderabbitai I think you are forgetting something - we should have less numa nodes not more, and your tests left out the odd number of cores case, which customers' have used in the past.

jcpowermac:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 🧩 Analysis chain
> `@jcpowermac` You are correct that the test does not cover an odd explicit value.
> 
> `NumCoresPerSocket` controls virtual socket topology. It does not directly define the NUMA-node count. With an omitted value, this change lets vCenter retain the template topology. The lab validation must confirm the guest-visible socket and NUMA-node topology. It must not assume that a higher socket count always means a higher NUMA-node count.
> 
> [plan_code_changes]
> 
> * Add an odd `numCoresPerSocket` test case.
> * Keep the test focused on the clone `ConfigSpec` contract.
> * Update the lab validation to report guest socket and NUMA-node topology for the template-derived case.
> 
> > 📝 Learnings were identified and not saved (knowledge base disabled). [Enable](https://app.coderabbit.ai/repository/141598983/settings/knowledge-base)
> 
> _You are interacting with an AI system._

@coderabbitai you are wrong, if you increase the number of sockets that directly increases numa nodes, which incurs a latency cost

jcpowermac:
/test unit
/test e2e-vsphere-ovn

/lgtm
/hold

I think the change is fine, it gets us back to the contract defined in the API. vsphere CI needs to pass, I would like to see the odd case.

